PR #30, opened 2022-08-30
[Auth] Add support for next-auth integration

adam-maj:


packages/auth/.gitignore
```diff
@@ -0,0 +1,3 @@
+/express
+/next
+/next-auth
```
jnsdls, line 3: these are checked in on purpose

jnsdls: if these are not checked in preconstruct will complain about it during every build

jnsdls (changes requested):
we should check in the entry-points

jnsdls (approved):
LGTM

PR #56, opened 2022-09-02
Add CLI tests

nachoiacovino:


packages/cli/e2e/deploy.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      waitForText,
+      waitForFinish,
+      getExitCode,
+      writeText,
+      getStdout,
+    } = await spawn("node", "./dist/cli/index.js deploy");
+
+    expect(await exists("BasicContract.sol")).toEqual(true);
+
+    await waitForText("Failed to find a supported project configuration");
```
jnsdls, line 24: let's use a string map for this one, too

packages/cli/e2e/release.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      waitForText,
+      waitForFinish,
+      getExitCode,
+      writeText,
+      getStdout,
+    } = await spawn("node", "./dist/cli/index.js release");
+
+    expect(await exists("BasicContract.sol")).toEqual(true);
+
+    await waitForText("Failed to find a supported project configuration");
```
jnsdls, line 24: let's use a string map for this one, too

jnsdls (changes requested):
anytime u see `waitForText` think -> `const mapping`

jnsdls (approved):
LGTM

PR #74, opened 2022-09-06
Update Opezeppelin forwarder address for Goerli

nkrishang:
## What's the change?
Updated the openzeppelin forwarder address for the Goerli network, used when deploying prebuilt contracts.

## Why do we need the change?
- The current `OZ_DEFENDER_FORWARDER_ADDRESS` is the wrong forwarder address, only for the Goerli network. 
- If you're deploying prebuilt contracts to Goerli via the dashboard or SDK, gasless transactions won't work for you on Goerli.

## How is the change implemented?
In `sdk/src/constants/addresses.ts`: added an `openzeppelinForwarder` field to `CONTRACT_ADDRESSES`. This allows setting a forwarder address per chain.

In `getDefaultTrustedForwarders` in `sdk/core/classes/factory.ts`: updated the function to account for the above `openzeppelinForwarder` field available per chain Id.


packages/sdk/src/core/classes/factory.ts
```diff
@@ -257,8 +257,11 @@ export class ContractFactory extends ContractWrapper<TWFactory> {
     const biconomyForwarder = chainEnum
       ? CONTRACT_ADDRESSES[chainEnum].biconomyForwarder
       : constants.AddressZero;
+    const openzeppelinForwarder = chainEnum
+      ? CONTRACT_ADDRESSES[chainEnum].openzeppelinForwarder
+      : OZ_DEFENDER_FORWARDER_ADDRESS;
```
joaquim-verges, line 262: default to `constants.AddressZero` to make it safer for unsupported chain Ids

joaquim-verges (approved):

PR #78, opened 2022-09-07
[Solana] Add support for NFT Drop

adam-maj:


packages/solana/src/classes/deployer.ts
```diff
@@ -100,4 +105,18 @@ export class Deployer {
       .run();
     return collectionNft.mint.address.toBase58();
   }
+
+  async createNftDrop(metadata: NFTDropMetadataInput): Promise<string> {
+    invariant(this.wallet.signer, "Wallet is not connected");
+    const parsed = NFTDropContractSchema.parse(metadata);
+
+    // TODO create a collection with metadata and associate it with the drop
```
joaquim-verges, line 113: Still need to do this right?

packages/solana/src/contracts/nft-drop.ts
```diff
@@ -0,0 +1,105 @@
...
+    this.wallet = wallet;
+    this.storage = storage;
+    this.metaplex = metaplex;
+    this.connection = metaplex.connection;
+    this.nft = new NFTHelper(metaplex);
+    this.dropMintAddress = new PublicKey(dropMintAddress);
+    this.claimConditions = new ClaimConditions(dropMintAddress, metaplex);
+  }
+
+  async get(mintAddress: string): Promise<NFTMetadata> {
```
joaquim-verges, line 34: Is there a built-in way to get all tokens from a candy machine?

packages/solana/src/types/contracts/nft-drop.ts
```diff
@@ -0,0 +1,65 @@
...
+export const NFTDropOutputSchema = z.object({
+  price: z.bigint(),
+  sellerFeeBasisPoints: z.bigint(),
+  itemsAvailable: z.bigint(),
+  goLiveDate: z.date().optional(),
+});
+
+export type NFTDropClaimInput = z.input<typeof NFTDropClaimSchema>;
+
+export type NFTDropOutput = z.output<typeof NFTDropOutputSchema>;
```
joaquim-verges, line 65: NFTDrop claim output ?

packages/solana/src/classes/deployer.ts
adam-maj: Yes - want to do this in a separate PR

adam-maj: I think this should go in a separate PR. Experimented with it a bit and its not as straightforward + doesn't seem obvious to me yet that we should do this for every drop (I want to look more into how the candy machine works internally/associates all its NFTs together already).

packages/solana/src/contracts/nft-drop.ts
```diff
@@ -0,0 +1,127 @@
...
+    return this.nft.get(mintAddress);
+  }
+
+  async getAll(): Promise<NFTMetadata[]> {
+    const info = await this.getCandyMachine();
+    const nfts = await Promise.all(
+      info.items.map(async (item) => {
+        const metadata = await this.storage.get(item.uri);
+        return { uri: item.uri, ...metadata };
```
joaquim-verges, line 43: Is there no address inside items we need to pass through here? Also name since it might not be in the JSON meta. 

Should work with candy machines not creates through us.

joaquim-verges: Add a Todo to add a query count / start here like for EVM. Gotta be able to paginate this

packages/solana/src/contracts/nft-drop.ts
```diff
@@ -0,0 +1,127 @@
...
+      .run();
+
+    return result.nft.address.toBase58();
+  }
+
+  private async getCandyMachine() {
+    return this.metaplex
+      .candyMachines()
+      .findByAddress({ address: this.dropMintAddress })
+      .run(); // TODO abstract return types away
```
joaquim-verges, line 125: Can remove the todo here since it's private

packages/solana/src/types/contracts/nft-drop.ts
```diff
@@ -0,0 +1,66 @@
...
+    .optional(),
+  splToken: z
+    .string()
+    .transform((a) => new PublicKey(a))
+    .optional(),
+  solTreasuryAccount: z
+    .string()
+    .transform((a) => new PublicKey(a))
+    .optional(),
+  splTokenAccount: z
```
joaquim-verges, line 31: Add a Todo to handle allow lists?

packages/solana/src/types/nft.ts
```diff
@@ -19,7 +19,7 @@ export const CommonTokenInput = z
  * @internal
  */
 export const CommonTokenOutput = CommonTokenInput.extend({
-  id: z.string(),
+  id: z.string().optional(),
```
joaquim-verges, line 22: Oh right... There's no address for lazy minted tokens

joaquim-verges (approved):
Just minor stuff, feel free to merge

joaquim-verges:
@adam-maj add a lil changeset too plz

packages/solana/src/contracts/nft-drop.ts
adam-maj: items includes unminted NFTs, so those don't have an address yet which is why we do it like this

packages/solana/src/types/nft.ts
adam-maj: Unfortunately

Issue #84, opened 2022-09-08
[CLI] - Set the name of the project in package.json file


jarrodwatts:
## Feature Description

When you create a new project running the `npx thirdweb create --app` command, you get set up with a project that has the name of the starter kit that gets cloned in the `package.json` file.

```
  "name": "next-javascript-starter",
```

We should change this to be the name that the user enters when they create the project.

## Goal

The `name` field in the `package.json` file should be the name the user enters when they create the app.

b4s36t4:
@jarrodwatts Raised a PR. Please have a look at it when you can. Thanks.

PR #103, opened 2022-09-10
[CLI] - Set package.json name to the name the user entered for the project

b4s36t4:
Solves the issue #84


packages/cli/src/create/helpers/create-app.ts
```diff
@@ -141,6 +160,23 @@ export async function createApp({
...
+    const packageString = fs.readFileSync(jsonFile);
+    try {
+      const packageJson = JSON.parse(packageString.toString());
+      if (packageJson) {
+        packageJson.name = appName;
+      }
+
+      fs.writeFileSync(jsonFile, JSON.stringify(packageJson, null, 2) + os.EOL);
+
+      console.log("Updated package.json.");
```
jarrodwatts, line 173: Can we remove these console logs? same with the catch block below.

packages/cli/src/create/helpers/create-app.ts
```diff
@@ -141,6 +160,23 @@ export async function createApp({
...
+      console.log();
+    } catch (e) {
+      console.log("Reading JSON error");
```
jarrodwatts, line 176: We should handle the error here.

I don't think failing to rename the package name is severe enough to exit the process, but we should maybe just console.log this - "Something went wrong renaming package, skipping..."

jarrodwatts:
Thanks Mahesh! Looks good!

A few comments from me and I've asked our engineer @joaquim-verges to take a look too

jnsdls:
Thanks for the PR! Please add a changeset (using `yarn changeset`) that briefly describes what these changes mean to a user.

In this case it can probably be a `patch` version change.

PR #104, opened 2022-09-10
[CLI] - Don't show error when non lowercase letter is entered for project name

### LINKED ISSUE #83: [CLI] - Fix "Invalid Project name" when a project with a capital letter is entered
## Feature Description

When you enter a name for a project on the CLI when you run `npx thirdweb create`, you currently cannot type any capital letters. We shouldn't show this error, we should just automatically convert it to all lowercase.

image

## Goal

The project name should be automatically all lowercase and not show this error.

b4s36t4:
Fixes #83


jarrodwatts:
Thanks Mahesh! This looks great.

One question - when you create the project - is the name transformed to lowercase?

This seems like the validation is changed but the actual project name doesn't get transformed.

e.g. if I type ethBio my project name should be ethbio

b4s36t4:
@jarrodwatts Right now we don't allow the appName to be the package name in json file. This PR only fixes validation which is stopping the user to create the repo. The other PR I can edit that to support the same in package.json as well.

jnsdls:
Thanks for the PR! Please add a changeset (using `yarn changeset`) that briefly describes what these changes mean to a user.

In this case it can probably be a `patch` version change.

b4s36t4:
Oh cool. Got it. Will update the code.

jnsdls:
fixed in #184

PR #107, opened 2022-09-11
Pass empty `trustedForwarders` for Pack

kumaryash90:


packages/contracts-js/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/thirdweb-dev/js/tree/main/packages/contracts-js",
   "dependencies": {
-    "@thirdweb-dev/contracts": "3.1.7"
+    "@thirdweb-dev/contracts": "3.1.8-0"
```
joaquim-verges, line 9: What does this change? Being able to override the trusted forwarder for pack?

Also can you run yarn build from the root package to get the new ABIs checked in?

kumaryash90: yes the initialize function is now accepting _trustedForwarders array, so the signature changed

joaquim-verges (approved):
looks good! ty @kumaryash90

PR #111, opened 2022-09-12
[SDK] Fix get owned for Signature Drop

nachoiacovino:
And any other contracts implementing Enumerable but not Supply


packages/sdk/src/core/classes/erc-721.ts
```diff
@@ -299,19 +299,28 @@ export class Erc721<
...
   public async getOwned(walletAddress?: string) {
-    return assertEnabled(this.query?.owned, FEATURE_NFT_SUPPLY).all(
-      walletAddress,
-    );
+    if (this.query?.owned) {
+      return this.query.owned.all(walletAddress)
+    } else {
+      const allNFTs = await this.getAll();
+      return (allNFTs || []).filter(({ owner }) =>
+        walletAddress === owner)
```
joaquim-verges, line 307: I think we need to lowercase this when comparing. Just in case

joaquim-verges (approved):

PR #139, opened 2022-09-16
Update Storage SDK

adam-maj:


packages/solana/test/mock/MockStorage.ts
```diff
@@ -63,113 +37,39 @@ export class MockStorage implements IStorage {
...
 
-  private async batchUploadProperties(metadatas: JsonObject[]): Promise<any> {
-    for (const file of metadatas) {
-      if (typeof file === "string") {
-        continue;
-      }
-      await this.uploadProperties(file);
-    }
-  }
+export function MockStorage(): ThirdwebStorage {
```
joaquim-verges, line 71: would it make sense to move this to the storage package? maybe as an internal class?

right now we have to duplicate this between solana / ts sdk

packages/solana/test/mock/MockStorage.ts
```diff
@@ -63,113 +37,39 @@ export class MockStorage implements IStorage {
...
+  const uploader = new MockStorageUploader();
+  const downloader = new MockStorageDownloader();
```
joaquim-verges, line 73: btw your mapping doesn't need to be declared top level. can just create a new one  in this method (I'd make it static too) and pass it in the mock uploader/downloader constructors. That way it's all nicely self contained

packages/storage/src/common/urls.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @internal
+ */
+export const DEFAULT_GATEWAY_URLS: GatewayUrls = {
+  // Note: Gateway URLs should have trailing slashes (we clean this on user input)
+  "ipfs://": [
+    "https://gateway.ipfscdn.io/ipfs/",
+    "https://cloudflare-ipfs.com/ipfs/",
+    "https://ipfs.io/ipfs/",
```
joaquim-verges, line 11: always found it error prone that you have to pass the `/ipfs/` at the end. so easy to forget

packages/storage/src/core/storage.ts
```diff
@@ -0,0 +1,113 @@
...
+  IpfsUploadBatchOptions,
+  IStorageDownloader,
+  IStorageUploader,
+  Json,
+  UploadOptions,
+} from "../types";
+import { StorageDownloader } from "./downloaders/storage-downloader";
+import { IpfsUploader } from "./uploaders/ipfs-uploader";
+
+export class ThirdwebStorage<T extends UploadOptions = IpfsUploadBatchOptions> {
```
joaquim-verges, line 19: make sure to add some docs / examples

packages/storage/src/types/data.ts
```diff
@@ -0,0 +1,55 @@
...
+    FileOrBufferSchema,
+    z.array(JsonSchema),
+  ]),
+);
+
+export const JsonObjectSchema = z.record(z.string(), JsonSchema);
+
+export type Json = JsonLiteral | FileOrBuffer | JsonObject | Json[];
+
+export type JsonObject = { [key: string]: Json };
```
joaquim-verges, line 55: I think i duplicated these in solana, can you check that we're importing from here in both ts SDK/solana?

packages/sdk/src/core/classes/delayed-reveal.ts
```diff
@@ -98,26 +101,29 @@ export class DelayedReveal<
...
 
-    const { baseUri: placeholderUri } = await this.storage.uploadMetadataBatch(
+    const placeholderUris = await this.storage.uploadBatch(
       [CommonNFTInput.parse(placeholder)],
-      0,
-      this.contractWrapper.readContract.address,
-      await this.contractWrapper.getSigner()?.getAddress(),
+      {
+        rewriteFileNames: {
+          fileStartNumber: 0,
```
joaquim-verges, line 108: niceee

joaquim-verges (commented):
looks so much cleaner! excited for this

joaquim-verges:
@jnsdls any idea why the e2e react test would timeout on mount?

packages/solana/test/mock/MockStorage.ts
adam-maj: Was thinking of doing this but left it since we had the duplicate from before. I think this generally makes sense - could be useful for other people testing?

Otherwise might be bleeding of our own implementations

packages/storage/src/common/urls.ts
adam-maj: I'm thinking of this as a generic string match/replace - issue is that for other schemes, there's not necessarily an obvious thing to add on top like that - and would like the replacement to be as explicit as possible for the user. 

Definitely might be nice to do some additional checks for the user though

packages/storage/src/core/storage.ts
adam-maj: Indeed - will put these in separate PR like Solana

adam-maj:
> @jnsdls any idea why the e2e react test would timeout on mount?

Trying to debug it now - it doesn't like the `FileOrBufferOrStringSchema` and is saying it's `undefined` - definitely something with `Buffer` on browser here, trying to try stuff out.

adam-maj:
Fixed!

joaquim-verges (approved):
Lgtm!

PR #146, opened 2022-09-19
init fixing cli

jnsdls:


packages/cli/src/common/processor.ts
```diff
@@ -194,18 +193,9 @@ export async function processProject(
...
-        encodeURIComponent(hashes[0].replace("ipfs://", "")),
-    );
-  } else {
-    url = new URL(THIRDWEB_URL + "/contracts/" + command);
-    for (let hash of hashes) {
-      url.searchParams.append("ipfs", hash.replace("ipfs://", ""));
-    }
+  const url = new URL(THIRDWEB_URL + "/contracts/" + command);
+  for (let hash of hashes) {
+    url.searchParams.append("ipfs", hash.replace("ipfs://", ""));
```
joaquim-verges, line 198: does the dashboard handle single contract deploy/releases if i pass just one `ipfs` query param?

PR #147, opened 2022-09-19
Add documentation for storage

adam-maj:


packages/storage/src/types/upload.ts
```diff
@@ -22,15 +35,36 @@ export type UploadProgressEvent = {
...
   gatewayUrls?: GatewayUrls;
+  /**
+   * Whether or not to replace any URLs with schemes with resolved URLs before upload
+   */
   uploadWithGatewayUrl?: boolean;
 };
 
 export type IpfsUploadBatchOptions = {
+  /**
+   * If specified, will rewrite file names to numbers for use on-chain
```
joaquim-verges, line 50: could be more explicit here - like "useful to use with NFT contracts that map token ids to files"

joaquim-verges (approved):

PR #163, opened 2022-09-20
[Storage] - Update expected input types for storage

adam-maj:


adam-maj:
Both deploy and release working with this version

jnsdls:
@adam-maj add a changeset to this please (can be patch because it should not break anything at all)

PR #166, opened 2022-09-21
Jw/lazymintabletwfeature

jarrodwatts:


packages/sdk/src/core/classes/erc-1155-lazymintable.ts
```diff
@@ -121,6 +121,7 @@ export class Erc1155LazyMintable implements DetectableFeature {
    *
    * @param metadatas - The metadata to include in the batch.
    * @param options - optional upload progress callback
+   * @twfeature ERC1155LazyMintable
```
joaquim-verges, line 124: we don't want this one here, instead we want ppl to use the `erc1155.lazyMint` function directly from the `Erc1155` class.

consider this internal

packages/sdk/src/core/classes/erc-1155.ts
```diff
@@ -263,6 +263,7 @@ export class Erc1155<
    * ]
    * await contract.erc1155.airdrop(tokenId, addresses);
    * ```
+   * @twfeature ERC1155
```
joaquim-verges, line 266: this requires both ERC1155 and Multicall - that's why i didnt tag it initially

joaquim-verges (changes requested):

PR #167, opened 2022-09-21
[SDK] Syntax fixes for react native support

joaquim-verges:


jnsdls:
why would these fix react native support?

joaquim-verges:
with the default bundler that expo sets you up with, these were consider syntax errors

jnsdls:
yikes

PR #179, opened 2022-09-22
[React] Expose simple provider and useSDK hook for solana

joaquim-verges:


packages/react/src/solana/providers/ThirdwebSDKProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+ *          </WalletProvider>
+ *      </ConnectionProvider>
+ * )};
+ * ```
+ */
+export const ThirdwebSDKProvider: FC<
+  PropsWithChildren<TWSolanaProviderProps>
+> = ({ children }) => {
+  const { connection } = useConnection();
+  const wallet = useWallet();
```
jnsdls, line 47: if we're doing this we should just do the full wrapping provider

jnsdls: it's an extra 4-5 LOC at this point and now we're already pulling in the deps etc

joaquim-verges: i had it as peer dep. but yeah

it would definitely make it simpler on the DX side, especially ppl who are not familiar with solana js

let me try it out

jnsdls (approved):
will clean it up further but this will work as a start

PR #180, opened 2022-09-22
Replace schemes after replacing files

adam-maj:


packages/storage/test/ipfs.test.ts
```diff
@@ -199,31 +199,6 @@ describe("IPFS", async () => {
     expect(json.description).to.equal("Uploading alone without a directory...");
   });
 
-  it("Should upload without directory if specified on class", async () => {
```
jnsdls, line 202: is this test no longer relevant or why was it removed?

jnsdls (commented):
just a question

PR #182, opened 2022-09-22
[SDK] - Add getMinimumNextBid function to marketplace

### LINKED ISSUE #87: [SDK] - Add "getMinimumBidValue" to marketplace
## Feature Description

Currently with the marketplace prebuilt contract there is no way to get the minimum bid a user must place to outbid the previous highest bid.

Let's add it!

## Goal

Add a function to the marketplace smart contract called `getMinimumNextBid`

The logic needs to be something like this:
```
if (auction has current highest bid) {
  return currentHighestBid + (bid buffer * highest bid)
}
else {
  return minimum bid value
}
```

This must take into account [Bid Buffers](https://portal.thirdweb.com/sdk/interacting-with-contracts/marketplace#auction-bid-buffers); i.e. the next bid must be a certain amount higher than the current highest bid.

jnebab:
fixes #87


.changeset/quick-peas-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/sdk": major
```
jnsdls, line 2:
```suggestion
"@thirdweb-dev/sdk": patch
```
this should not be `major` version but can be done under `patch` since it has no breaking changes or big new features

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -438,6 +438,26 @@ export class Marketplace implements UpdateableNetwork {
...
+  /**
+   * returns the minimum bid a user can place to outbid the previous highest bid
+   * @param listingId - the listing id of the auction
+   */
+  public async getMinimumNextBid(
+    listingId: BigNumberish,
+  ): Promise<BigNumberish> {
+    const currentAuction = this.auction;
+    const currentBidBuffer = await this.getBidBufferBps();
+    const minimumBid: BigNumberish = await this.getMinimumNextBid(listingId);
```
jnsdls, line 450: this seems like it's calling itself recursively? I assume you meant to call a function to determine the *absolute* minimum bid for the listing instead?

jnsdls (changes requested):
hey @jnebab, thanks for the PR!

I have a small change request to the changeset + a question about what I assume is an oversight in the implementation

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
jnebab: Hi @jnsdls, thanks for the feedback. I completely missed this. will push another commit to handle this

jnebab: Hi @jnsdls , I have made a new commit. I honestly not 100% with the minimum value but I assumed that if there's no current auction yet and no winning bid, means nothing has made a bid yet, so basically, I have to get the reservePrice and add bid buffer to it.

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -438,6 +438,29 @@ export class Marketplace implements UpdateableNetwork {
...
+  public async getMinimumNextBid(
+    listingId: BigNumberish,
+  ): Promise<BigNumberish> {
+    const currentAuction = this.auction;
+    const currentBidBuffer = await this.getBidBufferBps();
+    const winningBid = await currentAuction.getWinningBid(listingId);
+    const activeListing: AuctionListing = await currentAuction.getListing(
+      listingId,
+    );
+    if (currentAuction && winningBid) {
```
joaquim-verges, line 454: currentAuction should always be defined right? also if you move it to `MarketplaceAuction` you won't need this

joaquim-verges (changes requested):
thank you for your contribution!

approach looks good, would like the function to be moved to `MarketplaceAuction` though, so that the API is : `marketplace.auction.getMinimumNextBid(listingId)`

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -438,6 +438,29 @@ export class Marketplace implements UpdateableNetwork {
...
 
+  /**
+   * returns the minimum bid a user can place to outbid the previous highest bid
+   * @param listingId - the listing id of the auction
+   */
+  public async getMinimumNextBid(
+    listingId: BigNumberish,
+  ): Promise<BigNumberish> {
+    const currentAuction = this.auction;
+    const currentBidBuffer = await this.getBidBufferBps();
```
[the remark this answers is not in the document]
jnebab: Hi @joaquim-verges, do you want to include moving the `getBidBufferBps()` function to `marketplace-auction.ts`? Because it's been originally in the `marketplace.ts` class and also, not mentioned as part of the ticket. I don't mind doing it though if moving things such as this will be allowed.

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -438,6 +438,29 @@ export class Marketplace implements UpdateableNetwork {
     await this.contractWrapper.multiCall(encoded);
   }
 
+  /**
+   * returns the minimum bid a user can place to outbid the previous highest bid
+   * @param listingId - the listing id of the auction
+   */
+  public async getMinimumNextBid(
```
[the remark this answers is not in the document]
jnebab: `this.auction` is being use on other functions as well in this class. So I assumed, this was the appropriate place to add the function and also since the `getBidBufferBps` is also here. It made it more convenient for me to implement here basically.

gregfromstl:
Adding "fixes #87" to link this PR with the Issue

jnebab:
> Hey @jnebab, could you add "fixes #87" to the PR description so it links to the corresponding Issue?

Hi, done updating the description

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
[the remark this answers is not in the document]
joaquim-verges: yep, as mentioned above , i would move `getBidBufferBps()` into `MarketplaceAuction` but keep `Marketplace.getBidBufferBps()` exposed to avoid breaking the API.

The implementation of `Marketplace.getBidBufferBps()` would then just call into the implementation inside `MarketplaceAuction`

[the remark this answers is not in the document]
jnebab: hi @joaquim-verges, I made the change. Kindly check and let me know. Thanks much

packages/sdk/src/evm/core/classes/marketplace-auction.ts
```diff
@@ -459,6 +459,32 @@ export class MarketplaceAuction {
...
+    const winningBid = await this.getWinningBid(listingId);
+    const listing: AuctionListing = await this.getListing(listingId);
+    if (winningBid) {
+      const winningBidPrice = winningBid.currencyValue.value;
+      return winningBidPrice.add(currentBidBuffer.mul(winningBidPrice));
+    } else {
+      const price = listing.reservePrice;
+      return price.add(currentBidBuffer.mul(price));
+    }
+  }
```
jnsdls, line 486: hey @jnebab thx for doing this work, take a look at this suggestion to optimize the async calls a bit more

```suggestion
  public async getMinimumNextBid(
    listingId: BigNumberish,
  ): Promise<BigNumberish> {
    // we can fetch both of these at the same time using promise.all
    const [currentBidBuffer, winningBid] = await Promise.all([
      this.getBidBufferBps(),
      this.getWinningBid(listingId),
    ]);

    if (winningBid) {
      const winningBidPrice = winningBid.currencyValue.value;
      return winningBidPrice.add(currentBidBuffer.mul(winningBidPrice));
    } else {
      // we can delay getting the listing until we know there is no winning bid (since we don't need it for the calculation unless there is no winning bid)
      const listing = await this.getListing(listingId);
      const price = listing.reservePrice;
      return price.add(currentBidBuffer.mul(price));
    }
  }
```

jnsdls (commented):
overall looks good to me, small optimization is possible still though

jnsdls (approved):

PR #183, opened 2022-09-22
[Storage] - allow uploading the same file multiple times in uploadBatch if it is *the same file*

jnsdls:


jnsdls:
@adam-maj just tagged so u see it

nachoiacovino:
does this make it so it works the same as before or we're making it so we're only uploading it once and using the same URI for both?

jnsdls:
> does this make it so it works the same as before or we're making it so we're only uploading it once and using the same URI for both?

the second one of those ;)

nachoiacovino:
> > does this make it so it works the same as before or we're making it so we're only uploading it once and using the same URI for both?
> 
> the second one of those ;)

nice

PR #188, opened 2022-09-22
Remove JSON types

adam-maj:


joaquim-verges:
do we need a similar change on solana?

also add a changeset 🙏

jnsdls:
seems like this broke some things?

adam-maj:
> Details

Yes - the big number schema seems to be catching strings

PR #189, opened 2022-09-22
Add progress tracking to file upload

adam-maj:


packages/storage/src/core/storage.ts
```diff
@@ -206,7 +206,7 @@ export class ThirdwebStorage<T extends UploadOptions = IpfsUploadBatchOptions> {
 
     if (files.length) {
       // Upload all files that came from the object
-      const uris = await this.uploader.uploadBatch(files);
```
joaquim-verges, line 209: oof.. make sure that all other calls pass the options properly

joaquim-verges (approved):

packages/storage/src/core/storage.ts
adam-maj: This one is tricky - the way we used to do it is only pass through options (and onProgress) through to one of the calls on `batchUpload` because otherwise the progress would reset in the middle (since we upload files first, then the top-level object). So we may want to only pass through onProgress to the file upload.

Can keep this as a fix for now though

PR #191, opened 2022-09-22
get all offers for a listing

### LINKED ISSUE #88: [SDK] - Add "getOffers" to Marketplace
## Feature Description

Currently, there is no way to view all of the offers that have been made to a direct listing.

We should add this functionality so that the user can view all of the offers on a given listing ID.

## Goal

Add the capability to call:

```
marketplaceContract.getOffers(listingId)
```

That returns an array of `Offer`s.

### Info

This should be implemented using the [Contract Events](https://portal.thirdweb.com/sdk/advanced-features/contract-events) feature.

mykcryptodev:
fixes #88

### DIFF
--- a/packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
+++ b/packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
@@ -16,7 +16,7 @@ import { ListingType } from "../../enums";
 import { MarketplaceContractSchema } from "../../schema/contracts/marketplace";
 import { SDKOptions } from "../../schema/sdk-options";
 import { DEFAULT_QUERY_ALL_COUNT } from "../../types/QueryParams";
-import { AuctionListing, DirectListing } from "../../types/marketplace";
+import { AuctionListing, DirectListing, Offer } from "../../types/marketplace";
 import { MarketplaceFilter } from "../../types/marketplace/MarketPlaceFilter";
 import type { Marketplace as MarketplaceContract } from "@thirdweb-dev/contracts-js";
 import type ABI from "@thirdweb-dev/contracts-js/dist/abis/Marketplace.json";
@@ -294,6 +294,39 @@ export class Marketplace implements UpdateableNetwork {
    */
   public async getTimeBufferInSeconds(): Promise<BigNumber> {
     return this.contractWrapper.readContract.timeBuffer();
+  }
+
+  /**
+   * Get all the offers
+   *
+   * @remarks Fetch all the offers for a specified direct listing.
+   * @example
+   * ```javascript
+   * const offers = await marketplaceContract.getOffers(listingId);
+   * const firstOffer = offers[0];
+   * ```
+   *
+   * @param listingId - the id of the listing to fetch offers for 
+   */
+   public async getOffers(
+    listingId: BigNumberish,
+  ): Promise<(Offer)[]> {
+    // get all new offer events from this contract
+    const events = await this.events.getEvents("NewOffer");
+    // get only the events for this listing id
+    const listingEvents = events.filter((e) => e.data.args.listingId.eq(listingId));
+    // derive the offers from the events
+    const offers = listingEvents.map((e) : Offer  => {
+      return {
+        listingId: e.data.args.listingId,
+        buyerAddress: e.data.args.offeror,
+        quantityDesired: e.data.args.quantityWanted,
+        pricePerToken: (e.data.args.totalOfferAmount).div(e.data.args.quantity),
+        currencyValue: e.data.args.totalOfferAmount,
+        currencyContractAddress: e.data.args.currency,
+      };
+    });
+    return offers;
   }
 
   /** ******************************
--- a/.changeset/eight-doors-shave.md
+++ b/.changeset/eight-doors-shave.md
@@ -1,0 +1,5 @@
+---
+"@thirdweb-dev/sdk": patch
+---
+
+Get all offers for a listing


packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -296,6 +296,39 @@ export class Marketplace implements UpdateableNetwork {
...
+   *
+   * @param listingId - the id of the listing to fetch offers for 
+   */
+   public async getOffers(
+    listingId: BigNumberish,
+  ): Promise<(Offer)[]> {
+    // get all new offer events from this contract
+    const events = await this.events.getEvents("NewOffer");
+    // get only the events for this listing id
+    const listingEvents = events.filter((e) => e.data.listingId.eq(listingId));
```
jnsdls, line 317: @joaquim-verges correct me if I'm wrong but I think this might be able to be combined in the `getEvents` via a filter call directly on-chain (meaning having to query less total listings?)

jnsdls (commented):
thanks for the PR @mykcryptodev!

I think there is a slightly more optimal way to achieve the same result that I need @joaquim-verges to take a look at and confirm. Other than that this looks good to me.

Thanks for adding a test-case for it, also!

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -296,6 +296,39 @@ export class Marketplace implements UpdateableNetwork {
...
+   * ```javascript
+   * const offers = await marketplaceContract.getOffers(listingId);
+   * const firstOffer = offers[0];
+   * ```
+   *
+   * @param listingId - the id of the listing to fetch offers for 
+   */
+   public async getOffers(
+    listingId: BigNumberish,
+  ): Promise<(Offer)[]> {
```
joaquim-verges, line 313: you don't need the parenthesis around Offer here. Would you mind setup prettier so it applies the common formatting for you? (in vscode, settings, default formatter > prettier)

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -296,6 +296,39 @@ export class Marketplace implements UpdateableNetwork {
...
+    // get all new offer events from this contract
+    const events = await this.events.getEvents("NewOffer");
```
joaquim-verges, line 315: you can get typed event objects back here, which will help catch any type conversion issues:

`const events = await this.events.getEvents<NewOfferEventObject>("NewOffer");`

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -296,6 +296,39 @@ export class Marketplace implements UpdateableNetwork {
...
+    // get only the events for this listing id
+    const listingEvents = events.filter((e) => e.data.listingId.eq(listingId));
+    // derive the offers from the events
+    const offers = listingEvents.map((e): Offer  => {
+      return {
+        listingId: e.data.listingId,
+        buyerAddress: e.data.offeror,
+        quantityDesired: e.data.quantityWanted,
+        pricePerToken: (e.data.totalOfferAmount).div(e.data.quantityWanted),
+        currencyValue: e.data.totalOfferAmount,
```
joaquim-verges, line 325: this type is wrong - currencyValue is an object.

check out what we do in `common/marketplace.ts` in the `mapOffer()` function

joaquim-verges (commented):
thanks for the contribution! just a few comments

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
joaquim-verges: there is a way to setup a filter passing the specific listing id, but you'd need to replicate some of the low level parsing we're doing. Don't think it's worth it right now. We can go with this approach first and then optimize it later.

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
mykcryptodev: I pulled in the mapOffer function and am now using that (see [cad4099](https://github.com/thirdweb-dev/js/pull/191/commits/cad40990589bacd37850e0cd7aa11c2f59ecc33d))

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -296,6 +298,44 @@ export class Marketplace implements UpdateableNetwork {
...
+          this.contractWrapper.getProvider(),
+          BigNumber.from(listingId),
+          {
+            quantityWanted: e.data.quantityWanted,
+            pricePerToken: e.data.totalOfferAmount.div(e.data.quantityWanted),
+            currencyContractAddres: e.data.currency,
+            buyerAddress: e.data.offeror,
+            currency: e.data.currency,
+            listingId: e.data.listingId,
+          },
```
joaquim-verges, line 331: oof i didn't realize this offer parameter can be any, we should really type that. It's letting a few type errors go unnoticed:

the passed in `offer` object should have:

```ts
{
    quantityDesired: BigNumberish,
    pricePerToken: BigNumberish,
    currency: string,
    offeror: string,
    quantityWanted: BigNumberish
}
```
so no `buyerAddress` and no `currencyContractAddres`

feel free to add a type for this 🙏

packages/sdk/test/marketplace.test.ts
```diff
@@ -598,6 +598,35 @@ describe("Marketplace Contract", async () => {
...
+      );
+
+      // fetch all offers for the listing
+      sdk.updateSignerOrProvider(adminWallet);
+      const offers: Offer[] = await marketplaceContract.getOffers(
+        directListingId,
+      );
+      
+      // check that the offers are returned
+      assert.equal(offers.length, 2);
```
joaquim-verges, line 628: can you assert the expected values inside one of these offers? that should help alleviate the potential type/property name issues

joaquim-verges (commented):
really close! just a couple more comments

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
mykcryptodev: @joaquim-verges I thought the purpose of the `mapOffer` function was to take input of "any" and return that payload as type "Offer". If that's the case, wouldn't forcing the input to be of type "Offer" defeat the purpose?

joaquim-verges: The goal of mapOffer was to transform a contract struct to a more dev friendly Offer type. Now you're adding a way to map from a contract event to a Offer type.

So the inputs to mapOffer are not quite the output type

packages/sdk/test/marketplace.test.ts
mykcryptodev: Added tests in [2503be2](https://github.com/thirdweb-dev/js/pull/191/commits/2503be211bf48a7040bed17acc284fb139a11310)

packages/sdk/src/contracts/prebuilt-implementations/marketplace.ts
mykcryptodev: Forgive my n00bishness with TypeScript. 

Is what I did here in commit [1933294](https://github.com/thirdweb-dev/js/pull/191/commits/19332946d6b8f3146e62d03e101d031b01099e4d) what you were looking for?

I suspect it isn't because I had to add some optional keys for the newly added type to support backwards compatibility.

I was thinking about creating a new `mapOffer` function specifically for my use case and I'm happy to do that if it is a preferred path forward.

joaquim-verges (approved):
Thank you for your contribution @mykcryptodev ! and thank you for being patient with me :)

This is going to be super helpful for others, will be included in the next release!

Issue #81, opened 2022-09-08
[SDK] - Add ability to view and update owner of a contract


jarrodwatts:
## Feature Description

We have the [Ownable](https://portal.thirdweb.com/extensions/features/ownable) functionality on contract extensions and also in many of our prebuilt contracts.

Currently, there is no way to utilize this feature from the SDK.

## Goal

The user should be able to call 

`contract.getOwner` => Get the wallet address of the current owner.
`contract.setOwner` => Attempt to transfer ownership (requires the user be the owner of the smart contract0

jnsdls:
fixed in #194 thx @aeither!

will be released when #227 is merged

PR #194, opened 2022-09-23
[SDK] - Add ability to view and update owner of a contract

aeither:
Issue https://github.com/thirdweb-dev/js/issues/81


packages/sdk/src/contracts/prebuilt-implementations/edition-drop.ts
```diff
@@ -225,6 +228,19 @@ export class EditionDrop extends StandardErc1155<DropERC1155> {
...
+   * Get current owner of the contract
+   *
+   * @example
+   * ```javascript
+   * const owner = await contract.getOwner();
+   * ```
+   * @returns The owner address.
+   */
+  public async getOwner(): Promise<string> {
+    return this.owner.get();
```
joaquim-verges, line 241: the pattern we've been doing is directly accessing these common functions using the namespace.

`owner` is public so ppl can call `contract.owner.get()` and `contract.owner.set()`

no need to re-expose these at the top level

packages/sdk/src/core/classes/contract-owner.ts
```diff
@@ -22,14 +22,14 @@ export class ContractOwner<TContract extends Ownable>
...
    * @returns the owner address
    */
-  public async get(): Promise<string> {
+   public async get(): Promise<string> {
     return this.contractWrapper.readContract.owner();
   }
 
   /**
-   * set tje new owner of the contract
+   * set the new owner of the contract
```
joaquim-verges, line 30: 🙏

packages/sdk/src/contracts/prebuilt-implementations/edition-drop.ts
aeither: I deployed all the pre-built contracts with the dashboard to check in the explorer section to make sure it contains `setOwner` and `owner`.

- Edition
- Edition Drop
- NO - Marketplace
- Multiwrap
- NFT Collection
- NFT Drop
- Pack
- Signature Drop
- NO - Split
- NO - Token
- NO - Token Drop
- NO - Vote

Few contracts doesn't have owner so I implemented only on those that have.
I re-exposed at the top level as requested in the goals in the issue #81

packages/sdk/src/contracts/prebuilt-implementations/edition-drop.ts
```diff
@@ -225,6 +228,19 @@ export class EditionDrop extends StandardErc1155<DropERC1155> {
...
+  /**
+   * Get current owner of the contract
+   *
+   * @example
+   * ```javascript
+   * const owner = await contract.getOwner();
+   * ```
+   * @returns The owner address.
+   */
+  public async getOwner(): Promise<string> {
```
joaquim-verges, line 240: you don't need these top level functions. Let's just expose it via the namespace: `contract.owner.get()` directly

.changeset/eleven-walls-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/sdk": minor
```
joaquim-verges, line 2: Can you make this a patch version instead of minor?

```suggestion
"@thirdweb-dev/sdk": patch
```

joaquim-verges (approved):
thank you for your contribution @aeither ! merging this. will be available in the next release

PR #206, opened 2022-09-26
[SDK] - Relayer gas override

kumaryash90:


packages/sdk/src/core/classes/contract-wrapper.ts
```diff
@@ -450,6 +450,11 @@ export class ContractWrapper<
       gas = BigNumber.from(500000);
     }
 
+    // check for gas override in callOverrides
+    if (BigNumber.from(callOverrides?.gasLimit).gt(gas)) {
```
joaquim-verges, line 454: can calloverrides be undefined? I think it should always be there. 

`gasLimit` can be undefined though, and im not sure what happens if you call `BigNumber.from(undefined)` - might want to add a `if (callOverrdes.gasLimit && BigNumber.from(callOverrides?.gasLimit).gt(gas)` to be safe

kumaryash90: makes sense, updating

packages/sdk/src/core/classes/contract-wrapper.ts
```diff
@@ -450,6 +450,14 @@ export class ContractWrapper<
       gas = BigNumber.from(500000);
     }
 
+    // check for gas override in callOverrides
+    if (
+      callOverrides.gasLimit &&
+      BigNumber.from(callOverrides?.gasLimit).gt(gas)
+    ) {
+      gas = BigNumber.from(callOverrides?.gasLimit);
```
joaquim-verges, line 458: don't need the `?` here and above

joaquim-verges (approved):

PR #211, opened 2022-09-27
nft-drop prepareClaim -> public

jakeloo:


jnsdls:
@joaquim-verges we should look at just handling this predictably everywhere, we have a bunch of places where `prepareClaim` is public and others where it's not, I don't really understand the reasoning (if any) behind it

jnsdls (changes requested):
definitely needs a changeset

joaquim-verges:
Was private initially in pre builts, but had to make it public for extensions. Just need to delete the top level ones in pre built. They all call the same helper function under the hood.

People should use `contract(.erc721/.erc1155).claimConditions.prepareClaim(...)`

Lets not make this one public, should be deleted instead

joaquim-verges (changes requested):
Not needed

PR #231, opened 2022-09-28
[SOL] Add README.md

adam-maj:


packages/sdk/src/solana/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Install the latest version of the SDK with either `npm` or `yarn`:
+
+```shell
+npm install @thirdweb-dev/solana
+```
+
+```shell
+yarn add @thirdweb-dev/solana
+```
```
jnsdls, line 24: this is not correct anymore

packages/sdk/src/solana/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Quick Start
+
+The first thing to do to get started with Solana using thirdweb is to deploy a program. You can do this via the [Dashboard](https://thirdweb.com/dashboard), or via the SDK with the following snippet:
+
+```jsx
+import { ThirdwebSDK } from "@thirdweb-dev/solana";
```
jnsdls, line 31: nope

packages/sdk/src/solana/README.md
```diff
@@ -0,0 +1,69 @@
...
+  name: "New NFT",
+});
+const nft = await program.get(mintAddress);
+```
+
+## Learn More
+
+You can learn more about thirdweb and the Solana SDK with the following resources:
+
+- [Solana Docs](https://portal.thirdweb.com/solana)
```
jnsdls, line 67: is this still correct?

jnsdls (changes requested):
readme needs fixes, also not a fan of putting readmes in the src (I understand why ur doing it but I feel like there's better solutions here)

packages/sdk/src/solana/README.md
adam-maj: yea we just copy pasted from old version good catch

adam-maj: do yk why build is broken

jnsdls: no but looking at it

jnsdls (approved):
need to find a better way to have this readme but this is fine for now

jnsdls:
@adam-maj the side-effect of this is that test caches break every-time u update the readme now lol, seems pretty bad. why can't we have 3 readmes top-level? 1 general sdk entry point 1 for evm 1 for sol specific instead?

adam-maj:
> @adam-maj the side-effect of this is that test caches break every-time u update the readme now lol, seems pretty bad. why can't we have 3 readmes top-level? 1 general sdk entry point 1 for evm 1 for sol specific instead?

This is way better - I also didn't want to put it in src, we're just working in gather and wanted to fix it since docs are merged. Will change it rn on this branch

adam-maj:
@jnsdls build still breaking

PR #247, opened 2022-10-02
AppURI detection enabled [SDK]

furqanrydhan:
This pull request brings in AppURI extension detection in the SDK


packages/sdk/src/evm/constants/thirdweb-features.ts
```diff
@@ -60,6 +61,17 @@ export const FEATURE_METADATA = {
   features: {},
 } as const;
 
+export const FEATURE_APPURI = {
+  name: "AppURI",
+  namespace: "appuri",
```
adam-maj, line 66: Namespace reads wierdly. Maybe `contract.app` instead? Or if we want the URI part then capitalized is better like `contract.appUri`

packages/sdk/src/evm/contracts/smart-contract.ts
```diff
@@ -117,6 +120,13 @@ export class SmartContract<TContract extends BaseContract = BaseContract>
     return assertEnabled(this.detectOwnable(), FEATURE_OWNER);
   }
 
+  /**
+   * Set and get the appuri of the contract
+   */
+  get appuri(): ContractAppURI<AppURI> {
```
adam-maj, line 126: Same comment here about namespace or capitalization

packages/sdk/src/evm/contracts/smart-contract.ts
```diff
@@ -273,6 +283,12 @@ export class SmartContract<TContract extends BaseContract = BaseContract>
     return undefined;
   }
 
+  private detectAppURI() {
+    if (detectContractFeature<AppURI>(this.contractWrapper, "Ownable")) {
```
adam-maj, line 287: This should be detecting the `AppURI` feature - not `Ownable`

packages/sdk/src/evm/core/classes/appuri.ts
```diff
@@ -0,0 +1,51 @@
...
+import { ContractWrapper } from "./contract-wrapper";
+import type { AppURI } from "@thirdweb-dev/contracts-js";
+
+/**
+ * Have an official Application URI for this contract.
+ * @remarks Configure an official Application URI for this contract.
+ * @example
+ * ```javascript
+ * const contract = await sdk.getContract("{{contract_address}}");
+ * const appURI = await contract.appURI.get();
```
adam-maj, line 13: The namespace in this code example doesn't match the actual namespace that you've setup on the contract in the `smart-contract.ts` file (namespace is case sensitive)

packages/sdk/src/evm/core/classes/appuri.ts
```diff
@@ -0,0 +1,51 @@
...
+ * appURI = "ipfs://some_ipfs_hash";
+ *
+ * await contract.platformFee.set(appURI)
```
adam-maj, line 16: This should be `contract.appUri.set()` instead of `contract.platformFee.set()`

packages/sdk/src/evm/core/classes/appuri.ts
```diff
@@ -0,0 +1,51 @@
...
+  public async get() {
+    return await this.contractWrapper.readContract.appURI();
+  }
+
+  /**
+   * Set the appURI for the contract
+   * @param appURI - the uri to set (typically an IPFS hash)
+   */
+  public async set(appURI: string): Promise<TransactionResult> {
+    // TODO make it work with ContractMetadata
```
adam-maj, line 43: Are these 2 comments still relevant?

packages/sdk/src/evm/core/classes/appuri.ts
```diff
@@ -0,0 +1,51 @@
...
+  featureName = FEATURE_APPURI.name;
+  private contractWrapper;
+
+  constructor(contractWrapper: ContractWrapper<TContract>) {
+    this.contractWrapper = contractWrapper;
+  }
+
+  /**
+   * Get the appURI for the contract
+   * @returns the wallet address.
```
adam-maj, line 32: Return value isn't accurate (returns the `appUri` not the wallet address)

packages/sdk/src/evm/core/classes/appuri.ts
```diff
@@ -0,0 +1,51 @@
...
+   */
+  public async get() {
+    return await this.contractWrapper.readContract.appURI();
+  }
+
+  /**
+   * Set the appURI for the contract
+   * @param appURI - the uri to set (typically an IPFS hash)
```
adam-maj, line 40: Could also add a return value here for the transaction receipt of setting the appUri

packages/sdk/src/evm/constants/thirdweb-features.ts
furqanrydhan: I think all other namespaces were lowercase, assumed it was convention, if it isn't would prefer camelCase myself

packages/sdk/src/evm/contracts/smart-contract.ts
furqanrydhan: Same response

packages/sdk/src/evm/core/classes/appuri.ts
furqanrydhan: Fixed in latest commit

packages/sdk/src/evm/core/classes/appuri.ts
furqanrydhan: fixed

packages/sdk/src/evm/core/classes/appuri.ts
furqanrydhan: Correct, fixed

packages/sdk/src/evm/core/classes/appuri.ts
furqanrydhan: fixed in latest

packages/sdk/src/evm/contracts/smart-contract.ts
furqanrydhan: fixed in latest

packages/sdk/src/evm/core/classes/appuri.ts
furqanrydhan: fixed in latest

packages/sdk/src/evm/constants/thirdweb-features.ts
adam-maj: Looks like we actually only have one other feature namespace that has 2 words to face this problem, and in that case (lower in this file) we use camelCase - if you take a look at the platform fee namespace we do `platformFee` so `appUri` should be good here and then can change it in the actual getter function in `smart-contract.ts` file as well

packages/sdk/src/evm/core/classes/appuri.ts
adam-maj: Also remember to fix the namespace here if you change it above

packages/sdk/src/evm/constants/thirdweb-features.ts
adam-maj: Remember to change the namespace in the codesnippets you have elsewhere too

joaquim-verges: camelcase is the convention, we also have `contract.claimConditions.set()`

question is`contract.appUri` or `contract.appURI` ? i vote for the second to match the contract.

I also don't mind `contract.app.get()` tbh

packages/sdk/src/evm/contracts/smart-contract.ts
```diff
@@ -273,6 +283,12 @@ export class SmartContract<TContract extends BaseContract = BaseContract>
     return undefined;
   }
 
+  private detectAppURI() {
+    if (detectContractFeature<AppURI>(this.contractWrapper, "AppURI")) {
+      return new ContractAppURI(this.contractWrapper);
```
joaquim-verges, line 288: what did you decide regarding having the appURI inside ContractMetadata?

furqanrydhan: I want to have that as a fallback option but felt too complex for v1. Will add it in the next release but wanted to get this sdk out there

packages/sdk/src/evm/constants/thirdweb-features.ts
furqanrydhan: Will change this to appURI

packages/sdk/src/evm/contracts/smart-contract.ts
jnsdls: let's make this `appURI` as well then, in line with the comment above

jnsdls (commented):
just the naming change for `appURI` rest seems good to me

packages/sdk/src/evm/constants/thirdweb-features.ts
furqanrydhan: Fixed in latest

packages/sdk/src/evm/contracts/smart-contract.ts
furqanrydhan: Fixed in latest

jnsdls (approved):
LGTM - would be good to have a test for this but it's straight forward enough to not require it IMO

PR #253, opened 2022-10-04
[React] - allow passing in contractType in useContract()

jnsdls:


packages/react/src/evm/hooks/async/contracts.ts
```diff
@@ -149,7 +152,11 @@ export type UseContractResult<
  */
 export function useContract<
   TContract extends ValidContractInstance = SmartContract,
->(contractAddress: RequiredParam<ContractAddress>) {
+  TContractType extends ContractType = "custom",
```
joaquim-verges, line 155: wait so I can pass 2 conflicting types here?
like `useContract<NFTDrop, "marketplace">(...)` ?

packages/react/src/solana/hooks/program/useProgram.ts
```diff
@@ -92,13 +99,15 @@ export function programQuery<TProgramType extends ProgramType>(
  *
  * @public
  */
-export function useProgram<TProgramType extends ProgramType>(
-  address: RequiredParam<string>,
-  type?: TProgramType,
-) {
+export function useProgram<
+  TProgram extends ValidProgram,
+  TProgramType extends ProgramType | undefined = undefined,
```
joaquim-verges, line 104: hmmmm any way to not have the ProgramType generic?

packages/react/src/evm/hooks/async/contracts.ts
jnsdls: yes, but the runtime type would take precedence (because it actually changes behavior)

packages/react/src/solana/hooks/program/useProgram.ts
jnsdls: why?

PR #259, opened 2022-10-04
[CLI] Add npx thirdweb upload command

adam-maj:


packages/cli/src/cli/index.ts
```diff
@@ -141,6 +143,22 @@ $$$$$$\\   $$$$$$$\\  $$\\  $$$$$$\\   $$$$$$$ |$$\\  $$\\  $$\\  $$$$$$\\  $$$$
...
+    .action(async (path) => {
+      const storage = new ThirdwebStorage();
+      const uri = await upload(storage, path);
+      info(`Files stored at the following IPFS URI:`);
+      logger.info(chalk.blueBright(uri.toString()));
+
+      const url = storage.resolveScheme(uri);
+      info(`Open this link to view your upload:`);
+      logger.info(chalk.blueBright(url.toString()));
+      open(url.toString());
```
joaquim-verges, line 159: Maybe a bit weird to open the browser here?

packages/cli/src/storage/command.ts
```diff
@@ -0,0 +1,63 @@
...
+          uploadPath,
+        )} to upload.`,
+      );
+      process.exit(1);
+    }
+
+    const spin = spinner(
+      "Uploading directory to IPFS. This may take a while depending on file sizes.",
+    );
+    const uris = await storage.uploadBatch(files);
```
joaquim-verges, line 43: Not for right now, but adding progress % here would be clutch

adam-maj: Yea was thinking that - really wanted it but no upload progress for node :(

PR #263, opened 2022-10-05
[SDK] - align getProgram & make chainId getter available on EVM contracts

jnsdls:


packages/sdk/src/evm/contracts/prebuilt-implementations/marketplace.ts
```diff
@@ -36,7 +36,7 @@ import invariant from "tiny-invariant";
  * import { ThirdwebSDK } from "@thirdweb-dev/sdk";
  *
  * const sdk = new ThirdwebSDK("{{chainName}}");
- * const contract = sdk.getMarketplace("{{contract_address}}");
+ * const { contract } = sdk.getContract("{{contract_address}}", "marketplace");
```
joaquim-verges, line 39: Uuuh we destructuring now? Looks wrong

packages/sdk/src/evm/contracts/smart-contract.ts
```diff
@@ -148,19 +148,26 @@ export class SmartContract<TContract extends BaseContract = BaseContract>
     return assertEnabled(this.detectErc1155(), FEATURE_EDITION);
   }
 
+  private _chainId: number;
+  get chainId() {
+    return this._chainId;
```
joaquim-verges, line 153: Why not just `public chainId: number` directly?

packages/sdk/src/evm/core/interfaces/contract.ts
```diff
@@ -6,4 +6,5 @@ import { NetworkOrSignerOrProvider } from "../types";
 export interface UpdateableNetwork {
   onNetworkUpdated(network: NetworkOrSignerOrProvider): void;
   getAddress(): string;
+  chainId: number;
```
joaquim-verges, line 9: Oh, I see. Could have been getChainId() but whatevs

packages/sdk/src/evm/contracts/smart-contract.ts
jnsdls: because then u can technically set it

packages/sdk/src/evm/core/interfaces/contract.ts
jnsdls: yeah but since it's a getter only (u can never set it after constructing the class) a getter felt more correct (same with `getAddress()` to be fair) and `get getChainId()` seems excessive

packages/sdk/src/evm/contracts/prebuilt-implementations/marketplace.ts
jnsdls: wait yeah this is wrong the fuck am I doing lol

packages/sdk/src/evm/core/interfaces/contract.ts
jnsdls: tbh should probably do the same with address, but not trying to change it in this same ark of work rn (also would have to keep it backwards compat anyways)

PR #266, opened 2022-10-05
[React] Add hook to create a storage upload component

adam-maj:


packages/react/package.json
```diff
@@ -107,6 +107,7 @@
     "magic-sdk": "^8.1.1",
     "mime": "^3.0.0",
     "react-cool-dimensions": "^2.0.7",
+    "react-dropzone": "^14.2.2",
```
jnsdls, line 110: oof do we need this, would like to not force dependencies like this on devs

adam-maj: It's a nicer DX - also bundle size is quite small, checked it before I added

adam-maj: We can phase it out over time, this is a minimal flow

PR #279, opened 2022-10-08
Command Updates (detect -> build) + helper text

furqanrydhan:
Migrated detect command to build, deprecated detect (will leave it in the CLI for the next month). Also updated the helper text for multiple commands to see if they feel better


packages/cli/src/cli/index.ts
```diff
@@ -145,8 +141,8 @@ $$$$$$\\   $$$$$$$\\  $$\\  $$$$$$\\   $$$$$$$ |$$\\  $$\\  $$\\  $$$$$$\\  $$$$
 
   program
     .command("upload")
-    .description("Upload any file or directory using decentralized storage")
-    .argument("[upload-path]", "path to file or directory to upload")
+    .description("Upload any file or directory to decentralized storage (IPFS)")
```
adam-maj, line 144: This won't always be just IPFS - but this is fine for now

packages/cli/src/cli/index.ts
```diff
@@ -73,9 +73,7 @@ $$$$$$\\   $$$$$$$\\  $$\\  $$$$$$\\   $$$$$$$ |$$\\  $$\\  $$\\  $$$$$$\\  $$$$
 
   program
     .command("build")
-    .description(
-      "Compile contracts and detect implemented thirdweb contract extensions",
-    )
+    .description("Compile contract and detect thirdweb contract extensions")
```
adam-maj, line 76: `contract` -> `contracts` (this can be used on many contracts at once).

adam-maj: `build` also still feels like the wrong command here, but `detect` definitely wasn't right either, so this is fine for now. `info`, `compile` are other ideas

furqanrydhan: Why does build feel wrong? info / detect don't make sense. I can see compile as the function but feels like in the context of an app build is more in line with what people think about

PR #283, opened 2022-10-10
[WIP] app-uri

adam-maj:


packages/cli/src/storage/command.ts
```diff
@@ -33,22 +31,24 @@ export async function upload(
...
-        `npx thirdweb@latest upload path/to/file.jpg`,
-      )}\n`,
-    );
-    process.exit(1);
+    //TODO move up
+    /*    logger.error(*/
+    /*`Please pass a path to a file or directory to upload with the following format:\n   ${chalk.blueBright(*/
+    /*`npx thirdweb@latest upload path/to/file.jpg`,*/
+    /*)}\n`,*/
+    /*);*/
```
joaquim-verges, line 39: what was wrong with this? should clean up the commented code if not needed anymore

packages/cli/src/deploy/app.ts
```diff
@@ -0,0 +1,44 @@
...
+  const hasNPM = new NPMDetector().matches(".");
+
+  if (possibleProjects.length === 0) {
+    throw new Error("No supported project detected");
+  }
+
+  try {
+    if (hasYarn) {
+      console.log("yarn");
+      //run yarn-buld
```
joaquim-verges, line 31: cleanup and remove log

packages/cli/src/create/helpers/run-command.ts
```diff
@@ -1,11 +1,26 @@
 import spawn from "cross-spawn";
 
-export async function runCommand(command: string, args: string[]) {
+export async function runCommand(
+  command: string,
+  args: string[],
+  logging: boolean = true,
```
joaquim-verges, line 6: should prob not be true by default?

furqanrydhan: yeah good point i'll make it false

joaquim-verges:
@furqanrydhan remember to change the changeset to patch as well

.changeset/real-mugs-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/sdk": minor
```
jnsdls, line 2:
```suggestion
"@thirdweb-dev/sdk": patch
```

jnsdls (changes requested):
just change the bump to patch and address joaquim's logging -> false by default and then let's ship this

jnsdls (approved):
LGTM

PR #306, opened 2022-10-13
Add deploy event listener

adam-maj:


packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -65,6 +68,7 @@ export class ContractDeployer extends RPCConnectionHandler {
    * should never be accessed directly, use {@link ContractDeployer.getRegistry} instead
    */
   private _registry: Promise<ContractRegistry> | undefined;
+  public events: FactoryEvents | undefined;
```
joaquim-verges, line 71: this kinda sucks no?

how about having the `addDeployListener()` directly on deployer, that way you don't have to deal with the async-ness of the registry/factory ?

PR #307, opened 2022-10-13
[SDK] Fix multicall txns in a gasless setup

kumaryash90:
Issue ref: https://github.com/thirdweb-dev/ozdefender-autotask/issues/4#issuecomment-1272880337

### DIFF
--- a/packages/sdk/test/evm/gasless.test.ts
+++ b/packages/sdk/test/evm/gasless.test.ts
@@ -1,5 +1,6 @@
 /* eslint-disable turbo/no-undeclared-env-vars */
 import { ThirdwebSDK } from "../../src/evm";
+import { assert, expect } from "chai";
 import { ethers, Wallet } from "ethers";
 
 const RPC_URL = "https://rpc-mumbai.maticvigil.com/";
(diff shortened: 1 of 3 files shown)


packages/sdk/test/evm/gasless.test.ts
```diff
@@ -47,4 +48,43 @@ describe("Gasless Forwarder", async () => {
 
     await pack.open(0, 0);
   });
+
+  it.skip("should mint batch via multicall tx with relayer", async () => {
+    const NFT_COLLECTION_ADDRESS = "0x24A5aB2878B63716B001aa9AbE816c2662192B12";
+    const provider = ethers.getDefaultProvider(RPC_URL);
+    // const wallet = Wallet.createRandom().connect(provider);
+    const wallet = new Wallet("<PRIVATE KEY>", provider);
```
jnsdls, line 56: since ur doing gasless here u could just generate a fresh wallet via `ethers.wallet.createRandom()`? right (the wallet doesn't need any funds)

packages/sdk/test/evm/gasless.test.ts
```diff
@@ -47,4 +48,43 @@ describe("Gasless Forwarder", async () => {
...
+    const sdk = new ThirdwebSDK(wallet, {
+      gasless: {
+        openzeppelin: {
+          relayerUrl: "<RELAYER URL>",
```
jnsdls, line 60: now this I don't think will be feasible in unit tests without lots of work... would it be possible to instead intercept the transaction and see if the packed solidity is in there as expected correctly?

packages/sdk/test/evm/gasless.test.ts
kumaryash90: Had to test with admin controlled functions, so needed the deployer's key. Removed the key after testing and marked the test as `skip`.

packages/sdk/test/evm/gasless.test.ts
kumaryash90: Yeah, couldn't test locally. Used a real URL, removed after testing.

joaquim-verges (approved):

PR #312, opened 2022-10-13
[Solana] Add quantity to mint additional supply

adam-maj:


packages/sdk/src/solana/programs/nft-collection.ts
```diff
@@ -175,54 +179,16 @@ export class NFTCollection {
...
-      }
+    const originalEdition = toNftOriginalEdition(
+      toOriginalEditionAccount(originalEditionAccount),
+    );
 
-      editionMarkerNumber++;
-    }
-
-    return BigInt(totalSupply);
+    // Add one to supply to account for the master edition
```
joaquim-verges, line 190: I think we need to check if the token actually exists too, could have been burned so it's not always guaranteed to have supply 1

packages/sdk/src/solana/programs/nft-collection.ts
```diff
@@ -344,9 +310,9 @@ export class NFTCollection {
    * const address = await program.mintAdditionalSupply(nftAddress);
    * ```
    */
-  async mintAdditionalSupply(nftAddress: string) {
+  async mintAdditionalSupply(nftAddress: string, amount: number = 1) {
```
joaquim-verges, line 313: Let's make this param required instead of default to 1

joaquim-verges: Also let's return a number here instead of bigint

PR #315, opened 2022-10-17
Fix vite

furqanrydhan:
Trying to fix vite builds, moving multihash to non lazy seemed to help + cbor-x instead of cbor-web fixed the main set of issues. 

Have no idea on regression or other frameworks


packages/sdk/package.json
```diff
@@ -113,6 +113,7 @@
     "bn.js": "^5.2.1",
     "bs58": "^5.0.0",
     "cbor-web": "^8.1.0",
+    "cbor-x": "^1.4.1",
```
joaquim-verges, line 116: cbor-x looks pretty good, we can remove cbor-web then right?

packages/sdk/src/evm/common/feature-detection.ts
```diff
@@ -294,14 +295,15 @@ async function extractIPFSHashFromBytecode(
   );
 
   // load these lazily to avoid loading them when they are not needed
-  const [cbor, multiHashes] = await Promise.all([
-    import("cbor-web"),
-    import("multihashes"),
-  ]);
+  const [cbor] = await Promise.all([import("cbor-x")]);
```
joaquim-verges, line 298: Prob don't need a promise all anymore here. cbor-x allows to import just the decoding too:

`import { decode } from 'cbor-x/decode'`

packages/sdk/src/evm/common/feature-detection.ts
```diff
@@ -23,6 +23,7 @@ import {
 import { ExtensionNotImplementedError } from "./error";
 import { ThirdwebStorage } from "@thirdweb-dev/storage";
 import { BaseContract, BigNumber, ethers } from "ethers";
+import multihashes from "multihashes";
```
joaquim-verges, line 26: weird that this changed anything

packages/sdk/package.json
furqanrydhan: done in latest

packages/sdk/src/evm/common/feature-detection.ts
furqanrydhan: removed it and moved cbor to top

packages/sdk/src/evm/common/feature-detection.ts
furqanrydhan: agree, i'd love someone else to verify this was needed.

jnsdls:
going to take this on this afternoon

PR #318, opened 2022-10-17
[React] Add useMinimumNextBid hook

nachoiacovino:
Also changed the invariant for other marketplace hooks, seems weird we hardcode listingId 0 if the listing is invalid, so I did an invariant


jnsdls (approved):
LGTM

packages/react/src/evm/hooks/async/marketplace.ts
```diff
@@ -169,7 +170,8 @@ export function useWinningBid(
     ),
     () => {
       invariant(contract, "No Contract instance provided");
-      return contract.auction.getWinningBid(BigNumber.from(listingId || 0));
+      invariant(listingId, "No listing id provided");
```
jnsdls, line 173: same question about `0`

jnsdls (approved):
should be good

PR #319, opened 2022-10-17
[SDK, React] Expose total claimed/unclaimed supply for erc721 contracts

joaquim-verges:


packages/react/src/evm/hooks/async/drop.ts
```diff
@@ -94,18 +94,14 @@ export function useClaimedNFTs(
...
     cacheKeys.contract.nft.drop.totalUnclaimedSupply(contractAddress),
     () => {
-      invariant(contract, "No Contract instance provided");
-      // TODO make this work for custom contracts (needs ABI change)
-      invariant(
-        "totalUnclaimedSupply" in contract && contract.totalUnclaimedSupply,
-        "Contract instance does not support fetching unclaimed supply",
-      );
-      return contract.totalUnclaimedSupply();
+      invariant(erc721, "No ERC721 Contract instance provided");
```
nachoiacovino, line 101: we do it a little bit different everywhere else doing if (erc721) and doing an invariant otherwise but this works the same

nachoiacovino (approved):
crying from happiness rn

PR #324, opened 2022-10-19
Update storage download on fail

adam-maj:


packages/storage/src/common/utils.ts
```diff
@@ -114,16 +114,16 @@ export function replaceSchemeWithGatewayUrl(
...
   const schemeGatewayUrls = scheme ? gatewayUrls[scheme] : [];
 
-  if (!scheme) {
-    return uri;
-  }
-
-  if (index > schemeGatewayUrls.length) {
+  if ((!scheme && index > 0) || index >= schemeGatewayUrls.length) {
     throw new Error(
       "[GATEWAY_URL_ERROR] Failed to resolve gateway URL - ran out of gateway URLs to try.",
```
joaquim-verges, line 119: what if you just return undefined here instead of throwing and re-catching your own error?

joaquim-verges (commented):
looks like all the tests broke :p

joaquim-verges:
@adam-maj add a changeset too pls - let's get this merged today

joaquim-verges (approved):
make sure you run some tests with this on the dashboard/test script?

PR #333, opened 2022-10-20
[SDK] Support new Drop claim conditions mechanics

joaquim-verges:


packages/sdk/src/evm/common/legacy.ts
```diff
@@ -0,0 +1,55 @@
...
+import IThirdwebContractABI from "@thirdweb-dev/contracts-js/dist/abis/IThirdwebContract.json";
+import { ethers } from "ethers";
+
+// TODO (cc) use this everywhere
+export async function isPrebuilt(
+  contractWrapper: ContractWrapper<any>,
+  contractType: string,
+  maxVersion: number,
+): Promise<boolean> {
+  if (hasFunction<IThirdwebContract>("contractType", contractWrapper)) {
```
adam-maj, line 13: Should you be checking if the contract has the `contractVersion` function as well here since you're using it below

packages/sdk/src/evm/common/legacy.ts
```diff
@@ -0,0 +1,55 @@
...
+    try {
+      const [type, version] = await Promise.all([
+        ethers.utils.toUtf8String(
+          await contractWrapper.readContract.contractType(),
+        ),
+        await contractWrapper.readContract.contractVersion(),
+      ]);
+      const nameMatch = type.includes(contractType);
+      const versionMatch = maxVersion ? maxVersion <= version : true;
```
adam-maj, line 22: Little edge case, if you put 0 as `maxVersion` for this function the version match will still return true. Also we would probably want `maxVersion` to be a `?` type in the function signature if we're checking for undefined here

packages/sdk/src/evm/common/sharded-merkle-tree.ts
```diff
@@ -73,17 +79,42 @@ export class ShardedMerkleTree {
...
+            convertQuantityToBigNumber(entry.maxClaimable, tokenDecimals),
+          ],
+        );
+      case SnapshotFormatVersion.V2:
+        return utils.solidityKeccak256(
+          ["address", "uint256", "uint256", "address"],
+          [
+            entry.address,
+            convertQuantityToBigNumber(entry.maxClaimable, tokenDecimals),
+            convertQuantityToBigNumber(entry.price, 18), // TODO (cc) use currencyDecimals
```
adam-maj, line 106: I'm sure you remember this one but looks important so might as well call more attention to it (the TODO)

packages/sdk/src/evm/common/sharded-merkle-tree.ts
```diff
@@ -73,17 +79,42 @@ export class ShardedMerkleTree {
...
-    return hashLeafNode(
-      entry.address,
-      utils.parseUnits(entry.maxClaimable, tokenDecimals),
-    );
+  static hashEntry(
+    entry: SnapshotEntry,
+    tokenDecimals: number,
+    snapshotFormatVersion: SnapshotFormatVersion,
+  ): string {
+    // TODO (cc) hash price and currency
```
adam-maj, line 87: Are these still TODO?

packages/sdk/src/evm/common/sharded-merkle-tree.ts
adam-maj: Shouldn't be too bad to instantiate an IERC20 here to get the decimals

packages/sdk/src/evm/contracts/index.ts
```diff
@@ -198,6 +214,9 @@ export const NFTDropInitializer = {
   getAbi: async () =>
     (await import("@thirdweb-dev/contracts-js/dist/abis/DropERC721.json"))
       .default,
+  getV3Abi: async () =>
```
adam-maj, line 217: How come its `V3` for `DropERC721` but `V2` for the others

packages/sdk/src/evm/contracts/prebuilt-implementations/vote.ts
```diff
@@ -319,7 +319,7 @@ export class Vote implements UpdateableNetwork {
       this.contractWrapper.readContract.votingDelay(),
       this.contractWrapper.readContract.votingPeriod(),
       this.contractWrapper.readContract.token(),
-      this.contractWrapper.readContract.quorumNumerator(),
+      this.contractWrapper.readContract["quorumNumerator()"],
```
adam-maj, line 322: I think this will break this function

Shouldn't it be `this.contractWrapper.readContract["quorumNumerator()"]()` (you're missing the parentheses at the end to call the function)

packages/sdk/src/evm/core/classes/erc-1155-lazymintable.ts
```diff
@@ -232,17 +230,10 @@ export class Erc1155LazyMintable implements DetectableFeature {
...
-        return false;
-      }
-    } else {
-      return false;
-    }
+    const info = await getPrebuiltInfo(
+      this.contractWrapper.readContract.address,
+      this.contractWrapper.getProvider(),
+    );
+    return info && info.type === "DropERC1155" && info.version <= 2;
```
adam-maj, line 237: Should this be `<3` instead of `<=2` for version - do we do incremental version changes like `2.0.1` on contracts (assuming not, but figured I would mentioned)

packages/sdk/src/evm/common/legacy.ts
joaquim-verges: was not using this actually, removed

packages/sdk/src/evm/common/legacy.ts
joaquim-verges: yeah good catch

joaquim-verges: but ended up removing this function entirely

packages/sdk/src/evm/common/sharded-merkle-tree.ts
joaquim-verges: yeah i do need to do that - thanks for the reminder. the async part is annoying that's why i didn't do it immediately

packages/sdk/src/evm/contracts/index.ts
joaquim-verges: following the version of the contract

packages/sdk/src/evm/contracts/prebuilt-implementations/vote.ts
joaquim-verges: great catch thank you!

packages/sdk/src/evm/core/classes/erc-1155-lazymintable.ts
joaquim-verges: yeah, we don't do decimals for versions, but good point

PR #339, opened 2022-10-21
[React] Docs update

nachoiacovino:
Found some docs that were wrong but especially the deprecation notices, those didn't make sense


jnsdls (approved):
LGTM good finds

PR #341, opened 2022-10-21
Update RPC from alchemy to public endpoints to avoid alchemy rate-limit

atharvadeosthale:
Alchemy RPC used in the SDK is facing issues and being constantly rate-limited, rolled back to using public RPC for networks in the mean-time


joaquim-verges:
thank you @atharvadeosthale ! we decided to handle this differently in the end - see #342

PR #344, opened 2022-10-24
resolveReleasesFromAddress -> resolveContractUriFromAddress

nachoiacovino:


packages/sdk/src/evm/common/feature-detection.ts
```diff
@@ -246,6 +246,7 @@ export async function resolveContractUriFromAddress(
   const bytecode = await provider.getCode(address);
   if (bytecode === "0x") {
     const chain = await provider.getNetwork();
+    console.log({address, provider, chain});
```
joaquim-verges, line 249: remove log?

packages/sdk/src/evm/core/classes/contract-publisher.ts
```diff
@@ -195,13 +195,13 @@ export class ContractPublisher extends RPCConnectionHandler {
...
    */
-  public async resolveReleasesFromAddress(address: string) {
+  public async resolveContractUriFromAddress(address: string) {
     const contractUri = await resolveContractUriFromAddress(
       address,
       this.getProvider(),
     );
     invariant(contractUri, "Could not resolve contract URI from address");
-    return await this.resolvePublishMetadataFromCompilerMetadata(contractUri);
+    return contractUri;
```
joaquim-verges, line 204: can you summarize what the problem was @nachoiacovino ?

nachoiacovino: we weren't passing the provider as before so we were hitting the provider from polygon always instead of hitting the provider from the deployed contract for getting contractUri, so now we're calling this function with the deployed contract chainId then getting releases with the polygon sdk

joaquim-verges: got it that makes sense

joaquim-verges (approved):

PR #353, opened 2022-10-26
[Multi] Add useUpdateCreator hook, fix CreatorInput schema

nachoiacovino:


packages/sdk/src/solana/classes/helpers/creators-helper.ts
```diff
@@ -9,18 +9,19 @@ export function enforceCreator(
   creators: CreatorInput[] = [],
   owner: PublicKey,
 ): Creator[] {
+  console.log("enforceCreator", creators, owner);
```
joaquim-verges, line 12: remove log

packages/sdk/src/solana/types/programs/index.ts
```diff
@@ -27,7 +27,7 @@ export const CommonContractOutputSchema = CommonContractSchema.extend({
  */
 export const CreatorInputSchema = z.object({
   address: z.string(),
-  sharePercentage: z.number(),
+  share: z.number(),
```
joaquim-verges, line 30: only annoying thing is that now its not clear if the unit is bps or percentage. Maybe we can add a transform here to ensure the value is between 0 and 100 ? you can copy what we did for BasisPointsSchema

nachoiacovino: used the PercentSchema we already had

joaquim-verges (approved):
lg

PR #385, opened 2022-11-16
Add retroactive creator and royalty update option

adam-maj:


packages/sdk/src/solana/programs/nft-collection.ts
```diff
@@ -427,35 +430,131 @@ export class NFTCollection {
...
+          .update({
+            nftOrSft: await this.getCollection(),
+            creators: enforceCreator(
+              creators,
+              this.metaplex.identity().publicKey,
+            ),
+          }),
+      );
+
+      // TODO: Don't use a get all here, all we need is mint addresses :(
```
joaquim-verges, line 455: we should break getAll() into 2 functions tbh, but yeah part of the optimization work

joaquim-verges (approved):
lg

packages/sdk/src/solana/programs/nft-collection.ts
adam-maj: Doing this in my other branch w/ totalCount

PR #388, opened 2022-11-17
Add support for total supply and pagination on get all

adam-maj:


packages/sdk/src/core/schema/QueryParams.ts
```diff
@@ -1,19 +1,25 @@
...
-  /**
-   * the index to start from (default: 0)
-   */
-  start?: number;
-  /**
-   * how many items to return (default: 100)
-   */
-  count?: number;
-}
+export type QueryAllParams = Exclude<
```
joaquim-verges, line 22: what are we excluding here?

packages/sdk/src/solana/programs/nft-drop.ts
```diff
@@ -166,34 +170,45 @@ export class NFTDrop {
...
    * ```
    */
-  async getAll(): Promise<NFT[]> {
+  async getAll(queryParams?: QueryAllParams): Promise<NFT[]> {
+    const parsedQueryParams = QueryAllParamsSchema.parse(queryParams);
+
     // TODO: Add pagination to get NFT functions
     const info = await this.getCandyMachine();
-    const claimed = await this.getAllClaimed();
+    const claimed = await this.getAllClaimed(parsedQueryParams);
```
joaquim-verges, line 178: this wont work i think :( coz you're paginating the get all, then paginating the lazy minted items and "hoping" that the 2 slices will match but there's no guarantee... esspecially since claiming mints a random lazyminted item in the list

packages/sdk/src/core/schema/QueryParams.ts
adam-maj: Excluding the possibility of undefined from the query params

packages/sdk/src/solana/programs/nft-drop.ts
adam-maj: Hmm I have an idea

packages/sdk/src/core/schema/QueryParams.ts
jnsdls: just don't `.default()` inside the schema object and you get the same thing

packages/sdk/src/solana/programs/nft-drop.ts
adam-maj: Replaced it with a new method which should solve this - first you get all the claimed NFTs you can and fill the beginning of count w/ it, then any remaining count gets filled with unclaimed

packages/sdk/src/core/schema/QueryParams.ts
adam-maj: Yea but the default is intentionally there

packages/sdk/src/solana/programs/nft-drop.ts
```diff
@@ -172,44 +172,51 @@ export class NFTDrop {
...
+          },
+          owner: PublicKey.default.toBase58(),
+          supply: 0,
+          type: "metaplex",
+        } as NFT;
+      }),
+    );
+
+    // Always return claimed NFTs first, and then fill remaining query count with unclaimed NFTs
+    return [...claimedNfts, ...unclaimedNfts];
```
joaquim-verges, line 219: hmmm this means that the list will get shuffled around as they get claimed... maybe thats ok though. Eventually i think we're gonna end up separating claimed vs unclaimed nfts in 2 separate tabs cc @jnsdls

packages/sdk/src/solana/programs/nft-drop.ts
```diff
@@ -172,44 +172,51 @@ export class NFTDrop {
...
+      if (!isClaimed) {
+        unclaimedItems.push(item);
+      }
+    });
+
+    // Only fill the remaining count left over after claimed NFTs with unclaimed NFTs
+    const startIndex = Math.max(0, parsedQueryParams.start - totalClaimed);
+    const endIndex = Math.max(
+      0,
+      startIndex + parsedQueryParams.count - claimedNfts.length,
```
joaquim-verges, line 199: can you add a couple of tests for this to make sure we don't break it as we touch this?

PR #394, opened 2022-11-18
[CLI] - Add support for Solana with --solana flag

jarrodwatts:
This is Avneesh's work - just me committing it cuz he had some local issues


packages/cli/src/cli/index.ts
```diff
@@ -62,6 +62,7 @@ $$$$$$\\   $$$$$$$\\  $$\\  $$$$$$\\   $$$$$$$ |$$\\  $$\\  $$\\  $$$$$$\\  $$$$
       `Explicitly tell the CLI to bootstrap the app using pnpm`,
     )
     .option("--framework [name]", `The preferred framework.`)
+    .option("--solana", `Initialize as a Solana project.`)
```
jnsdls, line 65: this should likely only work in combination with `--app`

`npx thirdweb create --app --solana` makes sense 

`--solana` top level (same as `--app` seems wrong in terms of hierarchy

packages/cli/src/create/command.ts
```diff
@@ -15,6 +15,7 @@ let projectPath: string = "";
 let framework: string = "";
 let language: string = "";
 let baseContract: string = "";
+let chain: string = "";
```
jnsdls, line 18: if this is either `solana` or empty then why not make it a boolean value? `isSolana`

packages/cli/src/create/command.ts
```diff
@@ -123,16 +128,41 @@ export async function twCreate(pPath: string = "", options: any) {
...
   if (!options.template) {
+    if (projectType === "app" && !chain) {
+      const res = await prompts({
+        type: "select",
+        name: "chain",
+        message: CREATE_MESSAGES.chain,
+        choices: [
+          // For EVM, value is nothing because the repo names don't have "evm" in them.
+          { title: "EVM", value: "" },
+          { title: "Solana", value: "solana" },
```
jnsdls, line 139: why not give `evm` a value, too? also are users going to know what "EVM" is?

jnsdls: and btw if we're doing that then `--evm` should work, too to shortcut this

packages/cli/src/create/helpers/create-app.ts
```diff
@@ -43,7 +45,9 @@ export async function createApp({
       process.exit(1);
     }
   } else if (framework) {
-    frameworkPath = `${framework}-${language || "javascript"}-starter`;
+    frameworkPath = `${framework}-${language || "javascript"}-${
+      chain && "solana-"
+    }starter`;
```
jnsdls, line 50: we should *explicitly* check for chain === solana in this case, because I would assume that this would be the same thing for other chains down the line then (also again ,`evm` should have it's own "chain" value

jnsdls (changes requested):
some things to clean up, in general a good start though

jnsdls:
superseded by #403

PR #402, opened 2022-11-22
Expose sdk.fromWallet and wallet classes

adam-maj:


packages/sdk/src/evm/wallets/core/aws.ts
```diff
@@ -0,0 +1,85 @@
...
+        );
+      }
+
+      this.cachedSigner = new ethers.Wallet(privateKey, provider);
+      return this.cachedSigner;
+    } catch (error) {
+      // For a list of exceptions thrown, see
+      // https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html
+      throw error;
+    }
```
jnsdls, line 83: assuming we'll add handling of some sort here? If not why even wrap this all in a try catch in the first place?

packages/sdk/src/evm/core/sdk.ts
```diff
@@ -39,6 +40,48 @@ import invariant from "tiny-invariant";
...
+   * });
+   * const sdk = await ThirdwebSDK.fromWallet(wallet, "mainnet");
+   * ```
+   *
+   * @param wallet - the implementation of the AbstractWallet class to use for signing
+   * @param network - the network (chain) to connect to (e.g. "mainnet", "rinkeby", "polygon", "mumbai"...) or a fully formed RPC url
+   * @param options - the SDK options to use
+   * @returns an instance of the SDK
+   * @param storage - optional storage implementation to use
+   */
```
jnsdls, line 66: lets mark this `@beta` for now

jnsdls: as in the entire fn, also let's group the params have @returns below

jnsdls (changes requested):
couple of changes, couple of questions that we should think about and discuss

packages/sdk/src/evm/wallets/core/aws.ts
adam-maj: Might as well remove - just AWS default copy-paste

packages/sdk/src/evm/wallets/core/async.ts
```diff
@@ -0,0 +1,41 @@
...
+    this.getPrivateKey = options.getPrivateKey;
+  }
+
+  async getSigner(
+    provider?: ethers.providers.Provider,
+  ): Promise<ethers.Signer> {
+    const privateKey = await this.getPrivateKey();
+    return new ethers.Wallet(privateKey, provider);
+  }
+}
```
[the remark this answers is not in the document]
adam-maj: This one would be for all the other cloud providers/async ways to get PK that we should support - but we aren't rolling out support for everything initially - we should probably have something like that so people aren't just locked into using AWS for this (or having to write their own wallet implementation)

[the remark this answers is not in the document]
jnsdls: I guess how is this different than just proving the AbstractWallet and letting them implement on top of that?

packages/sdk/src/evm/wallets/core/aws.ts
```diff
@@ -0,0 +1,79 @@
...
+ * ```
+ */
+export class AWSSecretsManagerWallet extends AbstractWallet {
+  private secretId: string;
+  private secretKeyName: string;
+  private client: SecretsManagerClient;
+  private cachedSigner: ethers.Signer | undefined;
+
+  constructor(options: AwsSecretsManagerWalletOptions) {
+    super();
```
jnsdls, line 45: does `AbstractWallet` have a constructor? if not this is not needed

jnsdls (commented):
abstract wallet doesn't need to be called with super I believe?

packages/sdk/src/evm/wallets/core/aws.ts
adam-maj: No but it still yells at you to always call super for some reason

jnsdls (approved):
fix build etc then good to go

PR #403, opened 2022-11-22
feat: add solana support to cli

avneesh0612:


packages/cli/src/create/command.ts
```diff
@@ -10,14 +10,27 @@ import chalk from "chalk";
...
+  options: any,
+) {
+  if (pType === "app") {
+    projectType = "app";
+  }
+
+  if (pType === "contract") {
+    projectType = "contract";
+  }
+
```
jnsdls, line 33: these need to also include `options.app` and `options.contract` respectively here

also they need to be `else` statements because otherwise I can do something like

```thirdweb create app --contract``` this should end up with projectType of `app` but since `--contract` gets evaluated afterwards and separately I will end up with projectType of `contract`

Comment (jnsdls) [reply packages/cli/src/create/command.ts:33]: also if and when we're in `app` vs `contract` mode the subsequent valid options change:
if I pass `projectType` of `contract`:
- `language` should now be ignored below
- `framework` should only be allowed to be the contract frameworks
... and the opposite for `app`

File: packages/cli/src/create/command.ts [hunk 2]
```diff
@@ -62,7 +75,28 @@ export async function twCreate(pPath: string = "", options: any) {
...
   }
 
-  if (!projectType && !options.template) {
+  if (options.solana) {
+    chain = "solana";
+  }
+
+  if (options.evm) {
+    chain = "evm";
+  }
```
jnsdls, line 84: these are (currently) only valid for `app` create, right?

packages/cli/src/create/command.ts
```diff
@@ -76,7 +110,7 @@ export async function twCreate(pPath: string = "", options: any) {
     if (typeof res.projectType === "string") {
       projectType = res.projectType.trim();
     }
-  } else if (!projectType && options.template) {
+  } else if (!projectType || options.template || options.solana) {
```
jnsdls, line 113: what about `options.evm`?

I'd prefer us to just infer the projectType above where we check for the flags and then not need to check again here

packages/cli/src/create/command.ts
```diff
@@ -123,16 +157,41 @@ export async function twCreate(pPath: string = "", options: any) {
...
+        choices: [
+          // For EVM, value is nothing because the repo names don't have "evm" in them.
+          { title: "EVM", value: "evm" },
+          { title: "Solana", value: "solana" },
+        ],
+      });
+
+      if (res.chain === "solana") {
+        chain = "solana";
+      }
```
jnsdls, line 174: else chain = 'evm', no?

.changeset/fifty-feet-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"thirdweb": patch
```
jnsdls, line 2:
```suggestion
"thirdweb": minor
```
since we're now changing the params of `thirdweb create` this needs to be a minor version

jnsdls (changes requested):
more things to look at please :)

jnsdls:
please update integration tests for the new structure

avneesh0612:
updated the test @jnsdls

jnsdls (approved):
code-wise now looks good to me

packages/cli/src/create/command.ts
```diff
@@ -123,16 +138,43 @@ export async function twCreate(pPath: string = "", options: any) {
...
-          { title: "Next.js", value: "next" },
-          { title: "Create React App", value: "cra" },
-          { title: "Vite", value: "vite" },
-        ],
+        choices:
+          // Solana doesn't support Vite or CRA just yet:
+          chain === "solana"
+            ? [
+                { title: "Next.js", value: "next" },
+                // { title: "Create React App", value: "cra" },
```
jarrodwatts, line 170: let's uncomment this now

jarrodwatts: just CRA not vite

PR #412, opened 2022-11-27
[SDK] Update signTypedData domain to match contract

easonchai:
This is a patch for the ERC1155 Signature Mint function in the SDK.

The current function uses the domain `TokenERC1155` which should actually match the domain in the smart contract, which is [`SignatureMintERC1155`](https://github.com/thirdweb-dev/contracts/blob/main/contracts/extension/SignatureMintERC1155.sol#L18). Because the domains do not match, the signer recovered after processing the request will always be different. This means all transactions will fail as the signer recovered is not expected.

This change was also communicated and discussed with @adam-maj on Discord after a week of debugging.

Changes were tested according to contributing guidelines!

### DIFF
--- a/.changeset/khaki-panthers-grin.md
+++ b/.changeset/khaki-panthers-grin.md
@@ -2,4 +2,4 @@
 "@thirdweb-dev/sdk": patch
 ---
 
-Updated signTypedData domain to match contract from TokenERC1155 to SignatureMintERC1155
+Correctly handle contracts using the SignatureMintERC1155 ContractKit extension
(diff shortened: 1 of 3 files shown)


joaquim-verges:
@easonchai thank you for this! great catch. 

We still need to handle the case for our own Edition contract which uses `TokenERC1155` as the domain, i've added a case for this. Also added a test for contracts using the `SignatureMintERC1155` ContractKit extension. 

If all tests pass we should be good to merge :)

Thank you again!

joaquim-verges (approved):

PR #420, opened 2022-11-30
init unity-js-bridge package

jnsdls:


adam-maj:
Should we move the `AwsKmsWallet` and `AwsSecretsManagerWallet` out of SDK and into this new `@thirdweb-dev/wallets` package as well?

Issue #411, opened 2022-11-26
[SDK] - Expose convenience function to do `ethers.utils.formatEther` and `ethers.utils.parseEther`


jarrodwatts:
## Background

Working with `BigNumber` variables requires importing `ethers` and using `ethers.utils.formatEther` and `parseEther` functions which is not very discoverable through our SDK unless you are already familiar with the ethers dependency.

It would be great to expose convenience functions that expose these directly from our SDK package rather than ethers.

## Goal

[] Add `toEther` function
  - Users can import toEther from the @thirdweb-dev/sdk package and convert a number from wei to formatted value. Use the ethers.utils.formatEther functionality under the hood.

[] Add `toWei` function
  - Same as above, except use parseEther under the hood.

yehia67:
Hey @jarrodwatts 👋 
I would be happy to help!
But I am not sure where it is better to add them inside `sdk/src/evm/common/utils.ts` to be similar to the `ethers` package. Or should I add them on `sdk/src/evm/common/currency.ts` I see we have similar methods there like `normalizeAmount` and `fetchCurrencyValue`

joaquim-verges:
@yehia67 thanks for helping out! I think it makes sense to put them in `currency.ts` with the other currency helper functions :)

jnsdls:
fixed in #421 by @yehia67 thx!

yehia67:
@jnsdls Much welcome I am looking forward to the new contribution here 🚀

PR #421, opened 2022-11-30
[SDK] add two new utilities toEther and toWei

yehia67:
#411


nachoiacovino:
I think it might be better for it to be `formatUnits` and accept decimals as optional second param defaulting to 18, thoughts @joaquim-verges ? in this form we couldn't use it for USDC for example

yehia67:
@nachoiacovino  In that case we will have to change the name from `toWei` and `toEther` to `formatUnit` and `parseUnit`

joaquim-verges:
@yehia67 nah i actually like `toWei` and `toEth` better. How about adding a third method that would be `toUnits` that takes in the decimals for custom ERC20s?

joaquim-verges:
also make sure to add a changeset 🙏

nachoiacovino:
> @yehia67 nah i actually like `toWei` and `toEth` better. How about adding a third method that would be `toUnits` that takes in the decimals for custom ERC20s?

what about the other way around? from units to wei for custom erc20?

yehia67:
@joaquim-verges @nachoiacovino I can add two utilities `toUnits` and `parseUnits` 
Sure I will add the changeset on the two PRs sorry for that

packages/sdk/src/evm/common/currency.ts
```diff
@@ -192,3 +192,22 @@ export async function normalizeAmount(
   const decimals = await contractWrapper.readContract.decimals();
   return utils.parseUnits(AmountSchema.parse(amount), decimals);
 }
+
+export function toEther(amount: Amount): string {
+  return utils.formatEther(AmountSchema.parse(amount));
+}
+
+export function toWei(amount: BigNumber | string): BigNumber {
```
joaquim-verges, line 200: let's use `BigNumberish` here as the input type

packages/sdk/src/evm/common/currency.ts
```diff
@@ -192,3 +192,22 @@ export async function normalizeAmount(
...
+  return utils.parseEther(AmountSchema.parse(amount));
+}
+
+export function toUnits(amount: Amount, decimals: number): BigNumber {
+  return utils.parseUnits(AmountSchema.parse(amount), decimals);
+}
+
+export function formatUnits(
+  amount: BigNumber | string,
```
joaquim-verges, line 209: same here

packages/sdk/src/evm/common/currency.ts
```diff
@@ -192,3 +192,22 @@ export async function normalizeAmount(
...
+
+export function toWei(amount: BigNumber | string): BigNumber {
+  return utils.parseEther(AmountSchema.parse(amount));
+}
+
+export function toUnits(amount: Amount, decimals: number): BigNumber {
+  return utils.parseUnits(AmountSchema.parse(amount), decimals);
+}
+
+export function formatUnits(
```
joaquim-verges, line 208: how about calling this `toDisplayValue` instead so its clear what the intent is?

packages/sdk/src/evm/common/currency.ts
yehia67: Ah, so sorry, I switched between them.
But I want to add `BigNumber | string)` in case the number format is something like this `1000000000000000000`. The `utils.formatEther` support both big numbers and string, so the user doesn't need to transfer string to a bigNumber. wdyt?

packages/sdk/src/evm/common/currency.ts
yehia67: Yes much better

packages/sdk/src/evm/common/currency.ts
yehia67: @joaquim-verges I read my previous comments in bad English 😅 😅  So I updated it. Please tell me if it is unclear 🙏

joaquim-verges: `BigNumberish` (not `BigNumber`) is a union of `string | number | BigNumber` which is what we use in a lot of places

yehia67: Ah okay got it, I will fix this👍

yehia67:
@joaquim-verges should I remove all `./changest` files and leave only the latest one?

packages/sdk/src/evm/common/currency.ts
```diff
@@ -192,3 +192,22 @@ export async function normalizeAmount(
   const decimals = await contractWrapper.readContract.decimals();
   return utils.parseUnits(AmountSchema.parse(amount), decimals);
 }
+
+export function toEther(amount: Amount): string {
+  return utils.formatEther(AmountSchema.parse(amount));
+}
+
+export function toWei(amount: BigNumberish): BigNumber {
+  return utils.parseEther(AmountSchema.parse(amount));
```
joaquim-verges, line 201: Here you do y have to parse anything, just pass the BigNumberish value in

packages/sdk/src/evm/common/currency.ts
```diff
@@ -192,3 +192,22 @@ export async function normalizeAmount(
...
+
+export function toUnits(amount: Amount, decimals: number): BigNumber {
+  return utils.parseUnits(AmountSchema.parse(amount), decimals);
+}
+
+export function toDisplayValue(
+  amount: BigNumberish,
+  decimals: number,
+): string {
+  return utils.formatUnits(AmountSchema.parse(amount), decimals);
```
joaquim-verges, line 212: Same here

yehia67: Here `decimals` should be `BigNumberish` like the interface `function parseUnits(value: string, unitName?: BigNumberish):` I will fix it

packages/sdk/src/evm/common/currency.ts
yehia67: "pasreEther" params accept string only `parseEther(ether: string)`

joaquim-verges: right so it should be the opposite then:

`toEther(amount: BigNumberish)`
and
`toWei(amount: Amount)`

yehia67:
@joaquim-verges  for `toWei(amount: Amount)` The `parseEther(ether: string):` accept string only, so `Amount` will not pass because it has `string | number`

joaquim-verges:
`Amount.parse(amount)` returns a string so you can use that internally to transform `string | number` to just `string`

yehia67:
@joaquim-verges yes make 100% sense

jnsdls:
thx @yehia67!

PR #422, opened 2022-11-30
[React] add props for button title to change the text

yehia67:


joaquim-verges:
@yehia67 maybe call the prop `buttonTitle` to be explicit ? also don't forget to make a changeset

yehia67:
@joaquim-verges If I made `buttonTitle` mandatory. I will have to remove the default value "btnTitle = "Connect Wallet" and update the component in other places like `Web3Button` to pass the typescript check. Sounds good?

packages/react/src/evm/components/ConnectWallet/index.tsx
```diff
@@ -248,7 +250,7 @@ export const ConnectWallet: React.FC<ConnectWalletProps> = ({
               </span>
             )
           ) : (
-            <span style={{ whiteSpace: "nowrap" }}>Connect Wallet</span>
+            <span style={{ whiteSpace: "nowrap" }}>{btnTitle}</span>
```
jnsdls, line 253: why not just pass `children` at this point?

packages/react/src/evm/components/ConnectWallet/index.tsx
```diff
@@ -63,6 +63,7 @@ interface ConnectWalletProps extends ThemeProviderProps {
     loginOptional?: boolean;
   };
   className?: string;
+  btnTitle?: string;
```
jnsdls, line 66: if not using `children` this should still also allow users to pass jsx in addition to a string

jnsdls (commented):
couple of questions

packages/react/src/evm/components/ConnectWallet/index.tsx
yehia67: It is a dropdown so string was the first got in mind. But I think `children` will be better in case someone want to pass a styles text `<span style={styles}></span>`
I will implement this change, Thank you🙏🏻

yehia67: updated check again, please 🙏

.changeset/sharp-ducks-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/react": major
```
jnsdls, line 2: this is definitely not a major version, `patch` should be totally fine

```suggestion
"@thirdweb-dev/react": patch
```

packages/react/src/evm/components/ConnectWallet/index.tsx
```diff
@@ -63,6 +63,7 @@ interface ConnectWalletProps extends ThemeProviderProps {
     loginOptional?: boolean;
   };
   className?: string;
+  btnTitle?: React.ReactNode;
```
jnsdls, line 66: I believe this should be `JSX.Element`?

.changeset/sharp-ducks-attack.md
yehia67: Yes I am new changeset😅😅

packages/react/src/evm/components/ConnectWallet/index.tsx
jnsdls: or maybe `JSX.Element | string` ?

yehia67: I forgot the difference between`JSX.Element` and `React.ReactNode`, so I checked. 

 `JSX.Element | string` sounds good because it will accept only one element. No arrays will be passed, and it may be just a string. I will update it now

jnsdls:
@yehia67 looking good now - you only need one changeset file, please delete one of them and write the changeset in such a way that consumers of the package know what changed, we'll merge it once that's done and tests pass :)

PR #425, opened 2022-11-30
[Auth] include credentials when requesting user

adam-maj:


joaquim-verges (approved):
not sure why this is needed but 🤷

adam-maj:
> not sure why this is needed but 🤷

Needed if you're running an auth backend on a different URL than your frontend (like express server + frontend). The immediate need is for animoca as GMGN is building out the frontend

PR #428, opened 2022-12-01
[CLI] allow adding contracts to existing projects

adam-maj:


packages/cli/src/cli/index.ts
```diff
@@ -68,6 +68,7 @@ $$$$$$\\   $$$$$$$\\  $$\\  $$$$$$\\   $$$$$$$ |$$\\  $$\\  $$\\  $$$$$$\\  $$$$
       "-t, --template [name]",
       "A template to start your project from. You can use an template repository name from the official thirdweb-example org.",
     )
+    .option("-n, --name [name]", "Name of the new smart contract to create")
```
joaquim-verges, line 71: niiiiice - was thinking we should add a step in the flow to name your contract coz its so annoying to have it named `Contract.sol` everytime

packages/cli/src/create/command.ts
```diff
@@ -97,185 +107,178 @@ export async function twCreate(
...
-        type: "select",
-        name: "chain",
-        message: CREATE_MESSAGES.chain,
-        choices: [
-          { title: "EVM", value: "evm" },
-          { title: "Solana", value: "solana" },
-        ],
+        type: "text",
+        name: "path",
+        message: CREATE_MESSAGES.contractName,
```
joaquim-verges, line 133: yessss

packages/cli/src/create/command.ts
```diff
@@ -97,185 +107,178 @@ export async function twCreate(
...
-        }
-        return "Invalid project name: " + validation.problems?.[0];
-      },
-    });
-
-    if (typeof res.path === "string") {
-      projectPath = res.path.trim();
+  let onlyContract = false;
+  // Whether to only create a new contract without the project
+  const contractProjectType = await detect(projectPath, {});
```
joaquim-verges, line 112: wouldn't it be safer to rely on this being one of the contract project types we support instead of checking if there's a `contracts` folder in the current dir?

jnsdls: yeah can prob use detector for this

adam-maj: The detector is being used above - but probably to get the correct folder, instead of expecting src/contracts, we should just iterate through folders and check for one with a `.sol` file - and then afterwards default to src or contracts

PR #430, opened 2022-12-01
[CLI] Fix metadata upload with special characters

joaquim-verges:


packages/cli/src/common/processor.ts
```diff
@@ -164,7 +164,7 @@ export async function processProject(
     const metadataURIs = await Promise.all(
       selectedContracts.map(async (c) => {
         logger.debug(`Uploading ${c.name}...`);
-        return await storage.upload(JSON.parse(c.metadata), {
+        return await storage.upload(JSON.parse(JSON.stringify(c.metadata)), {
```
adam-maj, line 167: We've had logic similar to this before somewhere (don't remember where) and if we ever pass in something that's _already_ stringified into it, the double stringify adds even more escape characters to the string (if there are `"` for example) - not sure if that applies here.

joaquim-verges: no it doesnt because the end output that i paass to upload is still an object. This just cleans up the original string.

PR #441, opened 2022-12-02
CLI auto-updates on every run

adam-maj:


packages/cli/src/cli/index.ts
```diff
@@ -198,6 +197,108 @@ $$$$$$\\   $$$$$$$\\  $$\\  $$$$$$\\   $$$$$$$ |$$\\  $$\\  $$\\  $$$$$$\\  $$$$
...
+            inactive: "no",
+          });
+
+          if (res.upgrade) {
+            const updateSpinner = spinner(
+              `Upgrading CLI to version ${versionInfo.latest}...`,
+            );
+
+            // Check if thirdweb package is already installed in local package.json
+            const localPackages: string[] = await new Promise((resolve) => {
```
joaquim-verges, line 231: could put this in a util function to unbloat this file a bit

packages/cli/src/cli/index.ts
```diff
@@ -198,6 +197,108 @@ $$$$$$\\   $$$$$$$\\  $$\\  $$$$$$\\   $$$$$$$ |$$\\  $$\\  $$\\  $$$$$$\\  $$$$
...
+                  })
+                  .filter((item) => !!item) as string[];
+                resolve(packages);
+              });
+            });
+
+            const isLocal = localPackages.find((packageName) =>
+              packageName.includes("thirdweb@"),
+            );
+            const command = isLocal ? "npm i thirdweb" : "npm i -g thirdweb";
```
joaquim-verges, line 261: did you try it on replit?

jnsdls: what if I use `yarn` in my project, or `pnpm` ?

packages/cli/src/cli/index.ts
```diff
@@ -2,37 +2,36 @@
...
 import { processProject } from "../common/processor";
 import { cliVersion, pkg } from "../constants/urls";
-import { info, logger } from "../core/helpers/logger";
+import { info, logger, spinner } from "../core/helpers/logger";
 import { twCreate } from "../create/command";
 import { deploy } from "../deploy";
 import { upload } from "../storage/command";
 import { ThirdwebStorage } from "@thirdweb-dev/storage";
 import chalk from "chalk";
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
jnsdls, line 11: y tho?

packages/cli/src/cli/index.ts
```diff
@@ -198,6 +197,108 @@ $$$$$$\\   $$$$$$$\\  $$\\  $$$$$$\\   $$$$$$$ |$$\\  $$\\  $$\\  $$$$$$\\  $$$$
       await detectExtensions(options);
     });
 
+  // eslint-disable-next-line turbo/no-undeclared-env-vars
+  if (process.env.THIRDWEB_CLI_SKIP_INTRO !== "true") {
```
jnsdls, line 201: since ur looking this up multiple times might make sense to extract to const up top

jnsdls (commented):
all of this is scary

packages/cli/src/cli/index.ts
adam-maj: leftover from testing (still wip rn)

packages/cli/src/cli/index.ts
adam-maj: good idea

packages/cli/src/cli/index.ts
adam-maj: Just tried it out on repl.it, looks like everything works (global install working)

adam-maj:
Tested with both `npm` and `yarn` (should be same setup for `pnpm` as well), and on replit - all work as expected.

adam-maj:
And also when there is a local installation of the `thirdweb` package, vs global

adam-maj:
Would be good to get one other confirmation on this given the nature of the changes

adam-maj:
Everything should work on this version [`0.9.3-dev-d5b649c`](https://www.npmjs.com/package/thirdweb/v/0.9.3-dev-d5b649c) (if you install it locally or globally, and then run `npx thirdweb deploy` or anything else, it should auto-upgrade properly and run the command successfully all at once, regardless of package manager.

adam-maj:
I've accounted for + successfully tested all the edge cases:
1. Local installation with `npm`
2. Local installation with `yarn`
3. Local installation with `pnpm`
4. Global installation with `npm`
5. Global installation with `yarn` (doesn't get picked up by `npx`, so only gets hit if you run `thirdweb` directly)
6. Global installation with `pnpm` (doesn't get picked up by `npx`, so only gets hit if you run `thirdweb` directly)
7. No installation, defaults to `thirdweb@latest` - but you already have that version cached
8. No installation, defaults to `thirdweb@latest`, you don't have it cached so it needs to install (but we're passing `--yes` flag so you don't see the prompt)

I think we should be good to go

adam-maj:
The code path is:
1. Check for local installation, if it's there, get the package manager
2. Otherwise check for global installation on all 3 package managers
3. Otherwise , just default to `npx thirdweb@latest` (and auto-accept install)

packages/cli/src/cli/index.ts
```diff
@@ -198,6 +198,129 @@ $$$$$$\\   $$$$$$$\\  $$\\  $$$$$$\\   $$$$$$$ |$$\\  $$\\  $$\\  $$$$$$\\  $$$$
...
+                command = installation.isGlobal
+                  ? `yarn global add thirdweb`
+                  : `yarn add thirdweb`;
+                break;
+              case "pnpm":
+                command = installation.isGlobal
+                  ? `pnpm add -g thirdweb@latest`
+                  : `pnpm add thirdweb@latest`;
+                break;
+            }
```
jnsdls, line 279: we need to handle the case here of not being able to detect any supported package manager:
a) technically others exist
b) someone might invent a new one tomorrow

- we should fail gracefully in those cases not just execute an empty string

jnsdls (changes requested):
1 last thing that stands out to me, then I think we're good?

Issue #82, opened 2022-09-08
[React] - Support 3D File Formats in NFT Media Renderer UI Component


jarrodwatts:
## Feature Request

We have the [NFT Media Renderer](https://portal.thirdweb.com/sdk/ui-components/nft-renderer) component which renders the relevant HTML tag after determining the file type provided by the metadata prop.

This currently only supports image and video files. We would like to add support for common 3D formats such as `.glb`.

## Goal

Be able to provide an NFT that has a `.glb` or other 3D file format file as it's `image` property and have it rendered on the UI.

JustinTime42:
Hi @jarrodwatts,
If this is still a desired feature, I wouldn't mind having a go at it.

jarrodwatts:
That would be awesome @JustinTime42 ! Definitely still something we want to support

PR #470, opened 2022-12-22
Added support for rendering 3d NFTs

JustinTime42:
This adds support for .glb files to be displayed by NFTMediaRenderer per issue https://github.com/thirdweb-dev/js/issues/82.


JustinTime42:
This is a fairly basic, initial implementation using Three.js. It allows the user to pass in some simple options like camera position, ambient light, and Frustrum. I'm open to feedback on other options we might want to support. I'm also looking into file formats other than .glb as well.

jnsdls:
@JustinTime42 thanks for this PR, at first glance this looks good though I had hoped we might potentially be able to rely on [model viewer](https://web.dev/model-viewer) for the 3d rendering to avoid unnecessarily bundling three js etc (also supports more file types I believe)

Let me know if you want to take a stab at implementing this with model viewer instead, otherwise I can review it as is in the coming days and we can get this in as an immediate improvement. :)

JustinTime42:
@jnsdls Totally makes sense - thanks for the feedback! 
The latest commit changes to using model-viewer for 3D models. For the moment, I've had to strip it down a bit and add several 'any' types to make it work. I'm running into some issues with typescript complaining about the model-viewer custom web element. I'm not very familiar with making custom web elements play nice with React and Typescript, but I can study up on that and get it figured out. 

But I figured I'd commit what I have working so I can get feedback. If this looks like I'm generally on the right track I'll get to work fixing the types and getting everything working properly and safely.

JustinTime42:
@jnsdls 
I've got the 3D rendering working pretty well now using model-viewer. It downloads the model-viewer script when the ModelViewer component loads so we avoid bundling it with our package or downloading it unless it's needed.

I have not been able to find an effective way of properly typing the <model-viewer> element without installing model-viewer as a dependency, so I've left that as an 'any' type where I added model-viewer as a JSX intrinsic element. I'm definitely open to suggestions on this - I'm not a typescript expert yet so I might be missing something here. 

Meanwhile this latest commit seems to work pretty well and I think it'll make a decent first version of supporting 3D models. Could you take a look and let me know what you think?

joaquim-verges:
@JustinTime42 looks like you ran npm install instead of yarn install btw - mind deleting the package-lock.json?

JustinTime42:
@joaquim-verges Oops - package-lock.json deleted.

packages/react/src/evm/components/MediaRenderer.tsx
```diff
@@ -395,6 +405,44 @@ const LinkPlayer = React.forwardRef<HTMLAnchorElement, MediaRendererProps>(
...
+  
+    useEffect(() => {
+      loadModelViewer();
+    },[])
+  
+    const loadModelViewer = () => {
+      const existingScript = document.getElementById('modelViewer');
+      if (!existingScript) {
+        const script = document.createElement('script');
+        script.src = 'https://unpkg.com/@google/model-viewer/dist/model-viewer.min.js';
```
jnsdls, line 421: would prefer to dynamically load this from `node_modules` (bundle it in but as a code-split file using dynamic `import()`)

packages/react/src/evm/components/MediaRenderer.tsx
```diff
@@ -25,6 +25,8 @@ export interface SharedMediaProps {
    * Show the media controls (where applicable) (default false)
    */
   controls?: HTMLVideoElement["controls"];
+
+  children?: any;
```
jnsdls, line 29: why does this need to be any?

packages/react/src/evm/components/MediaRenderer.tsx
```diff
@@ -395,6 +405,44 @@ const LinkPlayer = React.forwardRef<HTMLAnchorElement, MediaRendererProps>(
 
 LinkPlayer.displayName = "LinkPlayer";
 
+const ModelViewer = React.forwardRef<HTMLCanvasElement, MediaRendererProps>(
```
jnsdls, line 408: I wonder if there is a pre-made react-wrapper around the `model-viewer` library that we can just pull in instead?

jnsdls (commented):
thanks this is the right direction for sure!

packages/react/src/evm/components/MediaRenderer.tsx
JustinTime42: That was my initial thought as well, so I did a little research on this and was surprised to find no well-used and well-maintained projects in this category. Maybe I'm missing something, but here's what I had found:

This one only supports STL format:
https://www.npmjs.com/package/react-3d-model-component

These two are not currently maintained, and a brief attempt at implementing them didn't seem to work. I'm not sure if they are broken or if I'm missing something.
https://github.com/devhims/model-viewer-react
https://www.npmjs.com/package/react-model-viewer

This one I got to work fine with vanilla Javascript, but its React component doesn't seem to be implemented properly.
https://naver.github.io/egjs-view3d

This was a fairly brief search and brief attempts at using what already existed, but not finding any real clear winning solution is what prompted me to implement it the way I did. 

I'll dig a little deeper though, as I may very well have missed something. It's possible one of these can be made to work, or maybe there's something else that I've not found.

packages/react/src/evm/components/MediaRenderer.tsx
JustinTime42: It doesn't :) 
I replaced it with ReactNode

packages/react/src/evm/components/MediaRenderer.tsx
jnsdls: sounds good this makes sense

JustinTime42:
@jnsdls I've looked a little more into this and I really haven't found a better pre-made react wrapper to use, but I've pulled mine into a separate component for lazy loading, and it seems to work well. I deployed a smart contract with some 3d NFTs for testing (models shamelessly ripped off the internet) and pushed my changes to my local dashboard code for testing and the screenshots below show what the NFT drawer of the Thirdweb Dashboard looks like using my changes to the js monorepo. The models load in fully interactable using default model-viewer settings. 

Screenshot 2023-01-18 at 12-57-57 Non Fungible Spacecraft Goerli
Screenshot 2023-01-18 at 12-58-32 Non Fungible Spacecraft Goerli

nachoiacovino:
Hello @JustinTime42 thanks so much, do you think you would be able to share the contract you used to test? I was testing on this contract and it wasn't rendering for me: https://thirdweb.com/polygon/0xBbcCd7a59fD4A7DC8DB710B6D6c338dF4174B778/nfts (glb file)

Seems like it's also not limiting the width/height so it's much wider than the other file types. Should have the same behaviour.

What files types did you tested with? Any you know for sure works and any that don't with this solution?

JustinTime42:
@nachoiacovino Well I seem to be having one of those annoying 'it works on my machine' moments.  Here's what it looks like on my local dashboard installation. I'll chase this down and see what I've changed here.

Untitled

nachoiacovino:
thanks @JustinTime42, do you think you can share a couple of contracts that works for you? will also try some stuff out and see if I can figure it out.

JustinTime42:
@nachoiacovino It seems to have to do with the way yarn.lock is being generated and the dependence on `@google/model-viewer` 

I was playing around with the dashboard code on a branch I have here: https://github.com/JustinTime42/thirdweb-dashboard/tree/gltf-support
On that branch, yarn.lock has a reference to `@google/model-viewer`, and it is that branch that is rendering the 3d models properly. If I checkout the main branch, and hook up my 3d renderer branch of the js monorepo via yalc, `@google/model-viewer` is absent from the yarn.lock file and the 3d models do not render properly. 

I not very familiar with yalc and turborepo. Am I missing a step in adding `@google/model-viewer` as a dependency so that it will be referenced as a sub-dependency properly?

jnsdls:
@JustinTime42 if that's the issue we might just need to set `connect-src * data: blob:` in the dashboard CSP

JustinTime42:
@jnsdls  I just put in this PR:https://github.com/thirdweb-dev/dashboard/pull/823. It makes that change to the CSP and also adds `.gltf` to the file extensions that the mint-form accepts for minting 3d models as NFTs.

jnsdls:
> @jnsdls I just put in this PR:[thirdweb-dev/dashboard#823](https://github.com/thirdweb-dev/dashboard/pull/823). It makes that change to the CSP and also adds `.gltf` to the file extensions that the mint-form accepts for minting 3d models as NFTs.

kicking off a deploy right now, did you validate it already running locally?

JustinTime42:
@jnsdls Yes I did
With this PR plus the one over on the dashboard side it is now rendering the 3d models. I tested it on the contract @nachoiacovino was using earlier as well as my own test contract here: [/goerli/0x27B7B636e3EF78631a3808cD97bDEE2Cf2144804/nfts](http://localhost:3000/goerli/0x27B7B636e3EF78631a3808cD97bDEE2Cf2144804/nfts)  

I also tested that dashboard will now accept .gltf files for minting.

jnsdls:
made a couple of changes, @JustinTime42. Let me know if this all still works in the same ways that you've been testing it!

JustinTime42:
@jnsdls The only adjustment needed is that `height:100%` ends up being zero height and looks like so: 
height100

I changed it to `style={{ objectFit: 'contain', width: "100%" }` which constrains the width to keep the Media column consistent and allows the height to maintain aspect ratio. It looks like this: 
object-contain-width100

packages/react/src/evm/components/MediaRenderer.tsx
```diff
@@ -445,6 +457,30 @@ export const MediaRenderer = React.forwardRef<
           {...restProps}
         />
       );
+    } else if (videoOrImageSrc.mimeType.startsWith("model")) {
+      return (
+        <Suspense
+          fallback={
+            poster ? (
```
nachoiacovino, line 464: I don't think this is necessary, we can pass the poster to the ModelViewer, <model-viewer> accepts a prop poster that does exactly what we want

nachoiacovino:
Hey @JustinTime42 awesome job!

Now I can finally see the NFTs on my dashboard, one thing that happens is that I see them on the NFT List, but when I click on the the NFTs, I don't see them rendered on the NFT Drawer.

image

Basically it seems like it doesn't have a width, because adding either a width or min-width fixes it: 
image

If we could figure out this last thing out we can ship this!

jnsdls:
e2e test on CRA is failing

nachoiacovino:
shipping it @JustinTime42 🚀

PR #471, opened 2022-12-24
[SDK] expose getSnapshotEntryForAddress

adam-maj:


jnsdls:
@adam-maj why is this not good to merge? seems harmless?

PR #492, opened 2023-01-12
[SDK] Fixed issue 487, return tx hash when deploy transaction is sent

cmayorga:
This patch will propagate deploy transaction event to the parent EventFactory and allow to get transaction hash once sent while wait for deploy to finish and return proxy address.

### DIFF
--- a/.changeset/tender-moons-allow.md
+++ b/.changeset/tender-moons-allow.md
@@ -1,0 +1,7 @@
+---
+"@thirdweb-dev/sdk": patch
+---
+
+Added a contractWrapper accessor method to FactoryEvents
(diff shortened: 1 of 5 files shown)


joaquim-verges (changes requested):
thank you so much for your contribution!

I think there's a different way to approach this, that could cover all contract deploys and not just the factory ones. Your PR helped me come to the conclusion that our initial approach was wrong :p 

I put up a PR here: https://github.com/thirdweb-dev/js/pull/494/files

joaquim-verges:
closing now that https://github.com/thirdweb-dev/js/pull/494 is merged

PR #493, opened 2023-01-13
[React] Move Solana to peer dependencies

joaquim-verges:


.changeset/lovely-hotels-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/react": patch
```
jnsdls, line 2: ur also changing things in sdk so missing a changeset for that

jnsdls (changes requested):

.changeset/lovely-hotels-bake.md
joaquim-verges: done

PR #494, opened 2023-01-13
[SDK] Fix deploy transaction listener for all deploy types `sdk.deployer.addDeployListener()`

joaquim-verges:
This now works with all deploy types - factory/proxy/direct

API is simpler too:

`sdk.deployer.addDeployListener((event) => console.log(event)}`


cmayorga:
Thanks @joaquim-verges . Still i need to go deeper in understanding all the huge code there :) So i tried to make a not invasive fix in order to break nothing....
I forked and checked new version. The problem is the tx hash is only coming at the end of deployment. 

But still need a solution to provide "submitted" event before wait for confirmations.

The whole reason for this is to get the transaction hash early on, so we can react... in edge cases like user leaving the dapp, or his connections breaks etc... this way ... we know what transaction to follow when we get back....

so the idea is to:

- send transaction (deploy selected module)
- get the transaction hash back immediately 
- handle the rest as before, like get the deployed feedback or anything esle...
hope it makes sense

I created a new PR with the fix for this, please check and let me to know

https://github.com/thirdweb-dev/js/pull/499

Thanks.

PR #496, opened 2023-01-13
[React] Render preview image for 3d files

nachoiacovino:
Temporary workaround


nachoiacovino:
We'll do it correctly, probably following #470

PR #499, opened 2023-01-16
[SDK] Adding and event to return transaction hash immediately after calling deploy

cmayorga:
Added implementation to emit a submitted event when deploy methods are called, in order to receive transaction hash as soon as transaction is send to network, before to wait to deploy to be finished and confirmed. 

The whole reason for this is to get the transaction hash early on, so we can react... in edge cases like user leaving the dapp, or his connections breaks etc... this way ... we know what transaction to follow when we get back....

so the idea is to:

- send transaction (deploy selected module)
- get the transaction hash back immediately
- handle the rest as before, like get the deployed feedback or anything else...

Hope it makes sense

(Metamask is not shown in the video)

https://user-images.githubusercontent.com/5519725/212624621-1405d672-224b-4d84-8303-3d7b0fd65bc3.mp4

### DIFF
--- a/.changeset/weak-schools-march.md
+++ b/.changeset/weak-schools-march.md
@@ -1,0 +1,5 @@
+---
+"@thirdweb-dev/sdk": patch
+---
+
+[SDK] Added event for submitted deploy transaction
--- a/packages/sdk/src/evm/common/currency.ts
+++ b/packages/sdk/src/evm/common/currency.ts
@@ -205,6 +205,9 @@ export function toUnits(amount: Amount, decimals: BigNumberish): BigNumber {
   return utils.parseUnits(AmountSchema.parse(amount), decimals);
 }
 
-export function toDisplayValue(amount: BigNumberish, decimals: BigNumberish): string {
+export function toDisplayValue(
+  amount: BigNumberish,
+  decimals: BigNumberish,
+): string {
   return utils.formatUnits(amount, decimals);
 }
--- a/packages/sdk/src/evm/core/classes/contract-deployer.ts
+++ b/packages/sdk/src/evm/core/classes/contract-deployer.ts
@@ -805,9 +805,9 @@ export class ContractDeployer extends RPCConnectionHandler {
     const deployer = await new ethers.ContractFactory(abi, bytecode)
       .connect(signer)
       .deploy(...constructorParams);
-	this.events.emit("contractDeploySubmitted", {		
-		transactionHash: deployer.deployTransaction.hash,
-	});	  
+    this.events.emit("contractDeploySubmitted", {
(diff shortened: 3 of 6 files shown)


joaquim-verges:
sorry for the delay here @cmayorga - busy week! I'll be looking into this in the next few days.

the use case makes sense and is def something we should support, just want to dig into it a bit more to make sure this is the right abstraction.

Would it be acceptable to have a separate `sdk.deployer.addTransactionListener()` instead of bundling it all into the same event?

cmayorga:
> sorry for the delay here @cmayorga - busy week! I'll be looking into this in the next few days.
> 
> the use case makes sense and is def something we should support, just want to dig into it a bit more to make sure this is the right abstraction.
> 
> Would it be acceptable to have a separate `sdk.deployer.addTransactionListener()` instead of bundling it all into the same event?

Hi @joaquim-verges, thank you for your comment. Not sure if needed to separate at all (i guess you mean to separate listeners, not just events). I think, at the end, we would want the same functionality as per other functions like, in example, "mint", where the user receives "submitted" as soon as transaction is created and "completed" once is settled. 
For deployListener i added a new deploy event called "contractDeploySubmitted" that is different to the "contractDeployed" event, so there are separate events at all sent to the same DeployListener.

Let me to know if this make sense.

joaquim-verges:
did a bit of work on this, taking inspiration from your PR, you'll now get 2 events for each deploy, one at submission one at deploy, like this
```
{
  status: 'submitted',
  transactionHash: '0x12e86885ff1ff0c94cfb7e238e4b0b8546f773a30def174acfb8ac85c662f5c9'
}
{
  status: 'completed',
  contractAddress: '0xFE153A4bC3927f541284E912238D9f822412B85d',
  transactionHash: '0x12e86885ff1ff0c94cfb7e238e4b0b8546f773a30def174acfb8ac85c662f5c9'
}
{
  status: 'submitted',
  transactionHash: '0xe8406df9a6719a77b79dd449d2ffee52302d85eff17df9a6f2866f5de9fcd35e'
}
{
  status: 'completed',
  contractAddress: '0x809d550fca64d94Bd9F66E60752A544199cfAC3D',
  transactionHash: '0xe8406df9a6719a77b79dd449d2ffee52302d85eff17df9a6f2866f5de9fcd35e'
}
```

implemented it in https://github.com/thirdweb-dev/js/pull/502 - will do a release with it shortly.

Sorry for the back and forth and thank you for pushing on this!

cmayorga:
@joaquim-verges Thank you very much for the support. This change will add a lot of value to our projects

joaquim-verges:
@cmayorga just released v3.6.10 with the change!

PR #500, opened 2023-01-18
[SDK] Fix getting domain separator error

retocrooman:
When executing a permit with a token called JPYC, which is a Japanese yen stablecoin, it was interrupted by a runtime error. The cause was to get the domain separator, and JPYC did not have that function. However, getting the domain separator should not be a required feature of the permit. So shouldn't we properly handle errors so that runtime errors do not occur?


joaquim-verges (approved):
Thank you for your contribution! i think it makes sense to make this optional since it's mainly used for polygon

PR #512, opened 2023-01-25
[ReactSDK] Changes useClaimedNFTs to only fetch claimed NFTs

ikethirdweb:
`useClaimedNFTs` was fetching all nfts from the contract by calling `useNFTs`. This makes it so we call getAllClaimed from the contract

Tested:
- Locally imported the package and made sure only claimed NFTs were returned

### DIFF
--- a/.changeset/rich-pigs-cry.md
+++ b/.changeset/rich-pigs-cry.md
@@ -1,5 +1,5 @@
 ---
-"@thirdweb-dev/react": minor
+"@thirdweb-dev/react": patch
 ---
 
-[ReactSDK] update useClaimedNFTs to only fetch claimed NFTs
+Updates useClaimedNFTs to only fetch claimed NFTs
--- a/.changeset/rich-pigs-cry.md
+++ b/.changeset/rich-pigs-cry.md
@@ -1,0 +1,5 @@
+---
+"@thirdweb-dev/react": minor
+---
+
+[ReactSDK] update useClaimedNFTs to only fetch claimed NFTs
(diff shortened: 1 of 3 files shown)


.changeset/rich-pigs-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/react": minor
```
jnsdls, line 2: This is a patch change, you can simply change the string in this file and commit it and it'll get picked up

.changeset/rich-pigs-cry.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+[ReactSDK] update useClaimedNFTs to only fetch claimed NFTs
```
jnsdls, line 5: Don't need the "[ReactSDK]" here, the changelog already applies to this package so that's automatic, we just do that on the PR titles themselves :)

jnsdls (changes requested):
Small changes to apply, then also need to fix the linting fail after that good to go

.changeset/rich-pigs-cry.md
ikethirdweb: I was actually thinking I could make this a major change since it's not backward compatible and any users using the `useClaimedNFTs` hook will get a different behavior now

ikethirdweb: Synced offline. Since there's no behavior change we'll leave it at patch

jnsdls (approved):
LGTM

PR #517, opened 2023-01-27
feat: add express, node.js, and react native templates

avneesh0612:


packages/cli/src/cli/index.ts
```diff
@@ -53,6 +53,9 @@ const main = async () => {
     .option("--cra", "Initialize as a Create React App project.")
     .option("--next", "Initialize as a Next.js project.")
     .option("--vite", "Initialize as a Vite project.")
+    .option("--reactNative", "Initialize as a React Native project.")
+    .option("--express", "Initialize as a Express project.")
+    .option("--node", "Initialize as a Node project.")
```
jnsdls, line 58: this is getting too busy, fine for now but we'll figure out a more scaleable solution for templates soon

jnsdls (approved):
LGTM thx

packages/cli/src/cli/index.ts
avneesh0612: okay sounds good

PR #522, opened 2023-01-30
[React Native SDK] init package

iketw:
Adds a ThirdwebProvider implementing he latest version of wagmi.

Tested by importing it into a react native app and using the provider to connect to a wallet.


packages/react-native/src/evm/constants/walletConnect.ts
```diff
@@ -0,0 +1,2 @@
+export const TW_WC_PROJECT_ID = '145769e410f16970a79ff77b2d89a1e0'
```
joaquim-verges, line 1: oooh so we can let ppl use this one by default? is there any limitations?

packages/react-native/src/evm/providers/full.tsx
```diff
@@ -0,0 +1,356 @@
...
+/**
+ * @internal
+ */
+export type WalletConnector = WalletConnectConnectorType;
+
+/**
+ * @internal
+ */
+export type ChainRpc<TSupportedChain extends SupportedChain> = Record<
+  TSupportedChain extends Chain ? TSupportedChain['id'] : TSupportedChain,
```
joaquim-verges, line 49: i think this changed slightly with the anyEVM branch - we should probably consolidate since that's coming out this week. cc @jnsdls

packages/react-native/src/evm/providers/full.tsx
```diff
@@ -0,0 +1,356 @@
...
+ * ```
+ *
+ * @public
+ *
+ */
+export const ThirdwebProvider = <
+  TSupportedChain extends SupportedChain = SupportedChain
+>({
+  sdkOptions,
+  chainRpc = DEFAULT_RPC_URLS,
```
joaquim-verges, line 175: yeah the options changed slightly (this one got deprecated)

packages/react-native/src/evm/constants/walletConnect.ts
ikethirdweb: I'm pretty sure they will throttle by projectID in the future

packages/react-native/src/evm/providers/full.tsx
ikethirdweb: Ok, I'll wait to merge

PR #524, opened 2023-01-31
fix: chain id of polygon zkEVM

avneesh0612:


jarrodwatts:
Seems like chainlist is wrong on this one, from the official polygon zk evm docs:

image

packages/chains/chains/1422.json
```diff
@@ -11,8 +11,8 @@
   },
   "infoURL": "https://polygon.technology/solutions/polygon-zkevm/",
   "shortName": "zkevmtest",
-  "chainId": 1402,
-  "networkId": 1402,
+  "chainId": 1422,
+  "networkId": 1422,
```
joaquim-verges, line 15: This get automatically pulled from the ethereum-list/chains repo, so it'll get overwritten. Best would be to put up a PR on their repo of it's wrong. Otherwise overwrite it in the script

avneesh0612: gotcha

avneesh0612: https://github.com/DefiLlama/chainlist/ is it this repo?

joaquim-verges: https://GitHub.com/ethereum-list/chains

avneesh0612: ah okay

joaquim-verges: Easy enough to override in our script that downloads all the chains tho, we already have an override for localhost in there

avneesh0612: hmm so should we just override instead?

PR #528, opened 2023-01-31
[CLI] Modify auto-update process with flag and caching

adam-maj:
4 changes in this PR:
- Added `--skip-update-check` flag that will always skip checking for updates
- Added disk caching to save the timestamp of the last update check. Update checks won't happen again if there has already been another check in the past 24h.
- Update check now happens on a `preAction` hook on the `program` rather than blocking the core flow of the main function. This means that it only runs on actual command calls and won't run on simpler runs like `thirdweb --version`
- Fixed broken link to `https://thirdweb.com/contracts` -> `https://thirdweb.com/explore` on deploy flow when there are no contracts in your repository.


jnsdls:
pls no more flags

joaquim-verges:
oooo the caching will be the game changer here, i'll test this in a bit

jnsdls (approved):
LGTM, couldn't break it

PR #547, opened 2023-02-05
[SDK] Add proper error catching for when user has no funds

adam-maj:
When the user has no funds, we want to throw a more informative error that lets users know this. However, we also need to preserve the UX of a transaction popping for users with wallets like MetaMask where they will see on UI that they have insufficient funds for the transaction, so we can't throw before `sendTransaction`.

The solution is to let `estimateGas` fail, but then if the error is due to insufficient funds, `callStatic` will succeed and we can let it pass through to `sendTransaction`. Then, when we catch `sendTransaction`, we can check if user balance is insufficient and throw an informative error.

**Update:** Additionally, if we manually set `gasLimit`, `sendTransaction` will always go through. However, `sendTransaction` can succeed, and then we can error on `tx.wait()` - in order to account for this, we also need to catch `tx.wait()` which will return a generic `transaction failed` error message, and make an `eth_call` to extract the proper error.

### DIFF
--- a/packages/sdk/src/evm/core/classes/contract-wrapper.ts
+++ b/packages/sdk/src/evm/core/classes/contract-wrapper.ts
@@ -415,6 +415,7 @@ export class ContractWrapper<
             ...(callOverrides.value ? [{ value: callOverrides.value }] : []),
           );
         } catch (err: any) {
+          console.log("Throwing hre instead!");
           throw await this.formatError(err, fn, args, callOverrides);
         }
       }
@@ -424,6 +425,7 @@ export class ContractWrapper<
     try {
       return await func(...args, callOverrides);
     } catch (err) {
+      console.log("Throwing hre instead!");
       const from = await (callOverrides.from || this.getSignerAddress());
       const value = await (callOverrides.value ? callOverrides.value : 0);
       const balance = await this.getProvider().getBalance(from);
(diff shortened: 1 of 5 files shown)


packages/sdk/src/evm/core/classes/contract-wrapper.ts
```diff
@@ -415,21 +415,32 @@ export class ContractWrapper<
             ...(callOverrides.value ? [{ value: callOverrides.value }] : []),
           );
         } catch (err: any) {
+          console.log("Throwing hre instead!");
```
joaquim-verges, line 418: cleanup pls

packages/sdk/src/evm/core/classes/contract-wrapper.ts
```diff
@@ -415,21 +415,32 @@ export class ContractWrapper<
...
-          "Failed to estimate gas for transaction. You may have insufficient funds in your account to execute this transaction.",
-        );
       }
     }
 
     // Now there should be no gas estimate errors
     try {
       return await func(...args, callOverrides);
     } catch (err) {
+      console.log("Throwing hre instead!");
```
joaquim-verges, line 428: here too

packages/sdk/test/evm/errors.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    );
+
+    try {
+      await sdk.deployer.deployNFTCollection({
+        name: "Should Fail Collection",
+        primary_sale_recipient: "0x0000000000000000000000000000000000000000",
+      });
+      expect.fail();
+    } catch (err) {
+      expect(err.message).to.contain(
```
joaquim-verges, line 18: we have a nice utility for these btw - `expectError`

joaquim-verges (approved):

PR #550, opened 2023-02-06
[SDK] Add solidity stack traces to error messages

adam-maj:
Because we have self-verified contracts, when we have a revert error, we should be able to find where the error is coming from in the smart contract code (solidity stack trace). This PR is an attempt at an initial (simple) form of this flow.

When we have a revert message, we fetch the contract source, and search through the files for the revert message. If we successfully find the error message in the contract source, then we display the lines surrounding the revert in the source file, along with the file name.

**Example:**
image

This is a simplified approach, **as there are still a number of unhandled cases:**
- If the revert message isn't constant, but is constructed with variables in between, (ie. `revert("User adress" + address + "does not have permission Minter")`, it will not get caught since we are checking for the revert string directly.
- If the revert message appears multiple times in the different source files, we'll just pickup the first one (which could be wrong)


packages/sdk/src/evm/core/classes/contract-wrapper.ts
```diff
@@ -510,6 +513,23 @@ export class ContractWrapper<
     // Parse the revert reason from the error
     const reason = parseRevertReason(error);
 
+    // Get contract sources for stack trace
+    let sources: ContractSource[] | undefined = undefined;
+    const storage = new ThirdwebStorage(); // Take thirdweb storage in constructor at some point?
+    try {
+      const metadata = await fetchContractMetadataFromAddress(
```
joaquim-verges, line 520: Since we're fetching this everytime, how about putting the name of the contract in the error as well. That should be almost always available

packages/sdk/src/evm/core/classes/contract-wrapper.ts
```diff
@@ -510,6 +513,23 @@ export class ContractWrapper<
     // Parse the revert reason from the error
     const reason = parseRevertReason(error);
 
+    // Get contract sources for stack trace
+    let sources: ContractSource[] | undefined = undefined;
+    const storage = new ThirdwebStorage(); // Take thirdweb storage in constructor at some point?
```
joaquim-verges, line 518: Do we not have one available in this class? Could at least reuse the instance instead of re-creating one everytime

joaquim-verges (approved):

PR #567, opened 2023-02-10
refactor: Release to Publish

shift4id:


jnsdls:
@shift4id can you add 2 changesets (via `yarn changeset`)
- both can be `patch`
- CLI changeset should explain that `release` is now `publish` and `release` is deprecated
- sdk changeset should just mention that `release` is now called `publish` (since it doesn't seem there's any functional changes)

packages/cli/README.md
```diff
@@ -41,7 +41,7 @@ Deploying contracts this way gives you access to:
 ## Releasing your contract
 
 ```shell
-npx thirdweb@latest release
+npx thirdweb@latest publish
```
jnsdls, line 44: latest is no longer required
```suggestion
npx thirdweb publish
```

packages/cli/README.md
```diff
@@ -94,7 +94,7 @@ Clone the repo and run this command after installing the CLI tool:
 
 ```bash
 $ cd examples/hardhat
-$ npx thirdweb@latest release
+$ npx thirdweb@latest publish
```
jnsdls, line 97:
```suggestion
$ npx thirdweb publish
```

jnsdls (commented):
2 nitpicks from me, other than that this seems OK - would still like for @adam-maj or @joaquim-verges  to also take a look :)

packages/cli/src/common/processor.ts
```diff
@@ -13,7 +13,7 @@ import path from "path";
 
 export async function processProject(
   options: any,
-  command: "deploy" | "release",
+  command: "deploy" | "publish",
```
joaquim-verges, line 16: did we change the dashboard url also? cause this will open /contracts/publish/...

PR #575, opened 2023-02-14
[Unity Bridge] Add support for Auth to unity-js-bridge

adam-maj:


packages/unity-js-bridge/src/thirdweb-bridge.ts
```diff
@@ -130,7 +142,7 @@ class ThirdwebBridge implements TWBridge {
     }
     const walletInstance = this.walletMap.get(wallet);
     if (walletInstance) {
-      await walletInstance.connect({ chainId });
+      await walletInstance.connect(chainId);
```
joaquim-verges, line 145: why the change here? I switched it to be an object coz there will be more options for other wallets

packages/unity-js-bridge/src/thirdweb-bridge.ts
```diff
@@ -196,6 +208,15 @@ class ThirdwebBridge implements TWBridge {
       }
     }
 
+    if (addrOrSDK.startsWith("auth")) {
```
joaquim-verges, line 211: you need to do the change in unity in `Wallet.cs` in the `Authenticate()` method, switch from sdk/auth/login to just auth/login

packages/unity-js-bridge/src/thirdweb-bridge.ts
adam-maj: It breaks the types & doesn't build for me

PR #579, opened 2023-02-14
[SDK] - mark `fromPrivateKey` as public

jnsdls:


jnsdls:
does not require changeset

PR #584, opened 2023-02-15
Remove dead code from `react` package

MananTank:


joaquim-verges:
@MananTank @iketw Some comments on usability, DX:

- `supportedWallets` should not be mandatory in `ThirdwebProvider` the comment says it'll use some defaults but still requires me to pass something
- `useContract` is should be re-exported in `/react` - don't make me import anythign from `/react-core` since it'll be an internal dependency
- `ConnectWallet` requires a `theme` attribute? feel like we could also default to something (prob dark) and not require any props.
- passing an empty array for `supportedWallets` results in a broken UI, should probably throw an error if empty array is passed explicitely
- code comment for `supportedWallets` should give me a few examples of wallets to pass (helps ppl find the rest of them) - its also not obvious that i need to pass class names rather than instances

on Device wallet:

- is the second login flow implemented? it always shows the "create" password screen for me

MananTank:
> is the second login flow implemented? it always shows the "create" password screen for me

That's weird. It works just fine for me: 

First Time ( clearing local storage )

image

Then:

image

packages/wallets/src/evm/connectors/device-wallet/index.ts
```diff
@@ -1,15 +1,16 @@
...
 } from "../../wallets/device-wallet";
-import type { Chain } from "@thirdweb-dev/chains";
 import type { Signer } from "ethers";
 import { providers } from "ethers";
 
 export type DeviceWalletConnectorOptions = {
-  chain: Pick<Chain, "chainId" | "rpc">;
+  chain: Chain;
   wallet: DeviceWalletImpl;
+  chains: Chain[];
```
joaquim-verges, line 13: prob don't need this, should just need one chain at a time

if we make `switchChain()` take a chain object it should work nicely

packages/wallets/src/evm/connectors/device-wallet/index.ts
```diff
@@ -1,15 +1,16 @@
...
   DeviceWalletConnectionArgs,
   DeviceWalletImpl,
 } from "../../wallets/device-wallet";
-import type { Chain } from "@thirdweb-dev/chains";
 import type { Signer } from "ethers";
 import { providers } from "ethers";
 
 export type DeviceWalletConnectorOptions = {
-  chain: Pick<Chain, "chainId" | "rpc">;
+  chain: Chain;
```
joaquim-verges, line 11: Pick didnt work for you here?

packages/wallets/src/evm/wallets/base.ts
```diff
@@ -48,15 +48,6 @@ export abstract class AbstractBrowserWallet<
...
     }
 
-    const connector = await this.getConnector();
-    const isConnected = await connector.isConnected();
-
-    // return if already connected
-    if (isConnected) {
-      this.#subscribeToEvents(connector);
-      return;
-    }
```
iketw: Why remove this code from here?

MananTank: Because of this, the wallet.addEventListener() does not work

packages/wallets/src/evm/connectors/device-wallet/index.ts
MananTank: It _can_ work, but since all wallets are instantiated the same way in `react-core` package, we would need to use that type in all wallets instead. 

image

Currently this `chain` option is only used by device-wallet. This will be removed soon so that it works similar to other wallets ( provide chain when connecting - not when creating wallet instance )

Now that we are planning to use the `Chain` type from `thirdweb-dev/chains` - we are gonna stop using the `Chain` type from `@wagmi/core` altogether

packages/wallets/src/evm/wallets/base.ts
MananTank: when auto-connecting a wallet, `isConnected` is `true` and because of that the call to `this.connect(...)` is skipped - but it shouldn't be. because of this connector's `connect` method is not called when auto-connecting and because of that wallet does not setup event listeners for the provider ( which is done in connector.connect method )

image

MananTank: checking the `connector.isConnected()` ( here the connecotr is wagmi adapter)  internally forwards it to `connector.isAuthorized()` 

image

Here, it checks for storage, provider and provider address and return true/false based on that. In case of autoconnect - we get the provider address and so it returns `true` - but these ends up skipping the steps required in `connect` method mentioned in previous comment.

image

So to fix this, I removed the condition and we always use the "connect flow" for autoconnect instead of using a short cut

packages/react-core/src/core/hooks/wallet-hooks.ts
```diff
@@ -0,0 +1,125 @@
+import { useThirdwebWallet } from "../providers/thirdweb-wallet-provider";
+import invariant from "tiny-invariant";
+
+/**
+ * @returns the current active wallet instance
+ */
+export function useActiveWallet() {
```
jnsdls, line 7: let's rename this to `useWallet()`

packages/react-core/src/core/hooks/wallet-hooks.ts
```diff
@@ -0,0 +1,125 @@
...
+  );
+  return context.disconnect;
+}
+
+/**
+ *
+ * @returns the connection status of the wallet
+ *
+ * It can be one of the following:
+ * 1. `unknown` - when wallet connection status is not yet known
```
jnsdls, line 60: what causes `unknown`? just a loading state? can we make it `loading` in that case?

packages/react-core/src/core/hooks/wallet-hooks.ts
```diff
@@ -0,0 +1,125 @@
...
+    "useConnectionStatus() must be used within a <ThirdwebProvider/>",
+  );
+  return context.connectionStatus;
+}
+
+/**
+ *
+ * @returns a method to create an instance of given wallet class
+ */
+export function useCreateWalletInstance() {
```
jnsdls, line 79: is this a public API? what is it used for?

packages/react-core/src/core/hooks/wallet-hooks.ts
```diff
@@ -0,0 +1,125 @@
...
+    "useSwitchChain() must be used within a <ThirdwebProvider/>",
+  );
+  return context.switchChain;
+}
+
+/**
+ *
+ * @returns a method to get the chainId of currently connected network/chain
+ */
+export function useActiveChainId() {
```
jnsdls, line 105: if we have it available can we instead return the chain object? `useActiveChain()` in that case?

packages/react-core/src/core/providers/thirdweb-provider.tsx
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * A partial map of chainIds to rpc urls to use for certain chains
+   * If not provided, will default to the rpcUrls of the chain objects for the supported chains
+   * @deprecated - use `chains` instead
+   */
+  chainRpc?: Record<number, string>;
+
+  theme?: "light" | "dark";
+
+  createWalletStorage: CreateAsyncStorage;
```
jnsdls, line 103: is this required? We should probably default it to the browser option (since tecnically people could use this directly stil)

packages/react-core/src/core/providers/thirdweb-provider.tsx
```diff
@@ -0,0 +1,188 @@
...
+  );
+};
+
+const ThirdwebSDKProviderWrapper = <TChains extends Chain[]>({
+  children,
+  ...props
+}: React.PropsWithChildren<
+  Omit<ThirdwebSDKProviderProps<TChains>, "signer">
+>) => {
+  const signer = useWalletSigner();
```
jnsdls, line 181: can this hook just be called `useSigner()`?

packages/react-core/src/core/providers/thirdweb-wallet-provider.tsx
```diff
@@ -0,0 +1,293 @@
...
+} from "react";
+
+let coordinatorStorage: AsyncStorage;
+
+type NonNullable<T> = T extends null | undefined ? never : T;
+type WalletConnectParams<W extends SupportedWallet> = NonNullable<
+  Parameters<InstanceType<W>["connect"]>[0]
+>;
+
+type ConnectionStatus = "unknown" | "connected" | "disconnected" | "connecting";
```
jnsdls, line 25: same question RE `loading` vs `unknown`?

packages/react-core/src/core/providers/thirdweb-wallet-provider.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+    const update = async () => {
+      const _signer = await activeWallet.getSigner();
+      const _chainId = await _signer.getChainId();
+      setSigner(_signer);
+      setActiveChainId(_chainId);
+    };
+
+    activeWallet.addListener("change", () => {
+      update();
```
jnsdls, line 260: this can technically fire multiple times queing up `update` calls and potentially leading to incorrect states (second update could resolve before first one)

it's probably a non-issue but calling out that it *could* happen

packages/react-core/src/evm/hooks/contracts/index.ts
```diff
@@ -1,12 +1,12 @@
...
-export * from "./useMarketplace";
-export * from "./useNFTCollection";
-export * from "./usePack";
-export * from "./useToken";
-export * from "./useTokenDrop";
-export * from "./useVote";
-export * from "./useSplit";
-export * from "./useMultiwrap";
-export * from "./useSignatureDrop";
+export { useEditionDrop } from "./useEditionDrop";
```
jnsdls, line 1: nice cleanliness update thx :)

packages/react-core/src/evm/hooks/wallet.ts
```diff
@@ -93,3 +95,33 @@ export function useAddress(): string | undefined {
...
+ * const App = () => {
+ *   const chain = useActiveChain()
+ *
+ *   return <div>{chain.chainId}</div>
+ * }
+ * ```
+ * @public
+ */
+export function useActiveChain(): Chain | undefined {
+  const chainId = useThirdwebConnectedWalletContext().chainId;
```
jnsdls, line 120: if we make the context hold the full chain object (if that's possible) we can save ourselves a find call here

packages/react-core/src/evm/providers/thirdweb-sdk-provider.tsx
```diff
@@ -124,7 +124,7 @@ const WrappedThirdwebSDKProvider = <
     type ForcedChainType = {
       rpc: string[];
       chainId: number;
-      nativeCurrency: { symbol: string; name: string; decimals: number };
+      nativeCurrency: { symbol: string; name: string; decimals: 18 };
```
jnsdls, line 127: nah keep it number, there are chains out there that we support that have 16 decimals

packages/react-core/src/index.ts
```diff
@@ -1,2 +1,258 @@
 // main entry point = evm
-export * from "./evm";
+
```
jnsdls, line 2: this seems like a lot of duplication, we're already doing this in `/evm/index.ts` no? or does doing export * from `./evm` break tree shaking or something?

jnsdls (commented):
couple of comments and questions :)

packages/react-core/src/core/hooks/wallet-hooks.ts
MananTank: ok

packages/react-core/src/core/hooks/wallet-hooks.ts
MananTank: status starts out with `unknown` - because when the app is loading, we just don't know status

Here are two possible scenarios when app is loading:

1. if wallet was previously connected + autoconnect enabled, the status goes from `unknown` => `connecting` => `connected`
2. if no wallet was connected before or autoconnect is not enabled, the status goes from `unknown` => `disconnected`


We can make it `loading` instead of `unknown` but it sounds a bit similar to `connecting` - it feels like we are trying to "connect" but we are not. Let me know what you think, I'm fine with either

packages/react-core/src/core/hooks/wallet-hooks.ts
MananTank: It _can_ be a public API, user can use it to create a wallet instance. 

This abstracts away the complexity of creating a wallet instance ( Wallet constructor takes a bunch of things as argument )

```js
const createInstance = useCreateWalletInstance();

// when we need to create an instance, we do:
const metamask = createInstance(MetamaskWallet)

// MetamaskWallet is a class
// metamask is wallet instance
```

packages/react-core/src/core/hooks/wallet-hooks.ts
MananTank: we have a `useActiveChain` hook also, we can keep this one and delete the `useActiveChainId`

packages/react-core/src/core/providers/thirdweb-provider.tsx
MananTank: this is required in `react-core`, but not in `react` or `react-native`
they export their own version of `ThirdwebProvider` - where `createWalletStorage` is omitted from props.

react/src/evm/providers/thirdweb-provider.tsx :
 
image

packages/react-core/src/evm/hooks/wallet.ts
MananTank: @jnsdls Yeah, it's on my todo list. we will do that

packages/react-core/src/index.ts
MananTank: Agree, It's a huge pain. I wanted to see whether it helps with tree-shaking or not but did not test it.

I tested the `export *` vs named exports now - looks like both have the tree-shaking issue. As soon as something from `@thirdweb-dev/sdk` or `@thirdweb-dev/auth` is included in export - it instantly kills the tree-shaking. 

Example: 

image

packages/react-core/src/core/providers/thirdweb-provider.tsx
MananTank: We already have a `useSigner()` hook. This is a duplicate and only used one-time

`useWalletSigner()` is used to provider the signer to `ThirdwebConnectedWalletContext` and `useSigner()` returns the signer from `ThirdwebConnectedWalletContext`. 

`useSigner()` is the one that is exported. `useSigner()` works in both cases - when app is wrapped with `<ThirdwebProvider/>` or `<ThirdwebSDKProvider/>`. but `useWalletSigner()` only works if the app is wrapped with `<ThirdwebProvider/>`

So What I can do it remove the `useWalletSigner()` hook and just use `useThirdwebWallet().signer` at a single place this is being used: 

image

packages/react-core/src/evm/providers/thirdweb-sdk-provider.tsx
MananTank: done

packages/react-core/src/core/providers/thirdweb-provider.tsx
MananTank: removed useWalletSigner()

packages/react-core/src/core/providers/thirdweb-provider.tsx
```diff
@@ -178,7 +180,7 @@ const ThirdwebSDKProviderWrapper = <TChains extends Chain[]>({
 }: React.PropsWithChildren<
   Omit<ThirdwebSDKProviderProps<TChains>, "signer">
 >) => {
-  const signer = useWalletSigner();
+  const signer = useThirdwebWallet()?.signer;
```
iketw, line 183: I liked the `useSigner` hook :(

iketw:
./release-pr

packages/react-core/src/core/providers/thirdweb-provider.tsx
MananTank: @iketw We still have it. I just removed it's duplicate which was only being used a single time internally

iketw:
./release-pr

iketw:
./release-pr

iketw:
./release-pr

packages/react-core/src/core/providers/thirdweb-wallet-provider.tsx
iketw: You just need to remove the listener in a returning function in the useEffect()

iketw (approved):
ship it! 🚀🚀

joaquim-verges:
MERRRRRRGE

PR #587, opened 2023-02-15
fix: chainId of polygon zkEVM testnet

avneesh0612:


jnsdls:
need a better way to do the override / preferably fix the source data

PR #590, opened 2023-02-16
[SDK] - allow passing Chain object to initialize SDK

jnsdls:


packages/sdk/src/evm/core/sdk.ts
```diff
@@ -159,10 +159,23 @@ export class ThirdwebSDK extends RPCConnectionHandler {
...
-    network: NetworkInput,
+    network: NetworkInput | Chain,
     options: SDKOptions = {},
     storage: ThirdwebStorage = new ThirdwebStorage(),
   ) {
+    if (
+      typeof network !== "string" &&
+      typeof network !== "number" &&
+      !ethers.Signer.isSigner(network) &&
+      !ethers.providers.Provider.isProvider(network)
```
joaquim-verges, line 170: Wonder if it's safer to check if object and has rpc property in there with at least one rpc?

packages/sdk/test/evm/sdk-init.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Goerli, Fncy } from "@thirdweb-dev/chains";
+// import { signers, storage } from "./before-setup";
+// import { AddressZero } from "@ethersproject/constants";
+// import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
+import { expect } from "chai";
+
+global.fetch = require("cross-fetch");
+
+describe("SDK Initialization", async () => {
+  // let adminWallet: SignerWithAddress, samWallet: SignerWithAddress;
```
joaquim-verges, line 11: Cleanup

jnsdls: was gonna also test it with signer etc but yeah

packages/sdk/src/evm/core/sdk.ts
jnsdls: I'm scared that `Signer` JS typeof would also be `object` (same for provider)

PR #598, opened 2023-02-17
Create .nvmrc for version v18.13.0

agustinkassis:
Able to use the proper Node version by simply running

```bash
nvm use
```


jnsdls:
hm why is this required, we support node 16+

PR #604, opened 2023-02-18
[sdk] - fix possible crash in StaticJsonRpcBatchProvider

jnsdls:


packages/sdk/test/evm/marketplace-v3.test.ts
```diff
@@ -172,7 +172,7 @@ describe("Marketplace V3", async () => {
     endTimestamp: Date = new Date(
       startTimestamp.getTime() + 5 * 24 * 60 * 60 * 1000,
     ),
-    reservePricePerToken = 0.05,
+    // reservePricePerToken = 0.05,
```
iketw, line 175: delete comment?

packages/sdk/src/evm/lib/static-batch-rpc.ts
```diff
@@ -66,6 +66,10 @@ export class StaticJsonRpcBatchProvider extends providers.StaticJsonRpcProvider
         // on whether it was a success or error
         batch.forEach((inflightRequest_, index) => {
           const payload = result[index];
+          // if there is no payload, reject the request
+          if (!payload) {
```
iketw, line 70: Did we see this crash somewhere? 
Do you think this error is because we got less results than items in the batch or because we are getting undefined for some of them? If it's the first case then all results after the first failure will be assigned to incorrect requests.

packages/sdk/src/evm/lib/static-batch-rpc.ts
```diff
@@ -66,6 +66,10 @@ export class StaticJsonRpcBatchProvider extends providers.StaticJsonRpcProvider
...
+            inflightRequest_.reject(new Error("No response for request"));
```
iketw, line 71: Should this be an if/else statement with `if(payload.error)` ? I think after rejecting we don't want to continue the execution

packages/sdk/test/evm/marketplace-v3.test.ts
```diff
@@ -918,7 +918,7 @@ describe("Marketplace V3", async () => {
         "The buyer should start with no tokens",
       );
       // await marketplaceContract.makeOffer(directListingId, 0.05, 1);
-      const offerId = await makeOffer(dummyNftContract.getAddress(), 0, 1);
+      // const offerId = await makeOffer(dummyNftContract.getAddress(), 0, 1);
```
iketw, line 921: delete comment

packages/sdk/test/evm/marketplace-v3.test.ts
```diff
@@ -934,10 +934,10 @@ describe("Marketplace V3", async () => {
 
     it("should allow offers to be made for listed tokens", async () => {
       sdk.updateSignerOrProvider(bobWallet);
-      const bobOfferId = await makeOffer(dummyNftContract.getAddress(), 0, 1);
+      // const bobOfferId = await makeOffer(dummyNftContract.getAddress(), 0, 1);
```
iketw, line 937: delete comment. Same for the others comments in this file

packages/sdk/src/evm/lib/static-batch-rpc.ts
jnsdls: yep, this crashed some folks in dashboard and now that we finally have sourcemaps in sentry we caught it :)

packages/sdk/src/evm/lib/static-batch-rpc.ts
jnsdls: good catch, will fix

jnsdls: actually no, we want to process each inflight request separately

iketw: Yeah, bit if there's no payload we should continue to the next request, if I understand this correctly

jnsdls: and we do, we're in a forEach loop there. we're resolving / rejecting n promises

jnsdls: added an explicit `else` case

PR #610, opened 2023-02-21
[chains] - testnet added

jnsdls:


packages/chains/chains/84531.ts
```diff
@@ -0,0 +1,34 @@
...
+      512,
+      256,
+      128,
+      64,
+      32,
+      16
+    ]
+  },
+  "testnet": true,
+  "slug": "base-testnet-goerli"
```
joaquim-verges, line 33: @jnsdls can we just do `base-goerli` instead of `base-testnet-goerli` ?

jnsdls: sure if I add another custom rule lol

PR #623, opened 2023-02-24
[SDK] Cleanup signer and provider logic & add exhaustive test cases

adam-maj:
Several changes to improve the code quality/safety of our SDK connection logic:
- Fixed the `fromWallet` and `fromSigner` methods to connect the signer to the provided network if applicable
- Added proper helper function (based on ethers) to detect whether a string is an RPC URL (we were detecting this incorrectly, and only counting `http` URLs as RPCs, ignoring `ws` URLs.
- Removed duplicated logic in `ThirdwebSDK` constructor, instead handling all `network` parsing in the `RPCConnectionHandler` class.
- Exhaustive test cases of all SDK connection flows, including connecting via `chainName`, `chainId`, `Chain` configuration, HTTP `rpcUrl`, and WebSocket `rpcUrl`, as well as using the `fromPrivateKey`, `fromSigner`, and `fromWallet` static methods.


jnsdls:
@adam-maj make sure this works correctly still with the react sdk & dashboard. Who knows what they're relying on in terms of behavior.

packages/sdk/src/evm/core/sdk.ts
```diff
@@ -90,12 +91,18 @@ export class ThirdwebSDK extends RPCConnectionHandler {
...
     options: SDKOptions = {},
     storage: ThirdwebStorage = new ThirdwebStorage(),
   ): ThirdwebSDK {
-    const sdk = new ThirdwebSDK(network || signer, options, storage);
-    sdk.updateSignerOrProvider(signer);
+    // If we are given a network, connect the signer to that network
+    let signerWithProvider = signer;
+    if (network) {
+      const provider = getChainProvider(network, options);
+      signerWithProvider = signer.connect(provider);
```
joaquim-verges, line 102: This can throw for certain signers. We should also check if there's already a provider set, otherwise we might overriding custom providers here

packages/sdk/src/evm/constants/urls.ts
```diff
@@ -108,68 +119,112 @@ export function getChainProvider(
...
+              new providers.JsonRpcBatchProvider(rpcUrl);
 
+          // Save the provider in our cache
+          RPC_PROVIDER_MAP.set(seralizedOpts, newProvider);
+          return newProvider;
         case "ws":
-          return new providers.WebSocketProvider(network, chainId);
-        default:
-          return ethers.getDefaultProvider(network);
+        case "wss":
```
joaquim-verges, line 219: oof nice catch here

packages/sdk/src/evm/core/classes/rpc-connection-handler.ts
```diff
@@ -85,37 +85,39 @@ export function getSignerAndProvider(
...
+  } else {
+    // Here, we must have a ChainOrRpcUrl, which is a chain name, chain id, rpc url, or chain config
+    // All of which, getChainProvider can handle for us
+    provider = getChainProvider(network, options);
   }
 
   if (options?.readonlySettings) {
-    provider = getReadOnlyProvider(
+    // If readonly settings are specified, then overwrite the provider
+    provider = getProviderFromRpcUrl(
```
joaquim-verges, line 104: I worry that we're always overriding the provider here, even when a provider is passed explicitely.

Like if I pass a custom WalletConnect UniversalProvider here for example, you will break me by using a regular json rpc provider.

I think the previous logic ONLY used the read only provider if one wasn't passed explicitely. that's safer i think.

So this would be:
`if (options?.readonlySettings && !provider)`

adam-maj: The previous logic always overwrote - but this suggestion is better I think

packages/sdk/src/evm/core/sdk.ts
adam-maj: If they specify a network + there's already a provider connected to the signer + the provider is on the wrong network, we should probably still overwrite. Will update for that case.

adam-maj: oof, that would make it async hmm

adam-maj: In that case, assuming we cant make `fromSigner` async, I actually prefer always setting provider ourselves then if they specify network. Network is optional here, so if they don't pass it then I think of it as "use the provider on my signer". But if they do pass it, I think of it as "connect my signer to this new provider I'm telling you to."

Ex: they can pass an RPC URL to the network argument, and it would be very weird if we just ignored it and didn't connect their signer to that RPC URL just because it already had a provider - wdyt?

jnsdls: If you connect a random provider to a signer that can throw for some signers if the provider is not on the same chain as the signer

adam-maj: I think this should throw an error in this case

joaquim-verges: might be ok because network is optional. guess we can try it

packages/sdk/src/evm/constants/urls.ts
```diff
@@ -108,68 +119,112 @@ export function getChainProvider(
     );
   }
 
-  return getReadOnlyProvider(rpcUrl, chainId);
+  return getProviderFromRpcUrl(rpcUrl, chainId);
```
joaquim-verges, line 122: nice cleanup here, like the comments to follow along

packages/sdk/src/evm/core/classes/rpc-connection-handler.ts
joaquim-verges: @adam-maj hm you're right the previous logic always overwrote. I guess its fine to leave it as is. its controllable via sdk options anyways

packages/sdk/src/evm/core/sdk.ts
joaquim-verges: adding the `!signer.provider` seems fine tho

joaquim-verges (commented):
i think this is decent cleanup, would be ok to merge- but its on you to double test everything

adam-maj:
I tested this all with dashboard @joaquim-verges which is the context on some of the fixes I just made today (adding the `!provider` to `readOnlyRpc` actually breaks dashboard - if we pass readOnlyRpc, we need to overwrite provider always.

PR #624, opened 2023-02-25
[SDK] Extension code comments

joaquim-verges:


packages/sdk/src/evm/core/classes/contract-royalty.ts
```diff
@@ -44,11 +44,13 @@ export class ContractRoyalty<
   }
 
   /**
-   * Gets the royalty recipient and BPS (basis points) of the contract
+   * Gets the royalty recipient and fee
```
nachoiacovino, line 47:
```suggestion
   * Get the royalty recipient and fee
```

I think we do "Get" instead of Gets everywhere?

packages/sdk/src/evm/core/classes/contract-royalty.ts
```diff
@@ -64,11 +66,13 @@ export class ContractRoyalty<
   }
 
   /**
-   * Gets the royalty recipient and BPS (basis points) of a particular token
+   * Gets the royalty recipient and fee of a particular token
```
nachoiacovino, line 69: same here

PR #628, opened 2023-02-25
[SDK] Add getContract and getContractFromAbi functions

adam-maj:


packages/sdk/src/evm/functions/contract/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  getCachedStorage,
+  inContractCache,
+} from "../core/cache";
+import { GetContractFromAbiParams, GetContractParams } from "./types";
+import { resolveContractType } from "./utils";
+
+export async function getContractFromAbi(
+  params: GetContractFromAbiParams,
+): Promise<ValidContractInstance> {
+  if (inContractCache(params.address)) {
```
jnsdls, line 20: I believe this needs to include more than just the address, otherwise you may get the same contract instance back for contract `0x123...` and `0x123...` if they're on different chains, so it needs to include at least chainId + address

packages/sdk/src/evm/functions/contract/index.ts
```diff
@@ -0,0 +1,102 @@
...
+import { getSignerAndProvider } from "../../core/classes/rpc-connection-handler";
+import { ValidContractInstance } from "../../core/types";
+import { AbiSchema } from "../../schema/contracts/custom";
+import {
+  cacheContract,
+  getCachedContract,
+  getCachedStorage,
+  inContractCache,
+} from "../core/cache";
+import { GetContractFromAbiParams, GetContractParams } from "./types";
```
jnsdls, line 14: instead of breaking out types, and functions let's break the files out by function name and co-locate the types in the same file:

```
functions/
  getContract.ts
  getContractFromAbi.ts
  ...
```

I think that will lead to a cleaner organization and you can always easily import all functions from within other functions by doing `import ... from "./function"`
   

Also I believe `index.ts` is causing a lot of problems for us and we should start moving away from it in favor of real paths and file names and proper named exports

packages/sdk/src/evm/functions/contract/index.ts
```diff
@@ -0,0 +1,102 @@
...
+      params.address,
+      AbiSchema.parse(parsedAbi),
+      provider,
+      params.sdkOptions,
+      getCachedStorage(params.storage),
+    ),
+    getCachedStorage(params.storage),
+    params.sdkOptions,
+    (await provider.getNetwork()).chainId,
+  );
```
jnsdls, line 45: you could optimize these to parallelize both the chainId resolution and the gompositepluginabi by calling them within a Promise.all before the `new SmartContract` and then passign them in sync

packages/sdk/src/evm/functions/contract/index.ts
```diff
@@ -0,0 +1,102 @@
...
+import { resolveContractType } from "./utils";
+
+export async function getContractFromAbi(
+  params: GetContractFromAbiParams,
+): Promise<ValidContractInstance> {
+  if (inContractCache(params.address)) {
+    return getCachedContract(params.address);
+  }
+
+  // TODO: Remove support for SDKOptions
```
jnsdls, line 24: what do we still need it for in here?

packages/sdk/src/evm/functions/contract/index.ts
```diff
@@ -0,0 +1,102 @@
...
+    params.sdkOptions,
+    (await provider.getNetwork()).chainId,
+  );
+
+  cacheContract(params.address, contract);
+  return contract;
+}
+
+export async function getContract(params: GetContractParams) {
+  if (inContractCache(params.address)) {
```
jnsdls, line 52: same as above, this requires more than just address in order to work reliably

packages/sdk/src/evm/functions/contract/index.ts
```diff
@@ -0,0 +1,102 @@
...
+    params.sdkOptions,
+  );
+
+  if (!params.contractTypeOrAbi || params.contractTypeOrAbi === "custom") {
+    const contractType = await resolveContractType({
+      address: params.address,
+      provider,
+    });
+    if (contractType === "custom") {
+      // Not even worth cacheing publisher since we would have to cache by network, which requires an RPC call
```
jnsdls, line 67: why would we not cache publisher by network?

jnsdls: globally?

packages/sdk/src/evm/functions/contract/types.ts
```diff
@@ -0,0 +1,21 @@
+import { NetworkInput, PrebuiltContractType } from "../../core/types";
+import { SDKOptions } from "../../schema/sdk-options";
+import { ThirdwebStorage } from "@thirdweb-dev/storage";
+import { ContractInterface } from "ethers";
+
+export type GetContractFromAbiParams = {
+  address: string;
+  abi: ContractInterface;
+  network: NetworkInput;
```
jnsdls, line 9: I think I'd prefer for this to just become `chain` tbh?

jnsdls: in terms of the actual input type I mean, param naming aside

packages/sdk/src/evm/functions/core/cache.ts
```diff
@@ -0,0 +1,30 @@
+import { ValidContractInstance } from "../../core/types";
+import { ThirdwebStorage } from "@thirdweb-dev/storage";
+
+const CONTRACT_CACHE = new Map<string, ValidContractInstance>();
+
+const STORAGE_CACHE = new Map<"cache", ThirdwebStorage>();
+STORAGE_CACHE.set("cache", new ThirdwebStorage());
```
jnsdls, line 7: this seems extremely odd, why would we immediately set something on the cache? did you want to just use a singleton?

jnsdls (commented):
left some thoughts, excited for the start of this!

packages/sdk/src/evm/functions/contract/index.ts
adam-maj: Nice I prefer this pattern

packages/sdk/src/evm/functions/contract/index.ts
adam-maj: Indeed

packages/sdk/src/evm/functions/contract/index.ts
adam-maj: For temporary compatibility since we take thirweb/alchemy/infura API keys in here which works with our chains RPC URL setup

adam-maj: As well as other small things like read only RPC which the provider functions use (for now)

packages/sdk/src/evm/functions/contract/index.ts
adam-maj: b/c checking if we have a cached `ContractPublisher` requires making an RPC request to get the current network we're on, which is an extra request and potentially slower than just instantiating a new one (?).

packages/sdk/src/evm/functions/contract/types.ts
adam-maj: That would be ideal - for now I like this for complete compatibility with current SDK + I think being able to pass in an RPC URL is nice. Wdyt?

packages/sdk/src/evm/functions/core/cache.ts
adam-maj: Yes this is very weird lmao not sure what I was doing here. Should just be a singleton lmao.

packages/sdk/src/evm/functions/contract/index.ts
adam-maj: I guess we'll be making that request to cache contract now anyway

packages/sdk/src/evm/functions/core/cache.ts
jnsdls: fwiw I could see a usecase for multiple instances of storage

joaquim-verges (commented):
seems very promising!

PR #631, opened 2023-02-25
[SDK] Detect appURI

nachoiacovino:


packages/sdk/src/evm/contracts/smart-contract.ts
```diff
@@ -286,4 +292,11 @@ export class SmartContract<TContract extends BaseContract = BaseContract>
     }
     return undefined;
   }
+
+  private detectAppURI() {
+    if (detectContractFeature<AppURI>(this.contractWrapper, "AppURI")) {
```
joaquim-verges, line 297: Here you should also check for the contract metadata extension

nachoiacovino: so check for both? you can have appURI without contract metadata

joaquim-verges: one OR the other.

`appUri = detect(AppUri) || detect(ContractMetadata)`

joaquim-verges: @nachoiacovino @furqanrydhan ^

PR #632, opened 2023-02-25
[Multi] Document marketplace v3 with examples

nachoiacovino:
Also fix the wrong quote thing and documented PackVRF


nachoiacovino:
doing a couple more changes on the same PR, marketplace v3 on sdk has no detections :D

joaquim-verges (approved):

PR #644, opened 2023-02-28
Allow passing gatewayUrls as array and directly

adam-maj:


packages/sdk/src/evm/common/storage.ts
```diff
@@ -0,0 +1,17 @@
+import { StorageConfigInput, StorageConfigSchema } from "../schema";
+import { ThirdwebStorage } from "@thirdweb-dev/storage";
+
+function isStorage(config: StorageConfigInput): config is ThirdwebStorage {
+  return config instanceof ThirdwebStorage;
```
jnsdls, line 5: this is scary because if I install my own thirdweb storage package this will *not* match

jnsdls (commented):
might work?

jnsdls (approved):
LGTM

PR #657, opened 2023-03-01
[SDK] Add prepare transaction to contract

adam-maj:


packages/sdk/src/evm/core/classes/contract-wrapper.ts
```diff
@@ -269,6 +269,94 @@ export class ContractWrapper<
...
+    const ethersFnName = `${functionName as string}(${fn.inputs
+      .map((i) => i.type)
+      .join()})` as TMethod;
+
+    // check if the function exists on the contract, otherwise use the name passed in
+    const fnName =
+      ethersFnName in this.readContract.functions ? ethersFnName : functionName;
+
+    // TODO validate each argument
+    if (fn.stateMutability === "view" || fn.stateMutability === "pure") {
```
joaquim-verges, line 335: Can we consolidate this code with .call? It's the same setup no?

packages/sdk/src/evm/core/classes/contract-wrapper.ts
```diff
@@ -269,6 +269,94 @@ export class ContractWrapper<
...
+      const signer = this.getSigner();
+      if (!signer) {
+        throw new Error("No signer found");
+      }
+
+      const signedTx = await signer.signTransaction(tx);
+
+      return {
+        transaction: tx,
+        signature: signedTx,
```
joaquim-verges, line 355: Hmmm i was hoping to get a transaction object here, something i can call estimate/simulate/encode on

adam-maj: This setup is jake specific - I actually don't want to write like this I think - really what I want is for these to call into `Transaction`, but that causes a circular dependency which makes test cases fail for now so I just did the quick solution to get a working version for Jake

Also won't have duplicated `call` code

PR #660, opened 2023-03-01
Fix Wrong Currency symbol in non-default chains

MananTank:
Fixed: 

image

### DIFF
--- a/packages/react-core/src/evm/providers/thirdweb-sdk-provider.tsx
+++ b/packages/react-core/src/evm/providers/thirdweb-sdk-provider.tsx
@@ -147,10 +147,17 @@ const WrappedThirdwebSDKProvider = <
       }
     }
 
+    // TODO: find a better way to fix the type error
+    type ForcedChainType = {
+      rpc: string[];
+      chainId: number;
+      nativeCurrency: { symbol: string; name: string; decimals: 18 };
+    };
+
     const mergedOptions = {
       readonlySettings,
       ...sdkOptions,
-      chains: supportedChains,
+      supportedChains: supportedChains as any as ForcedChainType[],
     };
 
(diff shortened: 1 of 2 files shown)


packages/react-core/src/evm/providers/thirdweb-sdk-provider.tsx
```diff
@@ -147,10 +147,17 @@ const WrappedThirdwebSDKProvider = <
...
+      rpc: string[];
+      chainId: number;
+      nativeCurrency: { symbol: string; name: string; decimals: 18 };
+    };
+
     const mergedOptions = {
       readonlySettings,
       ...sdkOptions,
-      chains: supportedChains,
+      supportedChains: supportedChains as any as ForcedChainType[],
```
joaquim-verges, line 160: Oof good catch

joaquim-verges (approved):

packages/react-core/src/evm/providers/thirdweb-sdk-provider.tsx
iketw, line 160: Nice catch. There are some mandatory fields in the Chain type that we don't have here. I'm worried this may cause other issues down the line if we assume we have those fields, e.g. `testnet`, `slug`, `shortName`, `chain`, `name`. Should we maybe do a `getChains(supportedChains)` where we fetch the actual Chain obj based on the chaindId?

jnsdls: this should be fine for now

jnsdls (approved):

rushby:
Is this causing the currency symbol to display as ETH for non ? Seeing that issue currently on 3.9.5

PR #662, opened 2023-03-01
app uri

furqanrydhan:


packages/sdk/src/evm/contracts/prebuilt-implementations/pack.ts
```diff
@@ -66,6 +73,7 @@ export class Pack extends StandardErc1155<PackContract> {
 
   public abi: Abi;
   public metadata: ContractMetadata<PackContract, typeof PackContractSchema>;
+  public app: ContractAppURI<BaseContract>;
```
adam-maj, line 76: Any reason we're passing `BaseContract` into these? We should be passing in the contract specific type for the prebuilts - like in this case `ContractAppURI<PackContract>`

packages/sdk/src/evm/contracts/smart-contract.ts
```diff
@@ -316,12 +317,22 @@ export class SmartContract<TContract extends BaseContract = BaseContract>
...
+  private detectApp() {
+    const metadata = new ContractMetadata(
+      this.contractWrapper,
+      CustomContractSchema,
+      this.storage,
+    );
+
+    if (detectContractFeature<IAppURI>(this.contractWrapper, "AppURI")) {
+      return new ContractAppURI(this.contractWrapper, metadata);
+    } else if (
```
adam-maj, line 329: These can just be in a single if statement with an `||`

packages/sdk/src/evm/core/classes/contract-appuri.ts
```diff
@@ -45,12 +45,14 @@ export class ContractAppURI<TContract extends BaseContract>
...
    */
-  public async get() {
-    if (detectContractFeature<AppURI>(this.contractWrapper, "AppURI")) {
+  public async get(): Promise<string> {
+    if (detectContractFeature<IAppURI>(this.contractWrapper, "AppURI")) {
       return await this.contractWrapper.readContract.appURI();
     }
 
-    return (await this.metadata.get()).appURI || "";
+    const appUri = (await this.metadata.get()).app_uri;
```
adam-maj, line 53: This will throw an error if `app_uri` isn't on metadata - which can happen since metadata is untyped.

packages/sdk/src/evm/core/classes/contract-appuri.ts
```diff
@@ -45,12 +45,14 @@ export class ContractAppURI<TContract extends BaseContract>
...
+
+    return Promise.resolve(`ipfs://${appUri.split("/ipfs/")[1]}` || Promise.resolve(""));
```
adam-maj, line 55: These don't need to be wrapped in `Promise.resolve()` since the function is async.

Also we're assuming that the `appUri` the user passes is an IPFS hash. But they can pass an HTTPS or any other type of URL (like arweave or sia or some other storage system) and this will break. We can't just assume it's IPFS here.

packages/sdk/src/evm/contracts/smart-contract.ts
furqanrydhan: sounds good

packages/sdk/src/evm/contracts/prebuilt-implementations/pack.ts
furqanrydhan: sounds good

packages/sdk/src/evm/contracts/smart-contract.ts
furqanrydhan: didn't work, leaving it

PR #666, opened 2023-03-01
CLI improvements

shift4id:


package.json
```diff
@@ -22,7 +22,7 @@
     "lint": "turbo run lint",
     "fix": "turbo run fix",
     "clean": "turbo run clean && rm -rf node_modules",
-    "format": "prettier --write \"**/*.{ts,tsx,md}\"",
```
jnsdls, line 25: this root format should already be hitting all packages, right?

PR #669, opened 2023-03-03
[SDK] Add ENS resolution function and support on AddressInput

adam-maj:
Discussed w/ @jnsdls:
- It makes sense for ENS resolution to be a check/parsing on the `AddressInput`. If the passed value isn't a valid address, we check if it's a valid ENS name.
- Cache mapping of ENS -> resolved address using SWR so we don't have to keep making requests
- This means that all schemas using `AddressInput` will now need to use `.parseAsync()` (unavoidable to have an async step for this somewhere since ENS resolution requires network request)
- We also intercept all parameters going to ethers (via `Transaction`, which we can assume all transactions will eventually switch to, as well as `contract.call`) for ABI inputs with `address` values and use the `AddressInput` on them to do ENS transformation (non-blocking), and then pass them through to `ethers` always if it fails.


jnsdls:
@adam-maj btw this fails a **bunch** of test cases

adam-maj:
Better off just merging ENS in one PR

PR #678, opened 2023-03-04
Add CLI command to deploy dynamic smart contracts

nkrishang:
Add the `--dynamic` flag to the `deploy` command:
```bash
thirdweb deploy --dynamic
```
This let's a user select a router contract type (`upgradeable`, `non-upgradeable`, `thirdweb-router`) and then select the extension contracts to add to the selected router.


packages/cli/src/common/processor.ts
```diff
@@ -265,3 +345,57 @@ export function getUrl(hashes: string[], command: string) {
...
+        });
+    }
+
+		extensions.push({
+      metadata,
+			functions
+    });
+    extensionDeployArgs.push({
+      name: contract.name,
+      amount: 0,
```
joaquim-verges, line 388: what's amount here?

packages/cli/src/constants/router-contracts.ts
```diff
@@ -0,0 +1,37 @@
...
+export interface TWRouterParams {
+  extensionRegistry: string;
+  extensionNames: string[];
+}
+
+export const EXTENSION_REGISTRY_ADDRESS: string = "0x0000000000000000000000000000000000000000";
+
+export const ROUTER_CONTRACTS: Record<string, PreDeployedRouter> = {
+  "upgradeable": {
+    name: "RouterUpgradeableFactory",
```
joaquim-verges, line 23: when these show up in the CLI - we should have a little explanation of what they are next to each. Just like we do in `npx thirdweb detect`

packages/cli/src/core/interfaces/Extension.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface Extension {
+    metadata: ExtensionMetadata;
+    functions: ExtensionFunction[];
+}
+
+export interface ExtensionDeployArgs {
+    name: string;
+    amount: number;
+    salt: string;
+    bytecode: string;
```
joaquim-verges, line 23: there's a little bit of duplicated code with what's in the SDK (in yash's marketplace v3 branch) should add a TODO here to consolidate

PR #681, opened 2023-03-06
[SDK] Add ENS support to all functions

adam-maj:
Discussed w/ @jnsdls:
- It makes sense for ENS resolution to be a check/parsing on the `AddressOrEns`. If the passed value isn't a valid address, we check if it's a valid ENS name.
- Cache mapping of ENS -> resolved address using SWR so we don't have to keep making requests
- This means that all schemas using `AddressOrEns` will now need to use `.parseAsync()` (unavoidable to have an async step for this somewhere since ENS resolution requires network request)
- Everywhere that takes an address as input should now get parsed with the `resolveAdress` function


packages/sdk/src/evm/common/ens.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const ENS_CACHE = new Map<string, CachedEns>();
+
+// TODO: Respect SDK RPC configuration and don't pull straight from ethers
+const provider = new providers.JsonRpcProvider(
+  `https://ethereum.rpc.thirdweb.com/${DEFAULT_API_KEY}`,
+);
+
+async function refreshCache(ens: string): Promise<string | null> {
+  const address = await provider.resolveName(ens);
```
joaquim-verges, line 18: can you make sure this works for cb.id and ens with subdomains? like thirdweb.deployer.eth ?

packages/sdk/src/evm/contracts/prebuilt-implementations/pack.ts
```diff
@@ -287,22 +288,22 @@ export class Pack extends StandardErc1155<PackContract> {
             tokenMetadata.decimals,
           );
           erc20Rewards.push({
-            contractAddress: reward.assetContract,
+            contractAddress: reward.assetContract as Address,
```
joaquim-verges, line 291: wait does this mean TS will yell at me for passing a string address now?

adam-maj: No, `Address` is just a `string` type. But I initially made it with address as `0x${string}` just to have errors propagate up and try to make the repo compatible with `Address` / `AddressOrEns` types regardless of if the types switch out underneath. But switched it back to just `string` to prevent the user type errors

PR #706, opened 2023-03-10
[WIP] Add optimistic getBalance function

adam-maj:
Very WIP - this isn't the pattern we'll go with. But for reference, standard `sdk.erc721.balanceOf` took `~1.192s` vs the `getBalance` taking `~143.15ms` so around a 10x improvement on this call.

### DIFF
--- a/packages/sdk/src/evm/functions/getBalance.ts
+++ b/packages/sdk/src/evm/functions/getBalance.ts
@@ -1,0 +1,23 @@
+import { getSignerAndProvider } from "../core/classes/rpc-connection-handler";
+import { NetworkInput } from "../core/types";
+import { AddressOrEns } from "../schema/shared";
+import type { IERC20 } from "@thirdweb-dev/contracts-js";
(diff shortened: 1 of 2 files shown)


packages/sdk/src/evm/functions/getBalance.ts
```diff
@@ -0,0 +1,23 @@
...
+  network: NetworkInput;
+  contractAddress: AddressOrEns;
+  walletAddress: AddressOrEns;
+};
+
+export async function getBalance(params: GetBalanceParams): Promise<BigNumber> {
+  const [, provider] = getSignerAndProvider(params.network, {});
+  const contract = new Contract(
+    params.contractAddress,
+    IERC20Abi,
```
jnsdls, line 18: I know this is a quick version to just play with it but I think we'll get this ABI (or even the pure function to call itself) from the `Extension` later

jnsdls (commented):
very excite

packages/sdk/src/evm/functions/getBalance.ts
adam-maj: yes would be much preferable

PR #708, opened 2023-03-10
[REACT] Fix/unable to override chain

ThianHooi:
#### Issue
- When trying to override certain value (example: RPC URL) when passing props to the `activeChain` in `ThirdwebProvider` component, I found that the RPC calls are not done through the custom RPC URL provided, instead default thirdweb RPC URL is used
- Found out that the issue happens [here](https://github.com/thirdweb-dev/js/blob/2ec2802119a3c375a1adaed1263ae1eae1384865/packages/react/src/evm/providers/thirdweb-provider.tsx#L248). When merging `supportedChains` and `activeChain`, `activeChain` with overridden value should replace the respective chain config in `supportedChain`. This causes the [RPC URLs mapping](https://github.com/thirdweb-dev/js/blob/2ec2802119a3c375a1adaed1263ae1eae1384865/packages/react/src/evm/providers/thirdweb-provider.tsx#L274) passed into Wagmi uses the default RPC URL.

#### What's this PR do?

- [x] feat: added `uniqueBy` util function that remove duplicated element in an array
- [x] feat: merge `supportedChain` and `activeChain` by removing duplicated chain (comparing `chainId`)


packages/react/src/evm/providers/thirdweb-provider.tsx
```diff
@@ -253,7 +254,9 @@ export const ThirdwebProvider = <
     ) {
       return supportedChains as Readonly<Chain[]>;
     }
-    return [...supportedChains, activeChain] as Readonly<Chain[]>;
+    return uniqueBy([activeChain, ...supportedChains], "chainId") as Readonly<
+      Chain[]
+    >;
```
iketw, line 259: Thanks or contributing to our repo! This is indeed an issue. What do you think if we do the following to filter out the `activeChain` from the `supportedChains` array? This way we don't need to add the `uniqueBy` function and can use a tested array function.

```
return [
      ...supportedChains.filter((c) => c.chainId !== activeChain.chainId),
      activeChain,
    ] as Readonly<Chain[]>;
```

ThianHooi: @iketw Sure, makes sense to use simpler implementation. Will make the change

iketw (approved):

PR #712, opened 2023-03-11
[React] Fix accepted types for some hooks

atharvadeosthale:
Hooks `useUnclaimedNFTs()`, `useClaimedNFTs()`, `useUnclaimedNFTSupply()`, and `useClaimedNFTSupply()` don't work with edition drop contracts, however, the comments on these hooks specified they could.

### DIFF
--- a/packages/react-core/src/evm/hooks/async/drop.ts
+++ b/packages/react-core/src/evm/hooks/async/drop.ts
@@ -21,6 +21,8 @@ import { useMutation, useQueryClient } from "@tanstack/react-query";
 import {
   NFTMetadataInput,
   QueryAllParams,
+  SignatureDrop,
+  SmartContract,
   UploadProgressEvent,
 } from "@thirdweb-dev/sdk";
 import { NFTDrop } from "@thirdweb-dev/sdk/dist/declarations/src/evm/contracts/prebuilt-implementations/nft-drop";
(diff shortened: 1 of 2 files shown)


nachoiacovino (changes requested):
we can't accept only NFTDrop which is the prebuilt, we need to accept DropContract (that has all drops) but exclude all erc1155 and erc20, so find how we do it elsewhere, but need to do a DropContract exclude EditionDrop, TokenDrop

atharvadeosthale:
`useClaimedNFTs` and `useUnclaimedNFTs` took only `NFTDrop` even before so I changed their link to point to `NFTDrop`. For the ones where we exclude edition drop and token drop I've added normal text and removed links

packages/react-core/src/evm/hooks/async/drop.ts
```diff
@@ -74,10 +76,10 @@ export function useUnclaimedNFTs(
  * const { data: claimedNFTs, isLoading, error } = useClaimedNFTs(contract, { start: 0, count: 100 });
  * ```
  *
- * @param contract - an instance of a {@link DropContract}
+ * @param contract - an instance of a {@link NFTDrop}
```
nachoiacovino, line 79: this is not correct now

nachoiacovino (changes requested):
fix the params plz :D

nachoiacovino (approved):
lgtm!

PR #713, opened 2023-03-13
[SDK] Display plugins on router release page

kumaryash90:


packages/sdk/src/evm/core/classes/contract-publisher.ts
```diff
@@ -357,11 +358,33 @@ export class ContractPublisher extends RPCConnectionHandler {
       invariant(signer, "A signer is required");
       const publisher = await signer.getAddress();
 
-      const predeployMetadata = await fetchRawPredeployMetadata(
+      const predeployMetadata = await fetchPreDeployMetadata(
```
joaquim-verges, line 361: be careful with this, i remember having a good reason for keeping the `raw` metadata around here, lets keep it as it was

packages/sdk/test/evm/before-setup.ts
```diff
@@ -268,6 +269,13 @@ export const mochaHooks = {
         gasSettings: {
           maxPriceInGwei: 10000,
         },
+        supportedChains: [
+          {
+            chainId: 31337,
+            rpc: ["http://localhost:8545"],
+            nativeCurrency: { name: "ETH", symbol: "ETH", decimals: 18 },
+          },
+        ],
```
joaquim-verges, line 278: why do you need this?

packages/sdk/test/evm/publisher.test.ts
```diff
@@ -524,4 +530,110 @@ describe("Publishing", async () => {
...
+    );
+    const compositeAbi = fullMetadata.compositeAbi;
+    expect(
+      compositeAbi != undefined &&
+        compositeAbi.length > MarketplaceV3__factory.abi.length,
+    );
+    // console.log(fullMetadata.compositeAbi);
+  });
+
+  // it("Composite Abi for Plugin Router", async () => {
```
joaquim-verges, line 569: cleanup

joaquim-verges:
@kumaryash90 can you take this to the finish line when you can?

packages/sdk/test/evm/before-setup.ts
kumaryash90: we added this for `getChainProvider` to work for the tests

packages/sdk/src/evm/core/classes/contract-publisher.ts
```diff
@@ -370,10 +372,14 @@ export class ContractPublisher extends RPCConnectionHandler {
 
         for (const [network, implementation] of implementationsAddresses) {
           if (implementation !== "") {
+            const compilerMetadata = await fetchContractMetadata(
+              predeployMetadata.metadataUri,
+              this.storage,
```
joaquim-verges, line 377: dont do this inside the for loop, its gonna redownload the same thing over and over. just do it under the raw fetch

kumaryash90: the loop will break as soon as it finds an implementation address.. so this fetch call will only run once

joaquim-verges: talked offline, the loop breaks at first hit so its ok, but should prob still be inside the try/catch

PR #719, opened 2023-03-13
[Chains] Override goerli default faucet

nachoiacovino:


jnsdls (commented):
run `build` for me please so the chain file actually gets updated :)

joaquim-verges (approved):

PR #721, opened 2023-03-14
[SDK, CLI] Add generated ABIs and types

adam-maj:


packages/cli/constants/constants.ts
```diff
@@ -9,6 +9,11 @@ export const CREATE_MESSAGES = {
   extensions: "What extensions do you want to add to your contract?",
 } as const;
 
+export const GENERATE_MESSAGES = {
+  deployerAddress: "What's the deployer address of your smart contracts?",
+  chain: "Which chain is your smart contract deployed to?",
```
joaquim-verges, line 14: where does this show?

packages/cli/src/cli/index.ts
```diff
@@ -405,6 +406,20 @@ const main = async () => {
       await detectExtensions(options);
     });
 
+  program
+    .command("generate")
+    .description(
+      "Generate types for your smart contract to strongly type the thirdweb SDK",
```
joaquim-verges, line 412: Could add "Preload ABIs and generate types for..."

packages/cli/constants/constants.ts
```diff
@@ -9,6 +9,11 @@ export const CREATE_MESSAGES = {
   extensions: "What extensions do you want to add to your contract?",
 } as const;
 
+export const GENERATE_MESSAGES = {
+  deployerAddress: "What's the deployer address of your smart contracts?",
```
joaquim-verges, line 13: would say something simpler like: 

`Deployer wallet address: ...`

maybe we need to explain that this will pull your dashboard contracts somewhere

packages/cli/src/generate/command.ts
```diff
@@ -0,0 +1,120 @@
...
+    return;
+  }
+
+  const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
+  if (packageJson.scripts?.postinstall?.includes("thirdweb generate")) {
+    return;
+  }
+
+  const postinstall = packageJson.scripts?.postinstall
+    ? packageJson.scripts.postinstall + ` && npx thirdweb generate`
```
joaquim-verges, line 103: wonder if we should skip the auto-update for post install scripts. anything that prompts the user will break in CI

adam-maj: Definitely, good point

jnsdls: this isn't true. the packages we use to prompt (at least one of the 3) handles CI and defaults to ur default value

packages/chains/chains/314.ts
```diff
@@ -10,7 +10,7 @@ export default {
   "rpc": [
     "https://filecoin.rpc.thirdweb.com/${THIRDWEB_API_KEY}",
     "https://api.node.glif.io/",
-    "https://rpc.ankr.com/filecoin"
+    "https://rpc.ankr.com/filecoin, https://filecoin-mainnet.chainstacklabs.com/rpc/v1"
```
jnsdls, line 13: this seems f'd but it's coming from source I guess?

jnsdls: submitted a PR to fix this @ source: https://github.com/ethereum-lists/chains/pull/2431

adam-maj: Lol good catch

PR #724, opened 2023-03-14
Connect wallet v2 various improvements

MananTank:
## react-core
- Stop forcing network switch when connecting wallet
- add autoSwitch option to force network switch in ThirdwebProvider
- remove deprecated and unused `chainRpc` prop from `ThirdwebProvider` 
- remove unused and confusing state 'activeChain' in wallet-provider
- Re export types from /sdk in /react-core

## wallets
- Fix wrong RPC urls and explorer URLs in wallets package
- make switchChain public in wallet connect v1 and remove walletName check

## react
- Add "Confirm in Wallet" UI in Web3Button, Network Switcher

## Other
- Fix hotlink DX

### DIFF
--- a/packages/react-core/src/core/providers/thirdweb-provider.tsx
+++ b/packages/react-core/src/core/providers/thirdweb-provider.tsx
@@ -5,7 +5,6 @@ import {
 } from "../../evm/providers/thirdweb-sdk-provider";
 import { DAppMetaData } from "../types/dAppMeta";
 import { SupportedWallet } from "../types/wallet";
-import { showDeprecationWarning } from "../utils";
 import { ThirdwebThemeContext } from "./theme-context";
 import {
   ThirdwebWalletProvider,
@@ -93,16 +92,14 @@ export interface ThirdwebProviderCoreProps<
   alchemyApiKey?: string;
   infuraApiKey?: string;
 
-  /**
-   * A partial map of chainIds to rpc urls to use for certain chains
-   * If not provided, will default to the rpcUrls of the chain objects for the supported chains
-   * @deprecated - use `chains` instead
-   */
-  chainRpc?: Record<number, string>;
-
   theme?: "light" | "dark";
 
   createWalletStorage: CreateAsyncStorage;
+
+  /**
+   * Whether or not to automatically switch to wallet's network to active chain
+   */
+  autoSwitch?: boolean;
 }
 
 // SDK handles this under the hood for us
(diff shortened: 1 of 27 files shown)


packages/react-core/src/core/providers/thirdweb-provider.tsx
```diff
@@ -117,33 +114,22 @@ export const ThirdwebProviderCore = <
...
     if (typeof props.activeChain === "string") {
-      return (
-        supportedChains.find((chain) => chain.slug === props.activeChain) ||
-        supportedChains[0]
-      );
+      return supportedChains.find((chain) => chain.slug === props.activeChain);
     }
 
-    return props.activeChain || supportedChains[0] || defaultChains[0];
+    return props.activeChain;
```
iketw, line 132: With this change we no longer have an `activeChain` in our SDK by default. Not sure about the downstream consequences here. cc: @jnsdls

packages/react-core/src/index.ts
```diff
@@ -1,2 +1,12 @@
...
+
+// export types from thirdweb-dev/sdk
+export type {
+  NFTMetadataInput,
+  NFT,
+  NFTMetadata,
+  NFTMetadataOrUri,
+} from "@thirdweb-dev/sdk";
+export type { CurrencyValue, TokenMetadata } from "@thirdweb-dev/sdk";
+export type { QueryAllParams } from "@thirdweb-dev/sdk";
```
iketw, line 12: There are other types we may be missing here and that users may need. We may need to re-export more?

Some of these types: `NFTDrop`, `Token`, `SmartContract`. In SDK we do `export * from "./types";`, so there are a bunch of types also there

packages/react-native/src/evm/wallets/connectors/coinbase-wallet/index.ts
```diff
@@ -226,7 +227,7 @@ export class CoinbaseWalletConnector extends Connector<
                 chainId: id,
                 chainName: chain.name,
                 nativeCurrency: chain.nativeCurrency,
-                rpcUrls: [chain.rpc[0] ?? ""],
+                rpcUrls: [getChainRPC(chain)],
```
iketw, line 230: Nice catch!

packages/wallets/src/evm/connectors/wallet-connect-v1/index.ts
```diff
@@ -251,32 +247,29 @@ export class WalletConnectV1Connector extends Connector<
           `Chain ${chainId} is not added in the list of supported chains`,
         );
       }
+      console.log({ chain });
```
iketw, line 250: Can remove this log

packages/react-core/src/core/providers/thirdweb-provider.tsx
MananTank: `WrappedThirdwebSDKProvider` handles the case where activeChain is `undefined` 

image

PR #731, opened 2023-03-15
react: use wallet.meta for wallet name and icon

MananTank:


packages/react/src/design-system/index.ts
```diff
@@ -115,11 +115,11 @@ export const radius = {
...
-  sm: "16px",
-  md: "24px",
-  lg: "32px",
-  xl: "48px",
+  xs: "12",
+  sm: "16",
+  md: "24",
+  lg: "32",
+  xl: "48",
 };
```
jnsdls, line 123: should be `as const` I believe to get the stricter typing

packages/react/src/wallet/ConnectWallet/screens/GetStartedScreen.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @next/next/no-img-element */
```
jnsdls, line 1: we should not have this rule in the `react` package in the first place, this is a `next` specific rule why is it enabled in our react package?

packages/wallets/src/evm/wallets/coinbase-wallet.ts
```diff
@@ -15,6 +15,11 @@ export class CoinbaseWallet extends AbstractBrowserWallet<{
 }> {
   connector?: TWConnector;
   coinbaseConnector?: CoinbaseWalletConnector;
+  static meta = {
+    iconURL:
+      "https://gateway.ipfscdn.io/ipfs/QmcJBHopbwfJcLqJpX2xEufSS84aLbF7bHavYhaXUcrLaH/coinbase.svg",
```
jnsdls, line 20: please don't hard-code the full url, instead:
```suggestion
      "ipfs:///QmcJBHopbwfJcLqJpX2xEufSS84aLbF7bHavYhaXUcrLaH/coinbase.svg",
```

packages/wallets/src/evm/wallets/device-wallet.ts
```diff
@@ -29,6 +29,12 @@ export class DeviceBrowserWallet extends AbstractBrowserWallet<
   static id = "deviceWallet" as const;
   options: WalletOptions<DeviceWalletOptions>;
 
+  static meta = {
+    name: "Device Wallet",
+    iconURL:
+      "https://gateway.ipfscdn.io/ipfs/QmcNddbYBuQKiBFnPcxYegjrX6S6z9K1vBNzbBBUJMn2ox/device-wallet.svg",
```
jnsdls, line 35:
```suggestion
      "ipfs://QmcNddbYBuQKiBFnPcxYegjrX6S6z9K1vBNzbBBUJMn2ox/device-wallet.svg",
```

packages/wallets/src/evm/wallets/metamask.ts
```diff
@@ -20,6 +20,12 @@ export class MetaMask extends AbstractBrowserWallet {
   isInjected?: boolean;
   walletConnectConnector?: WalletConnectV1ConnectorType;
 
+  static meta = {
+    name: "Metamask",
+    iconURL:
+      "https://gateway.ipfscdn.io/ipfs/QmZZHcw7zcXursywnLDAyY6Hfxzqop5GKgwoq8NB9jjrkN/metamask.svg",
```
jnsdls, line 26:
```suggestion
      "ipfs://QmZZHcw7zcXursywnLDAyY6Hfxzqop5GKgwoq8NB9jjrkN/metamask.svg",
```

packages/wallets/src/evm/wallets/wallet-connect-v1.ts
```diff
@@ -16,6 +16,12 @@ export class WalletConnectV1 extends AbstractBrowserWallet<WalletConnectV1Option
 
   static id = "walletConnectV1" as const;
 
+  static meta = {
+    name: "Wallet Connect",
+    iconURL:
+      "https://gateway.ipfscdn.io/ipfs/QmX58KPRaTC9JYZ7KriuBzeoEaV2P9eZcA3qbFnTHZazKw/wallet-connect.svg",
```
jnsdls, line 22:
```suggestion
      "ipfs://QmX58KPRaTC9JYZ7KriuBzeoEaV2P9eZcA3qbFnTHZazKw/wallet-connect.svg",
```

packages/wallets/src/evm/wallets/wallet-connect.ts
```diff
@@ -17,6 +17,12 @@ export class WalletConnect extends AbstractBrowserWallet<WalletConnectOptions> {
 
   static id = "walletConnect" as const;
 
+  static meta = {
+    name: "Wallet Connect",
+    iconURL:
+      "https://gateway.ipfscdn.io/ipfs/QmX58KPRaTC9JYZ7KriuBzeoEaV2P9eZcA3qbFnTHZazKw/wallet-connect.svg",
```
jnsdls, line 23:
```suggestion
      "ipfs://QmX58KPRaTC9JYZ7KriuBzeoEaV2P9eZcA3qbFnTHZazKw/wallet-connect.svg",
```

packages/wallets/src/evm/wallets/coinbase-wallet.ts
jnsdls: and then where we display them we resolve it back on the configured storage (which by default will end up using this gateway but this lets users overwrite which gateway to pull it from)

jnsdls (changes requested):
1 blocker:
- don't hard code the IPFS gateway url

packages/react/src/design-system/index.ts
MananTank: works just fine without it

image

adding `as const` would make of `iconSize.xs` - `"12"` instead of `string` - which is not really required here

image

packages/react/src/wallet/ConnectWallet/screens/GetStartedScreen.tsx
MananTank: yeah, this rule is added because we are using the `thirdweb` config. I can ignore the no-img-element rule

```js
// .eslintrc.cjs
module.exports = {
  root: true,
  extends: ["thirdweb"]
  rules: {
    "@next/next/no-img-element": "off", // <--
  },
};
```

packages/react/src/design-system/index.ts
jnsdls: okay, fine to leave it then

packages/react/src/wallet/ConnectWallet/screens/GetStartedScreen.tsx
jnsdls: ^ yeah we can start with that

iketw (approved):
lgtm!

packages/react/src/wallet/ConnectWallet/screens/Metamask/MetamaskConnecting.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { ConnectingScreen } from "../ConnectingScreen";
 
 export const MetamaskConnecting: React.FC<{ onBack: () => void }> = (props) => {
   return (
     <ConnectingScreen
       onBack={props.onBack}
       walletName="Metamask"
-      WalletIcon={MetamaskIcon}
+      walletIconURL={MetamaskWallet.meta.iconURL}
       supportLink="https://support.metamask.io/hc/en-us/articles/4406430256539-User-Guide-Troubleshooting"
```
jnsdls, line 10: I wonder if these should also live on the wallet meta

packages/react/src/wallet/ConnectWallet/screens/Metamask/MetamaskGetStarted.tsx
```diff
@@ -1,10 +1,10 @@
...
 import { GetStartedScreen } from "../GetStartedScreen";
 
 export const MetamaskGetStarted: React.FC<{ onBack: () => void }> = (props) => {
   return (
     <GetStartedScreen
-      WalletIcon={MetamaskIcon}
+      walletIconURL={MetamaskWallet.meta.iconURL}
       walletName="Metamask"
       chromeExtensionLink="https://chrome.google.com/webstore/detail/metamask/nkbihfbeogaeaoehlefnkodbefgpgknn"
       googlePlayStoreLink="https://play.google.com/store/apps/details?id=io.metamask"
```
jnsdls, line 10: and these, these feel meta-y?

packages/wallets/src/evm/wallets/metamask.ts
```diff
@@ -20,6 +20,12 @@ export class MetaMask extends AbstractBrowserWallet {
   isInjected?: boolean;
   walletConnectConnector?: WalletConnectV1ConnectorType;
 
+  static meta = {
+    name: "Metamask",
+    iconURL:
+      "ipfs://QmZZHcw7zcXursywnLDAyY6Hfxzqop5GKgwoq8NB9jjrkN/metamask.svg",
```
jnsdls, line 26: I think we *may* want to look at just having these as local base64 blobs / buffers hard-coded in the longer term, that'll still work with img.src regadless though so that's probably fine for now

jnsdls (approved):
no blockers but some comments

packages/react/src/wallet/ConnectWallet/screens/Metamask/MetamaskGetStarted.tsx
MananTank: Good catch!, rainbow-kit also puts all these in meta. we can do the same

image

packages/react/src/wallet/ConnectWallet/screens/Metamask/MetamaskConnecting.tsx
MananTank: it is coming from meta 🤔

iketw: I think he's talking about the `supportLink`

PR #733, opened 2023-03-15
[React] Fix wrong return on useActiveClaimConditionForWallet

nachoiacovino:
Not sure why we were doing this cc @jnsdls but was creating some bugs:
image


jnsdls (approved):
Fix lint error then good 2 go

PR #735, opened 2023-03-16
[sdk] - add functions to watch `blocks` and `transactions`

jnsdls:


packages/sdk/src/evm/functions/watchBlock.ts
```diff
@@ -0,0 +1,111 @@
...
+export type SharedBlockParams = {
+  network: NetworkInput;
+  sdkOptions?: SDKOptions;
+};
+
+export type WatchBlockNumberParams = SharedBlockParams & {
+  onBlockNumber: (blockNumber: number) => void;
+};
+
+export function watchBlockNumber(params: WatchBlockNumberParams) {
```
joaquim-verges, line 16: some docs would be nice on these

packages/sdk/src/evm/lib/static-batch-rpc.ts
```diff
@@ -53,6 +53,8 @@ export class StaticJsonRpcBatchProvider extends providers.StaticJsonRpcProvider
       provider: this,
     });
 
+    console.log("*** sendCurrentBatch() request_:", request_);
```
joaquim-verges, line 56: cleanup

packages/sdk/src/evm/functions/watchBlock.ts
jnsdls: coming right up ser

PR #736, opened 2023-03-16
[React] Move default api key to the React package

iketw:


packages/react-core/src/evm/providers/thirdweb-sdk-provider.tsx
```diff
@@ -218,7 +217,7 @@ export const ThirdwebSDKProvider = <
   // @ts-expect-error - different subtype of Chain[] but this works fine
   supportedChains = defaultChains,
   activeChain,
-  thirdwebApiKey = DEFAULT_API_KEY,
```
jnsdls, line 221: react-core also has to have (its own?) api key since it can be used independently

jnsdls (changes requested):
moving it is fine but react-core still has to have its own if it's used directly

jnsdls (approved):
fixed, merge

iketw:
> correct API keys for react-core, react-native and react

Thanks for adding it yourself! :)

PR #737, opened 2023-03-16
[SDK] Add multiplyGas function and bump polygon default gas settings

adam-maj:


packages/sdk/src/evm/core/classes/transactions.ts
```diff
@@ -223,6 +223,19 @@ export class Transaction<TResult = TransactionResult> {
     return this.contract.interface.encodeFunctionData(this.method, this.args);
   }
 
+  /**
+   * Multiply the gas limit by a factor
```
joaquim-verges, line 227: should add a quick example here.

ex: 1.2 for 20% more gas.

joaquim-verges (approved):

packages/sdk/src/evm/core/classes/transactions.ts
joaquim-verges: wonder if we should have one for gasPrice too, bit trickier tho

joaquim-verges:
add a changeset

packages/sdk/src/evm/core/classes/transactions.ts
```diff
@@ -223,6 +223,25 @@ export class Transaction<TResult = TransactionResult> {
...
+   *
+   * @example
+   * ```js
+   * // Multiple by a factor of 1.2 (increase by 20%)
+   * await tx.multiplyGasLimit(1.2)
+   * ```
+   */
+  async multiplyGasLimit(factor: number) {
```
jnsdls, line 235: it feels weird to make this async - we should set the % on the txn class and then do the operation with it applied (when we already estimate gas), otherwise I do this -> wait 5min -> then send later and it'll be off?

jnsdls (commented):
I don't think the strategy is right here

packages/sdk/src/evm/core/classes/transactions.ts
adam-maj: True - changed it to synchronous

PR #754, opened 2023-03-16
fix wallet connect v2 error 

MananTank:
Need to lock to a minor version for @walletconnect/ethereum-provider (~2.4.6) because 2.5.1 is not working for us

we can figure out how to make it work with the newer version later


jnsdls:
> /release-rp

ser

/release-pr

PR #755, opened 2023-03-16
[tooling] - switch to pnpm for faster CI

jnsdls:


jnsdls:
@joaquim-verges here's the pros and cons of this change:

**pros**
- cuts CI time by ~50%
- enforces proper dependencies are installed with each package (if not it will tell u ur missing one) (check the package.json changes in the packages to see what it caught immediately)

**cons**
- have to learn to type `pnpm` instead of `yarn`

PR #756, opened 2023-03-17
[SDK] any evm deploy setup

kumaryash90:


packages/sdk/src/evm/common/infra-data.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const Forwarder = {
+  name: "Forwarder" as const,
+  contractType: "forwarder" as const,
+  txInfo: InfraData["forwarder"] as InfraTxInfo,
+};
+
+export const EOAForwarder = {
+  name: "ForwarderEOAOnly" as const,
+  contractType: "eoaForwarder" as const,
```
joaquim-verges, line 30: why do we need 'contractType' here? i think it makes things more complicated than needed here.

just have a map from contract name to InfraTxInfo

packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -523,18 +525,37 @@ export class ContractDeployer extends RPCConnectionHandler {
       );
     } else {
       // implementation does NOT exist on chain, deploy the implementation first, then deploy a proxy
-      implementationAddress = await this.deployContractFromUri(
+      const deploymentInfo = await getDeploymentInfo(
```
joaquim-verges, line 528: this does the infra deployment too right? again needs to be explicit

packages/sdk/src/evm/common/any-evm-deploy.ts
```diff
@@ -0,0 +1,247 @@
...
+    } else if (p.name.includes("trustedForwarder")) {
+      contractTypes.push(EOAForwarder.contractType);
+      return EOAForwarder.txInfo.predictedAddress;
+    } else {
+      return "";
+    }
+  });
+  const encodedArgs = utils.defaultAbiCoder.encode(
+    constructorParamTypes,
+    constructorParamValues,
```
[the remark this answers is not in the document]
kumaryash90: should work the same, with encoded args empty

joaquim-verges:
changeset pls :)

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+  chainId: number,
+): KeylessDeploymentInfo {
+  const signature = ethers.utils.joinSignature(SIGNATURE);
+  const deploymentTransaction = getKeylessTxn(
+    {
+      gasPrice: 100 * 10 ** 9,
+      gasLimit: 100000,
+      nonce: 0,
+      data: CREATE2_FACTORY_BYTECODE,
+      chainId: chainId,
```
joaquim-verges, line 86: passing 0 here doesnt change the standard community factory address?

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+    deploymentInfo.deployment,
+    signer.provider,
+  );
+
+  // deploy community factory if not already deployed
+  if (!factoryExists) {
+    // send balance to the keyless signer
+    if (
+      (await signer.provider.getBalance(deploymentInfo.signer)).lt(
+        ethers.utils.parseEther("0.01"),
```
joaquim-verges, line 121: btw you can use `toWei("0.01")`

packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -819,36 +807,106 @@ export class ContractDeployer extends RPCConnectionHandler {
...
+          // any evm deployment flow -- with signer
+          // 0. Cache Infra contracts data
+          await this.computeAndCacheInfraContractsData();
+          console.log("computed and cached infra");
+          console.log(this.infraContractsInfoCache);
+
+          // 1. Deploy CREATE2 factory (if not already exists)
+          console.log("deploying create2 factory");
+          const create2Factory = await deployCreate2Factory(
+            this.getSigner() as Signer,
```
joaquim-verges, line 864: do you need `as Signer` here and below?

packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -1034,4 +1091,257 @@ export class ContractDeployer extends RPCConnectionHandler {
...
+    const chainId = (await this.getProvider().getNetwork()).chainId;
+
+    if (
+      !this.infraContractsInfoCache[chainId] ||
+      !this.infraContractsInfoCache[chainId]["create2Factory"]
+    ) {
+      const enforceEip155 = await isEIP155Enforced(this.getProvider());
+
+      const tempChainId = enforceEip155 ? chainId : 0;
+      const deploymentInfo = getCreate2FactoryDeploymentInfo(tempChainId);
```
joaquim-verges, line 1187: duplicated code here as well, you have a helper function that does this already

packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -1034,4 +1091,257 @@ export class ContractDeployer extends RPCConnectionHandler {
       return constructorParamValues[index];
     });
   }
+
+  private async computeAndCacheInfraContractsData() {
```
joaquim-verges, line 1095: add return types here and below

packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -1034,4 +1091,257 @@ export class ContractDeployer extends RPCConnectionHandler {
...
+            predictedAddress: address,
+            bytecode: metadata.compilerMetadata.bytecode,
+            encodedArgs: encodedArgs,
+          };
+        } else {
+          this.infraContractsInfoCache[chainId] = {
+            [contract.contractType]: {
+              predictedAddress: address,
+              bytecode: metadata.compilerMetadata.bytecode,
+              encodedArgs: encodedArgs,
```
joaquim-verges, line 1166: when is this cache actually gonna be hit? bytecode/encodedArgs is only useful for the very first deploy right?

it feels like we're over complicating the code quite a bit with this cache... and it doesnt seem that useful. What we want to avoid is having to recheck if the infra is there deployed right? why don't we just cache that?

like right now even with your cache, we still do the RPC call for `getCode` everytime, isnt that the most expensive thing to run?

I would first focus on getting the flow right, then we can optimize where it makes sense, this cache logic feels error prone to me

packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -1034,4 +1091,257 @@ export class ContractDeployer extends RPCConnectionHandler {
...
+    };
+  }
+
+  /**
+   * @internal
+   *
+   * Determine constructor params required by an implementation contract.
+   * Return abi-encoded params.
+   */
+  async encodeConstructorParamsForImplementation(
```
joaquim-verges, line 1257: i would put all these new functions in the `any-evm-utils.ts` file, or a new file. There's so much already in this class.

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -0,0 +1,748 @@
...
+    nonce: 0,
+  });
+
+  return {
+    ...deploymentTransaction,
+    deployment: create2FactoryAddress,
+  };
+}
+
+async function fetchURI(contractName: string): Promise<string> {
```
joaquim-verges, line 635: more explicit name for this plz - fetchPublishedContractURI

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -0,0 +1,748 @@
...
+      to: create2FactoryAddress,
+      data: initBytecodeWithSalt,
+    };
+    console.log("populated txn is: ", tx);
+
+    try {
+      await signer.estimateGas(tx);
+    } catch (e) {
+      console.log("error estimating gas while deploying prebuilt: ", e);
+      tx.gasLimit = BigNumber.from(7000000);
```
joaquim-verges, line 458: put this as optional param with default value

packages/sdk/src/evm/core/classes/contract-deployer.ts
kumaryash90: base class returns signer | undefined

joaquim-verges: you can do an invariant:
```
const signer = this.getSigner();
invariant(signer, "Signer required")
```

PR #764, opened 2023-03-17
add paper wallet, don't save device wallet connectParams

MananTank:
* Add Paper Wallet
* Don't save device wallet's connect params (because it contains password) 
* Fix Hotlink
* Improved UI if switch chain fails 
* Loading spinner when connecting to a wallet in wallet connector 
* removed network code-names in network selector
* disable the connect wallet button when it's in the loading state


joaquim-verges:
changeset + update to latest paper sdk 🙏

MananTank:
## Blocked because:

* something's wrong with how it is published 
* when installing the published version - wrong versions of react-core and wallets packages and error is thrown

jnsdls:
@MananTank I think this could've been caused by the same issue that was causing major bumps before - can you try now that I've merged the fix for that in from `main`?

MananTank:
@jnsdls still broken,

expected a dev build of react-core ( `3.11.0-dev-20230320105408-ce12084` but got a nightly build `3.11.0-nightly-20230319055511-e2ea5a71` )

image

let me try doing one more minor bump, so that we get a "never before seen" version - `3.12...`

jnsdls:
fixed

PR #766, opened 2023-03-17
[ReactCore] Making supportedChains optional again in the SDK provider

iketw:
- We need also inject the keys in the sdk provider in case users use the sdk provider on its own.
- Created a hook with the duplicated code
- Calling this code in the sdk will not overwrite the keys coming from react/rn

### DIFF
--- a/packages/react-core/src/core/providers/thirdweb-provider.tsx
+++ b/packages/react-core/src/core/providers/thirdweb-provider.tsx
@@ -117,7 +117,7 @@ export const ThirdwebProviderCore = <
     return props.supportedChains || (defaultChains as any as TChains);
   }, [props.supportedChains]);
 
-  const [supportedChainsWithKey, activeChainWithKey] =
+  const [supportedChainsWithKey, activeChainIdOrObjWithKey] =
     useUpdateChainsWithApiKeys(
       // @ts-expect-error - different subtype of Chain[] but this works fine
       supportedChainsNonNull,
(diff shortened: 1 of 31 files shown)


packages/react-core/src/core/providers/thirdweb-provider.tsx
```diff
@@ -115,44 +113,34 @@ export const ThirdwebProviderCore = <
 >(
   props: React.PropsWithChildren<ThirdwebProviderCoreProps<TChains>>,
 ) => {
-  const _supportedChains =
+  const supportedChains_ =
```
joaquim-verges, line 116: oof can we name this differently rather than supportedChains_ vs _supportedChains ? super hard to follow what's going on

packages/react-core/src/core/providers/thirdweb-provider.tsx
```diff
@@ -115,44 +113,34 @@ export const ThirdwebProviderCore = <
...
-      return supportedChains.find(
-        (chain) => chain.chainId === props.activeChain,
-      );
+    if (typeof _activeChain === "number") {
+      return supportedChains_.find((chain) => chain.chainId === _activeChain);
     }
-    if (typeof props.activeChain === "string") {
-      return supportedChains.find((chain) => chain.slug === props.activeChain);
+    if (typeof _activeChain === "string") {
+      return supportedChains_.find((chain) => chain.slug === _activeChain);
```
joaquim-verges, line 133: why is this supportedChains_ and not _supportedChains xD

joaquim-verges (approved):
lg

PR #767, opened 2023-03-17
[SDK, React, CLI] Make generate work with useContract

adam-maj:


packages/cli/src/generate/command.ts
```diff
@@ -82,7 +82,7 @@ export async function generate(options: GenerateOptions) {
   const packagePath = `${projectPath}/node_modules/@thirdweb-dev/generated-abis/dist`;
   if (!fs.existsSync(packagePath)) {
     throw new Error(
-      `Could not find files at ${projectPath}/node_modules/@thirdweb-dev/generated-abis`,
+      `Unable to cache ABIs. Please ensure that you're using @thirdweb-dev/sdk > 4.0.0`,
```
jnsdls, line 85: wait, we're shipping v4?

packages/react-core/src/evm/hooks/async/contracts.ts
```diff
@@ -19,13 +19,14 @@ import {
   useQueryClient,
   UseQueryResult,
 } from "@tanstack/react-query";
-import type {
+import {
```
joaquim-verges, line 22: we had `import type` here before

packages/cli/src/generate/command.ts
joaquim-verges: nope nope nope

adam-maj: This version will change to whatever minimum version actually has the generated-abis package as a hard dep

joaquim-verges: just put 'latest' version for now?

joaquim-verges (approved):
ship it

packages/cli/src/generate/command.ts
joaquim-verges, line 85:
```suggestion
      `Unable to cache ABIs. Please ensure that you're using the latest @thirdweb-dev/sdk package`,
```

PR #773, opened 2023-03-18
[SDK] Fix infinite recursion in ENS cache

adam-maj:
@jnsdls

### DIFF
--- a/packages/sdk/src/evm/common/ens.ts
+++ b/packages/sdk/src/evm/common/ens.ts
@@ -12,7 +12,10 @@ let provider: providers.Provider | undefined;
 
 const ENS_CACHE = new Map<string, Promise<CachedEns>>();
 
-export async function resolveEns(ens: string): Promise<string | null> {
+export async function resolveEns(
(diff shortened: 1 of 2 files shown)


packages/sdk/src/evm/common/ens.ts
```diff
@@ -40,11 +43,13 @@ export async function resolveEns(ens: string): Promise<string | null> {
   }
 
   const resolvedPromise = await ensPromise;
-  if (resolvedPromise.expirationTime > new Date()) {
+  if (resolvedPromise.expirationTime < new Date()) {
     // delete the cache if it's expired
     ENS_CACHE.delete(ens);
     // then call ourselves again to refresh the cache, but don't block on the result
-    resolveEns(ens);
+    if (depth === 0) {
```
iketw, line 50: Nice catch! Fixing the expirationTime comparison should fix this issue maybe we can remove this depth param? If we call it by mistake with depth === 1 this wouldn't update the cache.

PR #783, opened 2023-03-21
[sdk] - make AbiTypeBaseSchema default name instead of optional as well

jnsdls:


packages/sdk/src/evm/schema/sdk-options.ts
```diff
@@ -10,9 +10,9 @@ export const SDKOptionsSchema = z
   .object({
     // @ts-expect-error - zod doesn't know anything about readonly
     supportedChains: z.array(ChainInfoInputSchema).default(defaultChains),
-    thirdwebApiKey: z.string().optional().default(DEFAULT_API_KEY),
-    alchemyApiKey: z.string().optional().optional(),
```
jnsdls, line 14: image

@joaquim-verges

joaquim-verges: lowl

joaquim-verges (approved):

packages/sdk/src/evm/schema/contracts/common/index.ts
```diff
@@ -85,7 +85,7 @@ export const CommonTrustedForwarderSchema = z.object({
  * @internal
  */
 export const CommonSymbolSchema = z.object({
-  symbol: z.string().optional().default(""),
+  symbol: z.string().default(""),
```
jnsdls, line 88: default = behaves like optional but defaults if empty, so no reason to chain with optional

PR #785, opened 2023-03-21
[SDK] Change extension naming and remove enforced LazyMint

ciaranightingale:
Claim conditions & delayed reveal no longer look for lazy mint ABI

### DIFF
--- a/packages/sdk/src/evm/constants/erc1155-features.ts
+++ b/packages/sdk/src/evm/constants/erc1155-features.ts
@@ -1,7 +1,7 @@
 import IBurnableERC1155Abi from "@thirdweb-dev/contracts-js/dist/abis/IBurnableERC1155.json";
 import IClaimableERC1155 from "@thirdweb-dev/contracts-js/dist/abis/IClaimableERC1155.json";
 import DelayedRevealAbi from "@thirdweb-dev/contracts-js/dist/abis/IDelayedReveal.json";
-import IDropMultiPhase1155 from "@thirdweb-dev/contracts-js/dist/abis/IDrop1155.json";
+import IDrop1155 from "@thirdweb-dev/contracts-js/dist/abis/IDrop1155.json";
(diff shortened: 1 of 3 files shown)


nachoiacovino (changes requested):
we talked about it but just writing here to remind you.

this needs to be changed too on @twfeature mentions across js/react and eventually on dashboard once merged.

packages/sdk/src/evm/constants/erc1155-features.ts
```diff
@@ -24,40 +25,40 @@ export const FEATURE_EDITION_BURNABLE = {
 } as const;
 
 export const FEATURE_EDITION_CLAIM_CONDITIONS_V1 = {
-  name: "ERC1155ClaimConditionsV1",
+  name: "ERC1155DropSinglePhaseV1",
```
joaquim-verges, line 28: on thing i realized is that these names are used for autogenerating docs.

You need to look for ALL occurrences of this in this repo, but also in python and go :| 

You also need to look for occurences of these in the dashboard repo, because we rely on this string for contract detection.

and i'm not 100% sure what the docs repo relies on in terms of side bar, but it probably impacts it too.

packages/sdk/src/evm/constants/erc1155-features.ts
```diff
@@ -24,40 +25,40 @@ export const FEATURE_EDITION_BURNABLE = {
...
 export const FEATURE_EDITION_CLAIM_CONDITIONS_V2 = {
-  name: "ERC1155ClaimConditionsV2",
+  name: "ERC1155DropSinglePhase",
   namespace: "edition.drop.claim",
   docLinks: {
     sdk: "sdk.erc1155claimable",
     contracts: "DropSinglePhase1155",
   },
-  abis: [Erc1155Abi, ILazyMintAbi, IDropSinglePhase1155],
+  abis: [Erc1155Abi, IDropSinglePhase1155],
```
joaquim-verges, line 45: removing LazyMint Abi detection should be safe though, that can be done in another PR on its own and can easily be merged

PR #787, opened 2023-03-21
[ReactNativeSDK] Adds Local Wallet

iketw:
Implements and exports the new LocalWallet. No API changes.

- Uses `expo-document-picker` to import the wallet json files
- Uses `expo-file-system` to store and access wallet json files

Moves `useSupportedWallet` to react-core to be shared with RN as well


iketw:
./release-pr

PR #791, opened 2023-03-23
[auth, react-core, react] - clean up auth integration

jnsdls:


packages/wallets/src/evm/constants/wallets.ts
iketw, line 1: Nice cleanups. Thank you!

packages/react-core/src/evm/contexts/thirdweb-auth.tsx
```diff
@@ -36,28 +35,28 @@ const ThirdwebAuthContext = createContext<ThirdwebAuthContext | undefined>(
...
-      authUrl: value.authUrl?.replace(/\/$/, ""),
+      domain,
+      authUrl: authUrl?.replace(/\/$/, ""),
       auth: undefined,
     };
 
-    if (signer) {
-      context.auth = new ThirdwebAuth(new SignerWallet(signer), value.domain);
+    if (wallet) {
+      context.auth = new ThirdwebAuth(wallet, domain);
```
iketw, line 55: I think we need to do new EthersWallet here instead of using the `useWallet` hook

PR #793, opened 2023-03-23
Fix Autoconnect not working on prod bundle

MananTank:


iketw (approved):
LG!! Thanks for updating react native as well!

PR #795, opened 2023-03-23
[Wallets] Pass correct static id from the specific wallet

iketw:


packages/wallets/src/evm/wallets/wallet-connect.ts
```diff
@@ -15,7 +15,7 @@ export class WalletConnect extends AbstractBrowserWallet<WalletConnectOptions> {
 
   connector?: TWConnector;
 
-  static id = "walletConnect" as const;
```
jnsdls, line 18: why?

packages/wallets/src/evm/wallets/device-wallet.ts
```diff
@@ -26,9 +26,10 @@ export class DeviceBrowserWallet extends AbstractBrowserWallet<
 > {
   connector?: TWConnector;
   #walletImpl?: DeviceWalletImpl;
-  static id = "deviceWallet" as const;
```
jnsdls, line 29: why?

jnsdls (commented):
why no longer `as const`?

packages/wallets/src/evm/wallets/device-wallet.ts
iketw: We need to overwrite that property in the wallets that extend from those classes since that id is what we use to store the "lastConnectedWallet"

iketw: We were using the class name as the "lastConnectedWallet" but class names get minified in production builds which was the issue @MananTank was having

jnsdls: that technical constraint makes sense. That said this makes me think we're probably doing something wrong somewhere else.

PR #796, opened 2023-03-24
[SDK] Dont parse abis on detect features

iketw:
Removing abischema.parse from `extractFunctionsAbi` for perf improvements. It improved loading times by 26x since we call extractFunctionsFromAbi ~2k times the first time we load an ABI. This should be save to remove since we do set a default value in the schema, it's just that zod treats `.default` as an optional when it comes to the types.

Follow ups:
- Investigate how to improve extractFunctionsAbi, maybe look into parsing the ABI only once and store the parsed value during the creation of the contract?

### DIFF
--- a/packages/sdk/src/evm/common/feature-detection.ts
+++ b/packages/sdk/src/evm/common/feature-detection.ts
@@ -101,6 +101,19 @@ export async function extractConstructorParams(
 
 /**
  * @internal
+ * @param predeployMetadataUri
+ * @param storage
+ */
+export async function extractFunctions(
+  predeployMetadataUri: string,
+  storage: ThirdwebStorage,
+): Promise<AbiFunction[]> {
+  const metadata = await fetchPreDeployMetadata(predeployMetadataUri, storage);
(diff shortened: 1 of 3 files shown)


iketw:
./release-pr

packages/sdk/src/evm/common/feature-detection.ts
```diff
@@ -196,7 +195,9 @@ export function extractFunctionsFromAbi(
     const promise = out ? `: Promise<${out}>` : `: Promise<TransactionResult>`;
     const signature = `contract.call("${f.name}"${fargs})${promise}`;
     parsed.push({
+      // @ts-ignore we know AbiTypeBaseSchema.name is not going to be undefined since we're doing `.default("")`
```
joaquim-verges, line 198: i think you could avoid this if you wrap `f.inputs` into a `AbiInput[]` type, and put a default value for name in each of them.

nbd though i think this should be mostly fine

joaquim-verges (approved):
ship it.

we can def simplify the abi detection code, its def doing too much right now

PR #802, opened 2023-03-24
[ReactCore/Wallets] ReactCore manages all coordination 

iketw:
- Remove coordinatorStorage from the wallets package
- Have ReactCore manage lastWalletConnected and params


packages/react-core/src/core/providers/thirdweb-wallet-provider.tsx
```diff
@@ -123,7 +126,18 @@ export function ThirdwebWalletProvider(
...
+      const connectParams = params || walletParams;
+
+      await coordinatorStorage.setItem(
+        "lastConnectedParams",
+        JSON.stringify(connectParams),
+      );
+      await coordinatorStorage.setItem(
+        "lastConnectedWallet",
+        wallet.walletId,
+      );
```
jnsdls, line 139: since these are *always* linked I think, can we just store them in the same record?

PR #810, opened 2023-03-27
[ReactNative] Improve web3button and add more descriptive console.errors

iketw:


packages/react-native/src/evm/components/Web3Button.tsx
```diff
@@ -141,7 +141,6 @@ export const Web3Button = <TAction extends ActionFn>({
       content = "Switch Network";
     } else if (
       actionMutation.isLoading ||
-      !contract ||
```
iketw, line 144: If there's no contract the button should not be spinning, onError will be triggered instead.

PR #825, opened 2023-03-29
[Wallets, Core, RN, React] Allow wallets to be constructed without params where possible

iketw:
- [Wallets] Make optional all mandatory props and provide defaults
- [ReactCore] Pass through DappMeta, it's only used in the wallets package
- [Wallets] Remove `shouldAutoConnect` from the wallets package
- [React] Allow users to pass headlessMode for the CoinbaseWallet

After this is merged, the following should be possible:

```
  const w = new WalletConnectV1();
  w.connect();

  const w1 = new WalletConnect();
  w1.connect();

  const cb = new CoinbaseWallet()
  w1.connect();

  const safe = new SafeWallet();

  const device = new DeviceBrowserWallet();
```


packages/wallets/src/evm/wallets/base.ts
```diff
@@ -94,23 +94,10 @@ export abstract class AbstractBrowserWallet<
         address: data.account,
         chainId: data.chain?.id,
       });
-
-      if (data.chain?.id) {
-        this.walletStorage.setItem(
-          "lastConnectedChain",
-          String(data.chain?.id),
-        );
-      }
```
iketw, line 103: This was being saved but not used

packages/react/src/evm/providers/thirdweb-provider.tsx
```diff
@@ -74,7 +74,7 @@ export const ThirdwebProvider = <
           supportedWallets={
             supportedWallets || [
               metamaskWallet(),
-              coinbaseWallet(),
+              coinbaseWallet({headlessMode: true}),
```
MananTank, line 77: `headlessMode` must always be `true` when using ConnectWallet, so need to move it to the definition of `coinbaseWallet` function and omit the `headlessMode` from config type

packages/react/src/wallet/wallets/coinbaseWallet.ts
```diff
@@ -1,10 +1,10 @@
 import { CoinbaseWallet } from "@thirdweb-dev/wallets";
 import type { Wallet, WalletOptions } from "@thirdweb-dev/react-core";
 
-export const coinbaseWallet = () => {
+export const coinbaseWallet = (config?: { headlessMode: boolean }) => {
   return {
     id: CoinbaseWallet.id,
     meta: CoinbaseWallet.meta,
-    create: (options: WalletOptions) => new CoinbaseWallet({ ...options }),
+    create: (options: WalletOptions) => new CoinbaseWallet({ ...options, ...config }),
```
MananTank, line 8: hard code the `headlessMode: true` - should not be configurable

PR #834, opened 2023-03-29
[SDK] Update function signature of contract.call

adam-maj:


.changeset/loud-students-bake.md
```diff
@@ -0,0 +1,65 @@
+---
+"@thirdweb-dev/sdk": minor
+"@thirdweb-dev/react": minor
```
jnsdls, line 3: split this into 2 changesets, one for SDK one for `react-core` (since the useContractRead and useContractWrite hooks are coming from react-core (and will also affect react-native) - changesets will do that automatically if u just change react-core.

each changeset should clearly state the changes for itself not for other packages, if I'm a `sdk` user I don't care about changes to `react`, etc

packages/react-core/src/evm/hooks/async/contracts.ts
```diff
@@ -512,20 +512,22 @@ export function useContractEvents(
 export function useContractRead(
   contract: RequiredParam<ValidContractInstance>,
   functionName: RequiredParam<string>,
-  ...args: unknown[] | [...unknown[], CallOverrides]
+  args?: unknown[],
```
jnsdls, line 515: can we get the types all the way up to this as well?

jnsdls (changes requested):
smol changes

packages/react-core/src/evm/hooks/async/contracts.ts
adam-maj: I want to leave that for a separate PR b/c its much more involved (due to the overlap between ValidContractInstance types)

.changeset/chilled-apples-complain.md
```diff
@@ -0,0 +1,34 @@
+---
+"@thirdweb-dev/react-core": minor
+---
+
+Usage of the `useContractRead` and `useContractWrite` hooks has also changed:
```
jnsdls, line 5:
```suggestion
Usage of the `useContractRead` and `useContractWrite` hooks has changed:
```

jnsdls (approved):
tiny change then ship it

PR #837, opened 2023-03-30
[RectCore] fix auth logout by invalidating and storing invalidated cookie

### LINKED ISSUE #762: CORS issue for /auth/logout
Every other auth hook uses `credentials: "include"` (to account for CORS) in requests to the backend, except for logout hook.
This leads to a situation, where the server's attempt to clear cookie on logout is completely ignored by the browser.

`./packages/react-core/src/evm/hooks/auth/useUser.ts` should use the same approach to requests, just like every other hook in that folder.

lucoadam:
- Fixes #762

### DIFF
--- a/.changeset/beige-plums-hug.md
+++ b/.changeset/beige-plums-hug.md
@@ -1,0 +1,5 @@
+---
+"@thirdweb-dev/react-core": minor
+---
+
+Fixes auth logout by invalidating cookie by storing updated cookie credential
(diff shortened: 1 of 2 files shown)


.changeset/beige-plums-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/react-core": minor
```
jnsdls, line 2:
```suggestion
"@thirdweb-dev/react-core": patch
```

jnsdls (commented):
thanks for the PR @lucoadam - I think this should be a `patch` change since it's a bug fix

jnsdls:
@adam-maj can u take a look?

lucoadam:
> thx again @lucoadam!

Thank you guys for the awesome product

PR #845, opened 2023-03-30
hardhat deploy flow for alt chain

kumaryash90:


packages/cli/src/core/builder/hardhat.ts
```diff
@@ -87,19 +87,24 @@ export class HardhatBuilder extends BaseBuilder {
...
             !info.abi
           ) {
             logger.debug("Skipping", contractPath, "(no bytecode or metadata)");
             continue;
           }
 
           const bytecode = info.evm.bytecode.object;
-          const deployedBytecode = info.evm.deployedBytecode.object;
-          const { metadata, abi } = info;
+          const deployedBytecode = info.evm.deployedBytecode
```
joaquim-verges, line 97: what's the explanation here? deployedBytecode should always be there no?

kumaryash90: was trying this out for alt chains which have different structure for their compiled artifacts

kumaryash90: closing this PR for now

PR #851, opened 2023-03-31
[wallets] Fix autoconnect issues

MananTank:
## Fixes

* infinite loading spinner on connect wallet button when wallet is locked or  connection to app is closed
* network switch popup on page load when wallet is connected to different network than it was previously connected
* removed autoconnect timeout - don't need it


jnsdls (approved):
LGTM thx

PR #859, opened 2023-04-03
Fix build error in vite when connecting safe

MananTank:
### Issue
When connecting to safe, we get this error in Vite

image

Quick solution is to convert the dynamic import to static import. Since the connector itself is dynamically imported - this should have no effect on initial bundle size


jnsdls:
@MananTank does this mean we have to have safe dependencies as `dependencies` instead of `peerDependencies`

MananTank:
@jnsdls 
Right now, they are in `dependencies` but I don't think this code change forces us to add them to `dependencies` - since these imports are in the `connectors/safe/index.ts` and that is dynamically imported in wallet. So we should still be able to move them to `peerDependencies`

But I'm still not sure if it should be moved to `peerDependencies`. the benefit of doing that is quite small - "installing node modules is faster", but the downside is quite frustrating: if the developer decides to use safe wallet they will get an error and will need install it. 

I know we are doing it with aws-kms - which I think should be moved to a separate package altogether `wallets/node` and have environment/platform specific packages like:

* `wallets` - wallets for web
* `wallets/react-native` - wallets for react-native
* `wallets/node` - wallets for backend

MananTank:
@jnsdls After doing this code change, I'm getting the "global is not defined" error
but instead of adding the `global` => `globalThis` replacement, can we do it like this instead: 

```js
import react from '@vitejs/plugin-react';
import { defineConfig } from 'vite';
import { nodePolyfills } from 'vite-plugin-node-polyfills';

export default defineConfig({
	plugins: [
		react(),
		nodePolyfills({
			global: true,
		}),
	],
	define: {
		'process.env': {},
	},
});
```

MananTank:
@jnsdls After making this code change, I'm getting the "global is not defined" error. Instead of adding the `global` to `globalThis` replacement. Can we do this instead?

```js
import react from '@vitejs/plugin-react';
import { defineConfig } from 'vite';
import { nodePolyfills } from 'vite-plugin-node-polyfills';

export default defineConfig({
	plugins: [
		react(),
		nodePolyfills({
			global: true,
		}),
	],
	define: {
		'process.env': {},
	},
});
```

and add this in vite templates and mention this in docs

jnsdls:
> @jnsdls After making this code change, I'm getting the "global is not defined" error. Instead of adding the `global` to `globalThis` replacement. Can we do this instead?
> 
> ```js
> import react from '@vitejs/plugin-react';
> import { defineConfig } from 'vite';
> import { nodePolyfills } from 'vite-plugin-node-polyfills';
> 
> export default defineConfig({
> 	plugins: [
> 		react(),
> 		nodePolyfills({
> 			global: true,
> 		}),
> 	],
> 	define: {
> 		'process.env': {},
> 	},
> });
> ```
> 
> and add this in vite templates and mention this in docs

semes reasonable to me, let's just make sure we do this around / near the same time we release this

MananTank:
@jnsdls we can push the vite.config change right now in the Vite templates before we release this

jnsdls:
> 

okay let's do it. but to be clear that's not a blocker for this PR right?

MananTank:
@jnsdls Not a blocker as long as it's not released. 
Let me update all the vite examples/templates right now - just to be safe

PR #862, opened 2023-04-04
[React] Make tokenId optional one useNFTBalance

nachoiacovino:


packages/react-core/src/evm/hooks/async/nft.ts
```diff
@@ -275,7 +275,7 @@ export function useOwnedNFTs<TContract extends NFTContract>(
 export function useNFTBalance(
   contract: RequiredParam<NFTContract>,
   ownerWalletAddress: RequiredParam<WalletAddress>,
-  tokenId: RequiredParam<BigNumberish>,
+  tokenId?: RequiredParam<BigNumberish>,
```
iketw, line 278: Sorry if I'm missing some context on this but why do we want to make it optional? 

Is tokenId also optional in `nft.balanceOf(` below?

there's a `requiredParamInvariant(tokenId, "No tokenId provided");` statement below, will it fail if we don't pass a tokenId?

nachoiacovino: we want to make it optional because it's only required for erc1155, right now it can't be used with erc721 since you need to pass it

iketw: Is there a way we can fix the type here so that it is optional for `erc721` contracts only? Now people can call this with erc155 without a tokenId and it will fail for them

nachoiacovino: yeah it will fail on the invariant which is fine, we do this for all the other contracts that accept both erc721/erc1155, it's probably possible with typescript wizardry tho

adam-maj: I'm putting up a PR to fix the typing right now

adam-maj: Pushing a commit*

PR #863, opened 2023-04-04
[SDK] contract portal links

ciaranightingale:


packages/sdk/src/evm/constants/thirdweb-features.ts
```diff
@@ -151,7 +151,7 @@ export const FEATURE_PERMISSIONS = {
   namespace: "roles",
   docLinks: {
     sdk: "sdk.contractroles",
-    contracts: "Permissions",
+    contracts: "permissions",
```
jnsdls, line 154: this rename causes the test in `packages/sdk/test/evm/custom.test.ts` on line `239` to fail because it expects an uppercase `Permissions`

PR #864, opened 2023-04-04
[SDK] Relax the requirement for LazyMint

ciaranightingale:


jnsdls:
I think this change makes sense but would still want a @thirdweb-dev/product review before we merge this

adam-maj:
This makes sense to me, as long as none of the classes depending on these extensions use functions that are part of the `LazyMintAbi` extension - which I think might be the case. Would be good to double check this

jnsdls:
> This makes sense to me, as long as none of the classes depending on these extensions use functions that are part of the `LazyMintAbi` extension - which I think might be the case. Would be good to double check this

I assume this should fail build / get caught in tests?

PR #866, opened 2023-04-04
Add stable css selectors on web3button and connect wallet button

MananTank:


MananTank:
@jnsdls what do you think about the BEM style classes and `data-theme` and `data-is-loading` attributes added here ?

jnsdls:
> @jnsdls what do you think about the BEM style classes and `data-theme` and `data-is-loading` attributes added here ?

LGTM - makes sense

jnsdls (approved):
LGTM

PR #868, opened 2023-04-04
[Wallets] Rename base wallets

iketw:
- Make `InjectedWallet`'s params optional
- Remove connector storage from MetaMask wallet


jnsdls (commented):
requires a changeset.

also just checking these are all internal changes, no public API impact?

iketw:
> requires a changeset.
> 
> also just checking these are all internal changes, no public API impact?

We do export these classes, in the slim chance someone is extending from these, it will break for them

jnsdls (approved):

.changeset/curvy-taxis-float.md
```diff
@@ -0,0 +1,12 @@
...
+"@thirdweb-dev/react-native": patch
+"@thirdweb-dev/react-core": patch
+"@thirdweb-dev/wallets": patch
+---
+
+[Wallets]
+
+- Updates base wallet classes names
+- Makes InjectedWallet params optional
+- Removes connector storage constructor param from Injected and MetaMask wallets
```
jnsdls, line 12: if it's only a `wallets` change then let's only make a changeset for `wallets`

jnsdls (changes requested):
proper changesets please

.changeset/curvy-taxis-float.md
iketw: well, it's a wallets change but it affects the other packages since we export an updated base wallet class which `react-native`, `react-core` and `unity` use

jnsdls: But they depend on wallets so will get bumped because of that already

.changeset/curvy-taxis-float.md
```diff
@@ -0,0 +1,9 @@
+---
+"@thirdweb-dev/wallets": patch
+---
+
+[Wallets]
```
jnsdls, line 5: This is unnecessary, it's already scoped to the wallets package

jnsdls (changes requested):

PR #885, opened 2023-04-07
[SDK] Enable MarketplaceV3 deploys on any EVM

kumaryash90:


packages/sdk/src/evm/types/any-evm/deploy-data.ts
```diff
@@ -29,6 +37,7 @@ export type DeployedContractType =
   | "infra"
   | "implementation"
   | "create2Factory"
+  | "plugin"
```
joaquim-verges, line 40: let's use extensions rather than plugin everywhere

packages/sdk/src/evm/common/plugin.ts
```diff
@@ -123,3 +183,36 @@ export function joinABIs(abis: Abi[]): Abi {
...
+    fnInputs
+      .map((i: any) => {
+        return i.type === "tuple" ? getFunctionSignature(i.components) : i.type;
+      })
+      .join(",") +
+    ")"
+  );
+};
+
+export const generatePluginFunctions = (
```
joaquim-verges, line 199: use this in test instead of duplicating

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -713,6 +916,7 @@ async function encodeConstructorParamsForImplementation(
   provider: providers.Provider,
   storage: ThirdwebStorage,
   create2Factory: string,
+  customParams?: any[],
```
joaquim-verges, line 919: `constructorParamsValues`

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -721,51 +925,111 @@ async function encodeConstructorParamsForImplementation(
...
-          storage,
-          create2Factory,
-        );
-        return forwarderAddress;
-      } else {
-        return "";
-      }
-    }),
-  );
+  const constructorParamValues = customParams
```
joaquim-verges, line 928: add check that `constructorParamTypes.length === constructorParamValues.length`

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -721,51 +925,111 @@ async function encodeConstructorParamsForImplementation(
...
+      to: tx.to,
+      data: tx.data,
+    };
+  });
+  let transactionBatches: any[] = [];
+  let sum = 0;
+  let batch: PrecomputedTransactions[] = [];
+  transactionsWithoutNames.forEach((tx) => {
+    const gas = estimateGasForDeploy(tx.data);
+    if (sum + gas > 8_000_000) {
```
joaquim-verges, line 1017: constant + as default value

packages/sdk/src/evm/common/plugin.ts
```diff
@@ -105,6 +110,61 @@ async function getPluginABI(
...
+
+  let contractNames: string[] = [];
+
+  if (extendedMetadata.factoryDeploymentData?.implementationAddresses) {
+    const implementationsAddresses = Object.entries(
+      extendedMetadata.factoryDeploymentData.implementationAddresses,
+    );
+
+    try {
+      for (const [network, implementation] of implementationsAddresses) {
```
joaquim-verges, line 130: instead of a for loop, you can do 

`const implementation = implementationsAddresses.find((entry) => entry.implementation !== "")`

will return the first one that  matches your condition (or undefined if none are found)

packages/sdk/src/evm/common/plugin.ts
```diff
@@ -105,6 +110,61 @@ async function getPluginABI(
...
+        if (implementation !== "") {
+          const provider = getChainProvider(parseInt(network), {});
+          const contract = new ContractWrapper(
+            provider,
+            implementation,
+            getAllPluginsAbi,
+            {},
+          );
+
+          const pluginMap = await contract.call("getAllPlugins");
```
joaquim-verges, line 140: shouldn't assume that the caller of this function will always pass a router contract. this function should be doing the checks for what type of contract we're dealing with, and return [] if not a router contract

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -713,59 +860,120 @@ async function encodeConstructorParamsForImplementation(
...
+
+  gasLimit = Math.floor((gasLimit * 64) / 63);
+
+  return gasLimit;
+}
+
+export function createTransactionBatches(
+  transactions: PrecomputedDeploymentTransaction[],
+  upperGasLimit: number = 8_000_000, // TODO: document the estimation
+): any[] {
```
joaquim-verges, line 948: avoid using `any` unless absolutely necessary

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -616,6 +691,39 @@ export async function deployInfraWithSigner(
...
+        .deploy(txBatch);
+
+      return deployer;
+    }),
+  );
+
+  await Promise.all(
+    deployTxns.map((tx) => {
+      options?.notifier?.("deployed", "plugin");
+      return tx.deployed();
```
joaquim-verges, line 722: why not wait for deplyement confirmation inside the Promise.all above? 

you're also notifying before the actual confirmation here

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -494,17 +498,83 @@ export async function getDeploymentInfo(
...
+  if (isPluginRouter) {
+    const pluginNames = await getPluginNames(metadataUri, storage);
+    const pluginUris = await Promise.all(
+      pluginNames.map((name) => {
+        return fetchPublishedContractURI(name);
+      }),
+    );
+    const pluginMetadata = await Promise.all(
+      pluginUris.map((uri) => {
+        return fetchPreDeployMetadata(uri, storage);
```
joaquim-verges, line 522: this can be done in the promise all above as well. fetch both the uri + the metadata for each extension in parallel, rather than doing all uris, then all metadatas

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -494,17 +498,83 @@ export async function getDeploymentInfo(
   create2Factory?: string,
 ): Promise<DeploymentInfo> {
   const compilerMetadata = await fetchPreDeployMetadata(metadataUri, storage);
+  let finalInfraContracts: InfraContractType[] = [];
+  let pluginTransactions: PrecomputedDeploymentTransaction[] = [];
+  let customParams: any[] = [];
+
   if (!create2Factory) {
     create2Factory = await getCreate2FactoryAddress(provider);
```
joaquim-verges, line 506: create another const here, so the type is guaranteed to be `string`, or use an invariant - so you don't have to do `create2Factory as string` below which is error prone

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -494,17 +498,83 @@ export async function getDeploymentInfo(
...
+        return fetchPreDeployMetadata(uri, storage);
+      }),
+    );
+    const deploymentParams = await Promise.all(
+      pluginMetadata.map((metadata) => {
+        return constructDeploymentParamsForDeployer(
+          metadata,
+          provider,
+          storage,
+          create2Factory as string,
```
joaquim-verges, line 531: don't force cast, unless absolutely necessary - leads to errors that are really annoying to debug

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -713,59 +860,120 @@ async function encodeConstructorParamsForImplementation(
...
+            }
+            infraContracts.push(Forwarder.contractType);
+            const forwarderAddress = await computeForwarderAddress(
+              provider,
+              storage,
+              create2Factory,
+            );
+            return forwarderAddress;
+          } else {
+            return "";
```
joaquim-verges, line 917: this function needs comments + a // TODO here with a warning

i almost wonder if we should throw here, since we don't know what the constructor param value needs to be

kumaryash90: added a throw here

packages/sdk/src/evm/common/any-evm-utils.ts
joaquim-verges: `const create2factoryAddress = create2Factory ? create2Factory : await getCreate2FactoryAddress(provider);`

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -713,27 +697,48 @@ async function encodeConstructorParamsForImplementation(
...
+      return p.type;
+    }
+  });
 
-  const constructorParamTypes = constructorParams.map((p) => p.type);
   const constructorParamValues = await Promise.all(
     constructorParams.map(async (p) => {
+      if (constructorParamMap && constructorParamMap[p.name]) {
+        if (constructorParamMap[p.name].type) {
+          invariant(constructorParamMap[p.name].type === p.type);
```
joaquim-verges, line 717: add invariant error message

packages/sdk/src/evm/common/any-evm-utils.ts
joaquim-verges: `PrecomputedDeploymentTransaction[][]`

packages/sdk/src/evm/common/plugin.ts
```diff
@@ -105,6 +118,84 @@ async function getPluginABI(
   ).map((metadata) => metadata.abi);
 }
 
+export async function getMetadataForPlugins(
+  publishedMetadataUri: string,
+  storage: ThirdwebStorage,
+): Promise<PreDeployMetadataFetched[]> {
+  let pluginMetadata: PreDeployMetadataFetched[] = [];
+
+  const compilerMetadata = await fetchPreDeployMetadata(
```
joaquim-verges, line 127: lets use the cached version if possible (extract out of deployer)

packages/sdk/src/evm/common/plugin.ts
```diff
@@ -105,6 +118,84 @@ async function getPluginABI(
...
+      storage,
+    );
+
+    if (extendedMetadata.factoryDeploymentData?.implementationAddresses) {
+      const implementationsAddresses = Object.entries(
+        extendedMetadata.factoryDeploymentData.implementationAddresses,
+      );
+
+      try {
+        const entry = implementationsAddresses.find((item) => item[1] !== "");
```
joaquim-verges, line 149: `(item)` -> `([chainId, contractAddress])`

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -74,6 +76,8 @@ export const DEPLOYER_ABI = [
   },
 ];
 
+let deploymentPresets: Record<string, DeploymentPreset> = {};
```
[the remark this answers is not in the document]
kumaryash90: using try-catch for now

packages/sdk/src/evm/types/any-evm/deploy-data.ts
```diff
@@ -29,9 +28,24 @@ export type DeployedContractType =
...
   | "custom";
 
 export type DeploymentTransaction = {
-  contractType: DeployedContractType;
+  contractType: DeployedContractType | string;
```
joaquim-verges, line 35: why do you need `| string` here? can we not just add to `DeployedContractType` ?

packages/sdk/test/evm/any-evm-keyless.test.ts
```diff
@@ -67,13 +98,53 @@ describe("Any EVM Keyless Deploy", async () => {
     notificationCounter = 0;
     transactionCount = 0;
     contract = await deployTieredDrop();
-    expect(notificationCounter).to.equal(8);
-    expect(transactionCount).to.equal(4);
+    expect(notificationCounter).to.lessThanOrEqual(6);
+    expect(transactionCount).to.lessThanOrEqual(4);
```
joaquim-verges, line 102: why less than?

packages/sdk/test/evm/any-evm-keyless.test.ts
```diff
@@ -67,13 +98,53 @@ describe("Any EVM Keyless Deploy", async () => {
...
     expect(transactionCount).to.equal(1);
   });
+
+  it("deploy marketplacev3", async () => {
+    notificationCounter = 0;
+    transactionCount = 0;
+    const marketplace = await deployMarketplaceV3();
+
+    expect(notificationCounter).to.lessThanOrEqual(6);
+    expect(transactionCount).to.greaterThanOrEqual(3);
```
joaquim-verges, line 117: i see, the issue is that you split into 2 dependent tests here. 2 options:

1. put it all in one test (deploy tier drop, then marketplace, then tier drop again check the expected counts)
2. hardhat reset between each tests (in beforeEach)

joaquim-verges:
couple of errors due to the large throaway deployers on some chains:

- polygonZK: `unable to apply transaction even for the highest gas limit 30000000: not enough keccak counters to continue the execution`

- Linea: `err: tx data is too large (in bytes) (supplied gas 25010499)`

packages/sdk/src/evm/types/any-evm/deploy-data.ts
kumaryash90: String is for `preset`.
Since it is a broad grouping for implementation, infra, etc, didn't add it to the contract types.

packages/sdk/src/evm/common/any-evm-utils.ts
joaquim-verges: @kumaryash90 ping on this

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -888,7 +926,7 @@ export async function fetchAndCacheDeployMetadata(
   return data;
 }
 
-function estimateGasForDeploy(initCode: string) {
+export function estimateGasForDeploy(initCode: string) {
```
joaquim-verges, line 929: why did this didn't work for the zk chains? do we just need to tweak some of the values in there?

packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -972,17 +994,36 @@ export class ContractDeployer extends RPCConnectionHandler {
...
+        );
+
+        const transactionsforDirectDeploy = transactionsToSend
+          .filter((i) => {
+            return i.type !== "infra";
+          })
+          .map((i) => i.transaction);
+        transactionsforDirectDeploy.forEach((tx) => {
+          transactions.push({
+            contractType: "preset",
```
joaquim-verges, line 1008: what's the 'preset' contractType?

packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -972,17 +994,36 @@ export class ContractDeployer extends RPCConnectionHandler {
...
 
-        const transactionsToSend = deploymentInfo.map((i) => {
-          return i.transaction;
+        const transactionsToSend = deploymentInfo.filter(
+          (i) => i.transaction.data && i.transaction.data.length > 0,
+        );
+
+        const transactionsforDirectDeploy = transactionsToSend
+          .filter((i) => {
+            return i.type !== "infra";
```
joaquim-verges, line 1003: so forwarder,weth,clonefactory will get deployed together, everything else is deployed individually?

kumaryash90: yes

PR #887, opened 2023-04-07
[React] network selector revamp

MananTank:


packages/react-core/src/core/utils/updateChainRpcs.ts
```diff
@@ -5,8 +5,12 @@ import { getChainRPCs, ChainRPCOptions, Chain } from "@thirdweb-dev/chains";
...
  */
 export const updateChainRPCs = (chain: Chain, options: ChainRPCOptions) => {
-  return {
-    ...chain,
-    rpc: getChainRPCs(chain, options),
-  };
+  try {
+    return {
+      ...chain,
+      rpc: getChainRPCs(chain, options),
```
iketw, line 11: Do you know why is this throwing?

MananTank: @iketw throws when passing a `Chain` with empty `rpc` array

MananTank: In dashboard, we are gonna pass all 700+ chains as supportedChains and some of them have empty rpc array

PR #901, opened 2023-04-13
[wallets] Device Wallet Revamp

MananTank:
spec : https://www.notion.so/thirdweb/Device-Wallet-8560342a54c84e5cb6f00af4924b5c35

### DIFF
--- a/packages/unity-js-bridge/src/thirdweb-bridge.ts
+++ b/packages/unity-js-bridge/src/thirdweb-bridge.ts
@@ -6,7 +6,7 @@ import { ThirdwebStorage } from "@thirdweb-dev/storage";
 import { DAppMetaData } from "@thirdweb-dev/wallets";
 import type { AbstractClientWallet } from "@thirdweb-dev/wallets/evm/wallets/base";
 import { CoinbaseWallet } from "@thirdweb-dev/wallets/evm/wallets/coinbase-wallet";
-import { DeviceBrowserWallet } from "@thirdweb-dev/wallets/evm/wallets/device-wallet";
+import { DeviceWallet } from "@thirdweb-dev/wallets/evm/wallets/device-wallet";
 import { EthersWallet } from "@thirdweb-dev/wallets/evm/wallets/ethers";
 import { InjectedWallet } from "@thirdweb-dev/wallets/evm/wallets/injected";
 import { MetaMaskWallet } from "@thirdweb-dev/wallets/evm/wallets/metamask";
@@ -46,7 +46,7 @@ const WALLETS = [
   InjectedWallet,
   WalletConnect,
   CoinbaseWallet,
-  DeviceBrowserWallet,
+  DeviceWallet,
 ] as const;
 
 type PossibleWallet = (typeof WALLETS)[number]["id"];
@@ -147,7 +147,7 @@ class ThirdwebBridge implements TWBridge {
           });
           break;
         case "deviceWallet":
-          walletInstance = new DeviceBrowserWallet({
+          walletInstance = new DeviceWallet({
             dappMetadata,
           });
           break;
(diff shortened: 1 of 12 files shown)


test/e2e/cra-5/package.json
```diff
@@ -5,7 +5,11 @@
   "dependencies": {
     "@thirdweb-dev/react": "workspace:*",
     "@thirdweb-dev/sdk": "workspace:*",
+    "browserify-fs": "^1.0.0",
     "ethers": "^5.7.2",
+    "fs": "npm:browserify-fs",
```
joaquim-verges, line 10: we might need to expose `DeviceWallet.node()` in a separate export i think. safer than having node stuff in there that might break browser

packages/wallets/src/evm/wallets/device-wallet.ts
```diff
@@ -39,226 +42,416 @@ export class DeviceBrowserWallet extends AbstractClientWallet<
...
-  async getSigner(
-    provider?: ethers.providers.Provider,
-  ): Promise<ethers.Signer> {
-    if (!this.#wallet) {
-      throw new Error("Wallet not initialized");
+  /**
+   * initialize the wallet from saved data on storage
+   * @param password - password used for encrypting the wallet
+   */
+  async loadFromStorage(options: LoadFromStorageOptions) {
```
joaquim-verges, line 159: let's go with `load` here and `import` instead of your current `load`

so the 4 functions are
- save/load (local usage)
- export/import (external usage)

packages/unity-js-bridge/src/thirdweb-bridge.ts
```diff
@@ -182,8 +182,8 @@ class ThirdwebBridge implements TWBridge {
     const walletInstance = this.walletMap.get(wallet);
     if (walletInstance) {
       if (walletInstance.walletId === "deviceWallet" && password) {
-        const deviceWallet = walletInstance as DeviceBrowserWallet;
-        await deviceWallet.connect({ chainId, password });
+        const deviceWallet = walletInstance as DeviceWallet;
+        await deviceWallet.connect({ chainId });
```
joaquim-verges, line 186: how should we save/load in this flow? you're not using the password i pass in anymore

MananTank: I just made this change just to fix the build error for now, it's temporary and requires change.

We don't have automatic save anymore like we used to earlier. `save`, `load` needs to be done explicitly and password only required when encrypting

test/e2e/cra-5/package.json
MananTank: done, moved it to separate export and reverted the changes done in cra-5 package.json

New API: 

```ts
import { DeviceWalletNode } from '@thirdweb-dev/wallets/evm/wallets/device-walllet-node';

// use default file path for saving the data ("./wallet.json")
const deviceWallet = new DeviceWalletNode();

// OR 

// specify a file path for saving the data
const deviceWallet = new DeviceWalletNode({
  storageJsonFile: './xyz.json',
});
```

packages/unity-js-bridge/src/thirdweb-bridge.ts
MananTank: modified the temporary code to this: 

```ts
// device wallet needs to be generated or loaded before connecting
if (walletInstance.walletId === "deviceWallet") {
  const deviceWallet = walletInstance as DeviceWallet;

  // if password is provided, we can load and save
  if (password) {
    if (await deviceWallet.isSaved()) {
      await deviceWallet.loadFromStorage({
        strategy: "encryptedJson",
        password,
      });
    } else {
      await deviceWallet.generate();
      await deviceWallet.save({
        strategy: "encryptedJson",
        password,
      });
    }
  }

  // if no password is provided, we can only generate
  else {
    await deviceWallet.generate();
  }
}

await walletInstance.connect({ chainId });
```

packages/wallets/src/evm/wallets/device-wallet.ts
MananTank: done

joaquim-verges (approved):

PR #921, opened 2023-04-18
fix: set ethers@^5 version in peerDeps of @thirdweb/contracts-js

vanya2h:
This will keep in sync your ethers version with your main sdk package https://github.com/thirdweb-dev/js/blob/main/packages/sdk/package.json


joaquim-verges:
did you run into any issues with diff ethers versions?

PR #933, opened 2023-04-25
[SDK] Deployment of Extension Pattern contracts

kumaryash90:


packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -944,10 +995,33 @@ export async function encodeConstructorParamsForImplementation(
         }
 
         return nativeTokenWrapperAddress;
+      } else if (p.name && p.name.includes("linkTokenAddress")) {
+        const chainId = (await provider.getNetwork()).chainId;
+        let linkTokenAddress = getLinkTokenByChainId(chainId);
+
+        if (linkTokenAddress === ethers.constants.AddressZero) {
+          throw new Error(
+            "Chainlink LINK token not found on this network. Please provide as a custom param if available.",
```
joaquim-verges, line 1004: oof another tricky one for any EVM support. gonna need to restrict the chains at the publish level for packVRF

packages/sdk/src/evm/common/any-evm-utils.ts
```diff
@@ -978,6 +1052,21 @@ export async function encodeConstructorParamsForImplementation(
           deploymentPresets["Forwarder"] = deploymentInfo;
         }
 
+        return deploymentInfo.transaction.predictedAddress;
+      } else if (p.name && p.name.includes("entrypoint")) {
```
joaquim-verges, line 1056: is this for smart accounts?

packages/sdk/src/evm/common/plugin.ts
```diff
@@ -189,11 +190,80 @@ export async function getMetadataForPlugins(
           ).map((fetchedMetadata) => fetchedMetadata.compilerMetadata);
         }
       } catch {}
+    } else {
+      throw new Error(
+        "No implementation found for Router. Can't fetch plugin metadata",
```
joaquim-verges, line 195: are you sure we want to throw here? is not enough to return empty array?

packages/sdk/src/evm/common/plugin.ts
```diff
@@ -189,11 +190,80 @@ export async function getMetadataForPlugins(
...
+              extensionUris.map(async (uri) => {
+                return fetchAndCacheDeployMetadata(uri, storage);
+              }),
+            )
+          ).map((fetchedMetadata) => fetchedMetadata.compilerMetadata);
+        }
+      } catch {}
+    } else {
+      throw new Error(
+        "No implementation found for Router. Can't fetch extension metadata",
```
joaquim-verges, line 260: same here

packages/sdk/src/evm/common/any-evm-utils.ts
kumaryash90: Yes, Entrypoint seems to be the only dependency there, for constructor input.

packages/sdk/src/evm/common/plugin.ts
kumaryash90: I added it temporarily, till we change the publish flow for dynamic contracts.
Right now we need at least one implementation, to be able to resolve the default extensions.

The empty array will eventually throw when encoding constructor params, but the messaging won't be clear.

joaquim-verges:
closing for now @kumaryash90

PR #938, opened 2023-04-26
[Chains] Add icon to MinimalChain

nachoiacovino:


jnsdls (commented):
what else does this impact and why is this needed?

nachoiacovino:
> what else does this impact and why is this needed?

we only use this in dashboard for embeds, we can either do this or only pass the slug and resolve the chain on the embed, but that would mean you would not be able to use custom chains on the embed, afaik this is not used anywhere else.

PR #939, opened 2023-04-26
[SDK] Rename deployReleasedContract to deployPublishedContract

nachoiacovino:
Only function missing from the renaming, we're not using it internally anywhere so safe to rename without minor/major version.


jnsdls (changes requested):

joaquim-verges:
closing for now @nachoiacovino

PR #950, opened 2023-05-01
[Unity] fix unity wallet connect

joaquim-verges:


packages/unity-js-bridge/src/thirdweb-bridge.ts
```diff
@@ -135,10 +135,9 @@ class ThirdwebBridge implements TWBridge {
             dappMetadata,
           });
           break;
-        case "walletConnect":
-          walletInstance = new WalletConnect({
+        case "walletConnectV1":
+          walletInstance = new WalletConnectV1({
```
iketw, line 137: The reason for this is to support the most popular wallets correct? This is not fixing any bugs

joaquim-verges: yep

PR #955, opened 2023-05-02
[react] Local Wallet UI refactor, various fixes and cleanup

MananTank:
* Remove Guest Mode flag
* Remove credentialStorage flow
* Allow two modes - persist - `true` and `false`
* Fix Safe + Magic Link not working
* use walletIds object instead of using raw strings for wallet id
* Fix EOA wallets using wrong config in smartwallet or safe
* Fix opening the connect UI of wallet directly if there's a single wallet


packages/react-core/src/core/providers/thirdweb-provider.tsx
```diff
@@ -53,13 +53,6 @@ export interface ThirdwebProviderCoreProps<TChains extends Chain[]>
    */
   autoConnect?: boolean;
 
-  /**
-   * Whether or not to show a "Join as Guest" option in the wallet modal.
-   *
-   * Join as guest will connect the user to the dApp using a guess wallet users can export.
-   */
-  guestMode?: boolean;
```
iketw, line 61: Thanks for removing!

iketw (approved):
Do we need to add a changeset?

packages/wallets/src/evm/walletIds.ts
```diff
@@ -0,0 +1,10 @@
+export const walletIds = {
+  coinbase: "coinbaseWallet",
+  metamask: "metamask",
+  localWallet: "localWallet",
+  smartWallet: "SmartWallet",
+  safe: "Safe",
```
iketw, line 6: should have a consistent naming pattern here? All others start with lower case, we can rename these two to `safe` and `smartWallet`

packages/react-core/src/core/types/wallet.ts
```diff
@@ -23,11 +23,12 @@ export type WalletClass<I extends WalletInstance = WalletInstance> = {
   meta: (typeof AbstractClientWallet)["meta"];
 };
 
-export type Wallet<I extends WalletInstance = WalletInstance> = {
+export type Wallet<
+  I extends WalletInstance = WalletInstance,
+  Config extends Record<string, any> = Record<string, any>,
```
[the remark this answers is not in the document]
MananTank: done

packages/wallets/src/evm/walletIds.ts
```diff
@@ -0,0 +1,11 @@
+export const walletIds = {
```
iketw, line 1: There's a `constants` folder in this path, it's probably worth it to move this file in there

MananTank: done

PR #960, opened 2023-05-03
[REACT] Add onLogout function to ConnectWallet authOptions prop

frogeth:
Added an onLogout function to the authOptions props of the ConnectWallet.tsx component  to communicate when the user has logged out.


joaquim-verges (approved):
lg, thank you!

PR #964, opened 2023-05-04
Build deployArgs.json only on `deploy --dynamic` command

nkrishang:


packages/cli/src/common/processor.ts
```diff
@@ -118,6 +118,7 @@ export async function processProject(
     bytecodeUri: "",
   };
   if (options.dynamic) {
+    console.log("Dynamic selected!", options.dynamic);
```
joaquim-verges, line 121: remove log

joaquim-verges (approved):

PR #989, opened 2023-05-09
[React/RN] rename useThirdwebWallet to useProviderContext

iketw:


joaquim-verges (approved):
seems fine. was it mainly because of possible confusion of what the hook does?

MananTank:
Not sure about this name - this context only has data related to wallet connection - not all the data of ThirdwebProvider. How about `useWalletContext` ?

iketw:
> Not sure about this name - this context only has data related to wallet connection - not all the data of ThirdwebProvider. How about `useWalletContext` ?

PR #996, opened 2023-05-10
Various Connect Wallet changes

MananTank:
* add `theme` in `ConfiguredWallet.connectUI`'s props - to use theme aware UI for wallets
* add `useWalletConfig` hook to get the `ConfiguredWallet` object for active wallet
* add hooks `useSetConnectedWallet`, `useSetConnectionStatus`
* rename `useActiveChain` to `useChain` - keep the `useActiveChain` also with deprecated tag
* make `useSafe` hook await-able by returning the promise of connect() call
* add hook `useSmartWallet`
* allow rendering custom wallet details button via `<ConnectWallet detailsButton={} />` prop
* Rename "Export" to "Backup" in local wallet UI


packages/react-core/src/core/hooks/wallet-hooks.ts
```diff
@@ -114,3 +122,16 @@ export function useSwitchChain() {
   );
   return context.switchChain;
 }
+
+/**
+ *
+ * @returns a method to set a connected wallet instance
+ */
+export function useSetWallet() {
```
iketw, line 130: Can we rename to `useSetConnectedWallet` ?

packages/react-core/src/core/providers/thirdweb-wallet-provider.tsx
```diff
@@ -54,6 +54,9 @@ type ConnectFnArgs<I extends WalletInstance> =
         connectParams: NonNullable<WalletConnectParams<I>>,
       ];
 
+// maps wallet instance to it's wallet config
+const walletInstanceToConfig: Map<WalletInstance, ConfiguredWallet> = new Map();
```
iketw, line 58: Is this just to optimize not having to search for the config?

packages/react-core/src/evm/hooks/wallet.ts
```diff
@@ -103,24 +103,24 @@ export function useChainId(): number | undefined {
...
+ *   const chain = useChain()
  *
  *   return <div>{chain.chainId}</div>
  * }
  * ```
  * @see {@link https://portal.thirdweb.com/react/react.useActiveChain?utm_source=sdk | Documentation}
  * @public
  */
-export function useActiveChain(): Chain | undefined {
+export function useChain(): Chain | undefined {
```
iketw, line 123: Why this rename? This will return the active chain

packages/react-native/src/evm/components/ConnectWalletFlow/SmartWallet/SmartWalletFlow.tsx
```diff
@@ -37,6 +37,7 @@ export const SmartWalletFlow = ({
   const walletObj = useSupportedWallet(walletIds.smartWallet) as SmartWalletObj;
   const connect = useConnect();
   const targetChain = useWalletContext().activeChain;
+  const colorScheme = useColorScheme();
```
iketw, line 40: Should this be `const theme = useTheme()` instead of `useColorScheme`?

packages/react/src/wallet/wallets/magic/MagicConnectUI.tsx
```diff
@@ -45,7 +45,7 @@ export const MagicConnectUI = (props: MagicConnectUIProps) => {
   //     oauthProvider: oauthProvider,
   //   });
   //   setIsConnecting(false);
-  //   twContext.handleWalletConnect(magicWallet);
+  //   twContext.setWallet(magicWallet);
   //   props.onConnect();
   // };
```
iketw, line 50: Can we remove these comments?

iketw (commented):
Could you add some comments in this PR to outline all the changes? It will be a similar text to what you'll need for the changeset when you add it to the PR

packages/react-core/src/evm/hooks/wallet.ts
MananTank: it's confusing - we have a different meaning for "activeChain" in thirdweb provider. This hook returns the chain that the wallet is connected to

iketw: Discussed offline, we'll keep `useChain`. It returns the chain the wallet is connected to

packages/react-core/src/core/hooks/wallet-hooks.ts
MananTank: renamed

packages/react-core/src/core/providers/thirdweb-wallet-provider.tsx
MananTank: not because of that - but we can't search it - since the id in config.meta and wallet instance can be different

packages/react-native/src/evm/components/ConnectWalletFlow/SmartWallet/SmartWalletFlow.tsx
MananTank: theme returns the entire theme object, we only need 'light' | 'dark' here

packages/react-native/src/evm/components/ConnectWalletDetails/SmartWalletAdditionalActions.tsx
```diff
@@ -19,7 +19,7 @@ export const SmartWalletAdditionalActions = ({
   onExportPress: () => void;
 }) => {
   const personalWalletAddress = usePersonalWalletAddress();
-  const handleWalletConnect = useWalletContext().handleWalletConnect;
+  const { setConnectedWallet } = useWalletContext();
```
iketw, line 22: Can we do `const setConnectedWallet = useSetConnectedWallet()` instead?

iketw (approved):
one small request but feel free to merge if you don't have time, I can fix it later

PR #1014, opened 2023-05-15
Staking functions for the SDK

atharvadeosthale:
Working on creating staking functions for the SDK. First up, TS SDK, goal is to create something like

```js
await contract.staking20.stake(20);
```

Same can be made for ERC-721 and ERC-1155 staking.


atharvadeosthale:
ERC-20 staking API complete, let me know if you have any questions.

packages/sdk/src/evm/core/classes/staking-721.ts
```diff
@@ -0,0 +1,258 @@
...
+{
+  featureName = FEATURE_NFT_STAKE.name;
+  protected contractWrapper: ContractWrapper<T>;
+  protected storage: ThirdwebStorage;
+  protected _rewardToken: ContractWrapper<BaseERC20> | undefined;
+  protected _stakingToken: ContractWrapper<BaseERC721> | undefined;
+
+  private _chainId: number;
+  get chainId() {
+    return this._chainId;
```
joaquim-verges, line 52: same here, no need for this

packages/sdk/src/evm/core/classes/staking-721.ts
```diff
@@ -0,0 +1,258 @@
...
+  // READ FUNCTIONS
+  public async getRewardTokenBalance(): Promise<CurrencyValue> {
+    return fetchCurrencyValue(
+      await this.contractWrapper.getProvider(),
+      (await this.getRewardToken()).address,
+      await this.contractWrapper.readContract.getRewardTokenBalance(),
+    );
+  }
+
+  public async getRewardToken(): Promise<Currency & { address: string }> {
```
joaquim-verges, line 92: would make sense to put this code in a common helper file or a base class. There's a bunch of common logic between the diff staking contracts

joaquim-verges (commented):
Thank you for this @atharvadeosthale ! looks really good, just a few minor comments. I think you can finish erc1155 (with minimal duplicated code ideally) all and we can ship

packages/sdk/src/evm/core/classes/staking-721.ts
atharvadeosthale: One problem is that some functions might be similar but the underlying logic does differ in little ways. I'll do the moving to a base class once I'm finished with features. Then I'll find a way to make the classes have as less duplicate functions as possible

packages/sdk/src/evm/core/classes/staking-20.ts
```diff
@@ -0,0 +1,448 @@
...
+   * @example
+   * ```javascript
+   * // Get reward token info
+   * const rewardToken = await contract.staking20.getRewardToken();
+   * ```
+   *
+   * @returns see {@link Currency}
+   * @twfeature Staking20
+   */
+  public async getRewardToken(): Promise<Currency & { address: string }> {
```
joaquim-verges, line 239: can you create an explicit type for this return type? `CurrencyWithAddress` or something

packages/sdk/src/evm/core/classes/staking.ts
```diff
@@ -0,0 +1,57 @@
+import { BigNumber, BigNumberish, ethers } from "ethers";
+import { Amount } from "../../types";
+import { BaseERC1155, BaseERC20, BaseERC721 } from "../../types/eips";
+import { ContractWrapper } from "./contract-wrapper";
+import { AmountSchema } from "../../../core/schema/shared";
+import { AddressOrEns } from "../../schema";
+import { resolveAddress } from "../../common";
+
+export class Staking {
```
joaquim-verges, line 9: this doesnt need to be a class, i would just do helper functions instead of inheriting

joaquim-verges (commented):
@atharvadeosthale can you add unit tests for these?

joaquim-verges:
@atharvadeosthale any updates on this? would love to see some unit tests

atharvadeosthale:
hey, was caught up with some devrel tasks, I can now focus on completing this in the coming couple of days.

joaquim-verges:
@atharvadeosthale any updates here? do you need someone to take this over?

PR #1016, opened 2023-05-16
[SDK] Publish flow updates

kumaryash90:


packages/sdk/src/evm/core/classes/contract-deployer.ts
```diff
@@ -966,7 +1086,8 @@ export class ContractDeployer extends RPCConnectionHandler {
       extendedMetadata &&
       extendedMetadata.factoryDeploymentData &&
       (extendedMetadata.isDeployableViaProxy ||
-        extendedMetadata.isDeployableViaFactory) &&
+        extendedMetadata.isDeployableViaFactory ||
+        extendedMetadata.deployType === "customFactory") &&
```
joaquim-verges, line 1090: this is wrong no? should be autoFactory

packages/sdk/src/evm/types/deploy/deploy-options.ts
```diff
@@ -6,6 +6,7 @@ import { DeployedContractType } from "../any-evm/deploy-data";
 
 export type DeployOptions = {
   forceDirectDeploy?: boolean;
+  customFactory?: string;
```
joaquim-verges, line 9: what does this do?

kumaryash90: not using it - removed

PR #1035, opened 2023-05-19
add smart wallet base-contracts to create cli

ciaranightingale:


packages/cli/src/create/helpers/create-contract.ts
```diff
@@ -46,8 +46,13 @@ export async function createContractProject({
 }: ICreateContractProject) {
   // Check that the selected base contract is valid
   if (baseContract) {
-    const found = hasBaseContract(baseContract);
-
+    let found = false;
+    if (baseContract.includes(",")) {
+      const [account, factory] = baseContract.split(",");
```
joaquim-verges, line 51: why do we need to split here?

joaquim-verges: I imagine in the CLI you select "smart wallet" and it just generates 2 contracts for you right?

ciaranightingale: yeah exactly - can remove and just create an Account contract (would be cleaner) but at the time thought it would be better for users to have the factory contract for free

joaquim-verges:
@ciaranightingale closing for now - need to refresh it

PR #987, opened 2023-05-09
Add sepolia as a supported chain for WETH

mykcryptodev:
catalyst for this PR is that Goerli testnet is requiring a prohibitive amount of gas for testing.

To move to Sepolia, I am most interested in having a wrapped native currency address for testing things like NFT marketplace features. I used the [same WETH address on Sepolia used by Aave](https://staging.aave.com/reserve-overview/?underlyingAsset=0xd0df82de051244f04bff3a8bb1f62e1cd39eed92&marketName=proto_sepolia_v3).

I left other features such as biconomy and OZ gasless setups as TODOs.

PR #1036, opened 2023-05-19
update sepolia weth to an address that does not have a mint function

mykcryptodev:
an update from https://github.com/thirdweb-dev/js/pull/987

as brought to my attention by [this twitter thread](https://twitter.com/kassandraeth/status/1659586527498780676?s=12&t=FJ9Ia-2v_9ftxl0My0axTg), the Aave WETH contract has a mint function in it. WETH should not have a mint function that can be called by the contract owner. Until a defacto standard emerges, we should move to a WETH contract that mirrors mainnet and has a lot of usage. 

@joaquim-verges [was right](https://github.com/thirdweb-dev/js/pull/987#pullrequestreview-1425799427), this contract is the better one.

### DIFF
--- a/packages/sdk/src/evm/constants/currency.ts
+++ b/packages/sdk/src/evm/constants/currency.ts
@@ -37,7 +37,7 @@ export const NATIVE_TOKENS: Record<number, NativeToken> = {
     symbol: "SEP",
     decimals: 18,
     wrapped: {
-      address: "0xD0dF82dE051244f04BfF3A8bB1f62E1cD39eED92",
+      address: "0x7b79995e5f793A07Bc00c21412e50Ecae098E7f9",
       name: "Wrapped Ether",
       symbol: "WETH",
     },


mykcryptodev:
Btw, it looks like [Uniswap may be considering this WETH address](https://twitter.com/crypt0glitter/status/1659584192596914177) in their Sepolia implementation which could result in this becoming the defacto standard.

mykcryptodev:
@joaquim-verges [12m ago](https://twitter.com/crypt0glitter/status/1659621154821160965), a dev at Uniswap said that they are leaning towards another contract. I don't think it's worth updating this constantly until something more concrete emerges but I'll continue to keep my eye on what happens on Sepolia.

mykcryptodev:
@joaquim-verges FYI only, [Uniswap is officially going with a different contract ](https://twitter.com/Uniswap/status/1661362301646131200?s=20)-- not any activity yet but still keeping an eye out...

PR #1044, opened 2023-05-22
[react-core] do not re-throw auto-connect error

MananTank:


packages/react-core/src/core/providers/thirdweb-wallet-provider.tsx
```diff
@@ -346,11 +346,11 @@ export function ThirdwebWalletProvider(
         await wallet.autoConnect(walletInfo.connectParams);
         setConnectedWallet(wallet, walletInfo.connectParams, true);
       } catch (e) {
+        console.error(e);
         lastConnectedWalletStorage.removeItem(
           LAST_CONNECTED_WALLET_STORAGE_KEY,
         );
         setConnectionStatus("disconnected");
-        throw e;
```
dalmendray, line 353: Why not rethrow? Could you add some context here?

PR #1047, opened 2023-05-22
[chains] Fix tree shaking in thirdweb-dev/chains package

MananTank:
Problem: Importing a single Chain imports all the chains in bundle

### Before (bundle size: 490kB)

image

### After (0.8kB)

image


iketw:
Can we quantify how much this saves us? I'm guessing there's a big change in bundle size after this?

jnsdls (approved):
LGTM, massive!

PR #1054, opened 2023-05-23
[sdk] reduce cyclic deps

MananTank:


iketw:
@MananTank could you expand a bit more on how you found the cyclic dependencies, what you did to fix them and how fixing them improves performance?

MananTank:
@iketw This PR alone is not gonna do anything. I'll need to do a lot of refactor to remove all the cyclic deps. Once we remove all the cyclic deps - tree-shaking should work and that should result in reduced bundle size. 

To see the cyclic deps - you'll need to remove this line in .eslintrc.cjs of sdk pacakge `"import/no-cycle": "off"` -> this is disabled because SDK has tons of cyclic deps

PR #1095, opened 2023-05-31
Smart Wallet Factory class for creating accounts and fetching data about them.

nkrishang:


packages/sdk/src/evm/core/classes/smart-wallet-factory.ts
```diff
@@ -0,0 +1,187 @@
...
+   * @twfeature SmartWalletFactory
+   */
+  public async getAllWallets(): Promise<AccountEvent[]> {
+    const filter = {
+      fromBlock: 0,
+      toBlock: "latest",
+    }
+
+    const events = await this.events.getEvents("AccountCreated", filter);
+    console.log(events);
```
jnsdls, line 111: can we remove the console log?

packages/sdk/src/evm/core/classes/smart-wallet-factory.ts
```diff
@@ -0,0 +1,187 @@
...
+   * 
+   * @returns all wallets created via the smart wallet factory.
+   * 
+   * @twfeature SmartWalletFactory
+   */
+  public async getAllWallets(): Promise<AccountEvent[]> {
+    const filter = {
+      fromBlock: 0,
+      toBlock: "latest",
+    }
```
jnsdls, line 108: this very well may fail in RPCs but we can try

nkrishang: I thought we could use the ethers [`deployTransaction.blockNumber`](https://docs.ethers.org/v5/api/contract/contract/#Contract--properties) api to set the `fromBlock` but that wouldn't work due to this requirement:

```
If the Contract object is the result of a ContractFactory deployment, this is the transaction which was used to deploy the contract.
```

joaquim-verges: @adam-maj how far are we from being able to plug specific events into our RPC / dedicated endpoint pulling data from chainsaw you think?

.changeset/poor-lions-retire.md
```diff
@@ -0,0 +1,6 @@
+---
+"@thirdweb-dev/contracts-js": minor
+"@thirdweb-dev/sdk": minor
```
joaquim-verges, line 3: this should be patch version change, not minor

packages/sdk/src/evm/contracts/smart-contract.ts
```diff
@@ -292,6 +295,14 @@ export class SmartContract<
     return assertEnabled(this.detectOffers(), FEATURE_OFFERS);
   }
 
+  // TODO: documentation comments.
```
joaquim-verges, line 298: docs or remove?

PR #1100, opened 2023-05-31
[sdk] Reduce circular dependencies 

MananTank:


packages/sdk/src/evm/index.ts
```diff
@@ -9,7 +9,6 @@ export type { Role } from "./common/role";
 export * from "./schema/contracts/custom";
 export * from "./schema/contracts/common/claim-conditions";
 export * from "./schema/tokens/common/properties";
-export * from "./constants/chains";
```
MananTank, line 12: Removed because it's already exported from `export * from "./constants"` in this file

PR #1101, opened 2023-05-31
[Auth] Default invalid before state to 5 minutes ago

adam-maj:


packages/auth/src/core/auth.ts
```diff
@@ -54,7 +54,8 @@ export class ThirdwebAuth {
       nonce: parsedOptions?.nonce,
       expiration_time:
         parsedOptions?.expirationTime || new Date(Date.now() + 1000 * 60 * 5),
```
jnsdls, line 56: uhh shouldn't this be 24h instead of 5min?

adam-maj: no

PR #1134, opened 2023-06-05
Feature detection and SDK class for Smart Wallet.

nkrishang:


packages/sdk/src/evm/types/smart-wallet.ts
```diff
@@ -1,4 +1,48 @@
...
+};
+
+export type AccessRestrictions = {
+  startDate: Date;
+  expirationDate: Date;
+  nativeTokenLimitPerTransaction: BigNumber;
+  approvedCallTargets: string[];
+}
+
+export const AccessRestrictionsZod = z.object({
```
joaquim-verges, line 19: let's call this schema

packages/sdk/src/evm/types/smart-wallet.ts
```diff
@@ -1,4 +1,48 @@
...
+  startDate: StartDateSchema,
+  expirationDate: EndDateSchema,
+  nativeTokenLimitPerTransaction: AmountSchema.default(0),
+  approvedCallTargets: z.array(z.string()),
```
joaquim-verges, line 23: there's a schema for addresses too i believe

packages/sdk/src/evm/core/classes/smart-wallet.ts
```diff
@@ -0,0 +1,461 @@
...
+   * ```javascript
+   * const tx = await contract.smartWallet.updateAccess(signer, restrictions);
+   * const receipt = tx.receipt();
+   * ```
+   *
+   * @twfeature SmartWallet
+   */
+  updateAccess = buildTransactionFunction(
+    async(
+      signer: string,
```
joaquim-verges, line 398: i feel like we should call these inputs `signerAddress` to be explicit. Signer generally means the object

joaquim-verges (approved):

joaquim-verges (approved):
🚢 ! let's find some time after merging to add unit tests for this

joaquim-verges:
@nkrishang also the next step is to expose this via the SmartWallet class as well

PR #1138, opened 2023-06-06
feat: create Taro Testnet

jcstein:
This is a PR to add [Taro Testnet](https://tarotestnet.com) to the chain directory.


jcstein:
got it. thanks! so does it belong in `/overrides` or `/additional`?

jnsdls:
> got it. thanks! so does it belong in `/overrides` or `/additional`?

since this looks like you're adding a fully new chain this would go in `/additional`

jcstein:
alright, i think it should be good now @jnsdls thank you again for the help

PR #1140, opened 2023-06-06
[Chains] Update chains

kumaryash90:
- Add Zora Network to additional chains
Conflicts with another chain with Id 999 - we're overriding it, and adding Zora
**Note**: All chains in _additional chains_ will now override chains pulled from remote source

- Add public RPC for Astar network in overrides


jnsdls (approved):
LGTM

PR #1153, opened 2023-06-09
[Smart Wallet] Fix Celo support

kristofgazso:
Celo and its testnets do not support the BASEFEE opcode, which is called by the ERC-4337 EntryPoint when doing on-chain gas price checks. This can be avoided however by setting the maxPriorityFeePerGas to be the same as the maxFeePerGas


joaquim-verges (approved):
Thank you @kristofgazso !

PR #1161, opened 2023-06-12
added 1rpc endpoints

lucyxliu:
Offers developers an option to use 1RPC rpc endpoints to protect the privacy for their users.

### DIFF
--- a/.changeset/happy-kings-lie.md
+++ b/.changeset/happy-kings-lie.md
@@ -1,0 +1,5 @@
+---
+"@thirdweb-dev/chains": major
+---
+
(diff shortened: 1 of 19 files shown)


joaquim-verges:
hey @lucyxliu ! thanks for the contribution. 

Unfortunately this PR wont work as it will get overriten the next time someone builds the chains package. It's currently pulling from the community list at https://github.com/ethereum-lists/chains

We do have overrides in `data/overrides` but i feel like it might be more useful to the community to add the urls directly in the https://github.com/ethereum-lists/chains repo?

PR #1162, opened 2023-06-12
[SDK] Util for deterministic direct deployments

kumaryash90:


packages/sdk/src/evm/common/any-evm-utils/getInitBytecodeWithSalt.ts
```diff
@@ -13,8 +13,9 @@ import { getSaltHash } from "./getSaltHash";
 export function getInitBytecodeWithSalt(
   bytecode: string,
   encodedArgs: BytesLike,
+  salt?: string,
 ): string {
-  const saltHash = getSaltHash(bytecode);
+  const saltHash = salt ? getSaltHash(salt) : getSaltHash(bytecode);
```
joaquim-verges, line 18: don't process salt if its being passed in, just use it straight up so there's no surprises for the caller

joaquim-verges (approved):
looks great! lets add docs + portal page too

PR #1164, opened 2023-06-12
[REACT] Reduce circular dependencies

kienngo98:
Types moved to a shared file here: 
[Reduce circular dependencies: MediaRenderer <-> ModelViewer](https://github.com/thirdweb-dev/js/commit/aa757972f97d8f91fab467a85da91b4e6e09bf0d)


joaquim-verges:
thank you @kienngo98 ! running the tests. Will let @MananTank take a quick look but looks good to me

kienngo98:
@joaquim-verges Hi there! Sorry if it's a bit unrelated but I'm getting to know the codebase, and found this file (`.\packages\sdk\src\evm\common\metadata-resolver.ts`) that was commited by you 5 months ago. I was wondering why is the value hard-coded to "polygon"

MananTank:
@kienngo98 Thanks for the PR! 🔥 

We are actively trying to fix the tree-shaking issue - which also includes fixing circular dependencies. 

The circular dependency fixed here is only of types, If you run `dpdm` with `-T` flag, this circular dependency will not be shown. However it doesn't hurt to fix circular dependency with types too - so merging it! 

LGTM 🚀

PR #1167, opened 2023-06-13
[SDK] Fix circular deps issue between metadata-resolver <-> core/sdk

kienngo98:
# Before
This is mainly caused by the function `fetchContractMetadataFromAddress` inside `@r\packages\sdk\src\evm\common\metadata-resolver.ts` which is importing `{ ThirdwebSDK } from "../core/sdk";` for the line: `const polygonSDK = new ThirdwebSDK("polygon");`
However, `fetchContractMetadataFromAddress()` is also being used by the `class ThirdwebSDK`, causing the "circular deps" issue
image

  # Solution: 
This line in `metadata-resolver` caught my attention: `const polygonSDK = new ThirdwebSDK("polygon");`
Basically it's being used for the sole purpose of `multiChainRegistry.getContractMetadataURI()`
Then the simple solution is to make an instance of `MultichainRegistry` and use the method `getContractMetadataURI` directly; then get rid of ThirdwebSDK inside `metadata-resolver`

Old code: ~`const polygonSDK = new ThirdwebSDK("polygon");`~
New code: `const multiChainRegistry = new MultichainRegistry("polygon", undefined);`

Result (much better):
image


kienngo98:
I'll make another PR. I messed up something in Git lol

kienngo98:
Got these error when running build - not sure if related to this PR
image

joaquim-verges:
@kienngo98 thanks again for the contribution, I went ahead and simplified a bit more. With that I think we're free of circular dependencies! cc @MananTank

packages/sdk/src/evm/common/metadata-resolver.ts
```diff
@@ -60,11 +60,13 @@ export async function fetchContractMetadataFromAddress(
     );
     try {
       // try from multichain registry
-      const importedUri =
-        await polygonSDK.multiChainRegistry.getContractMetadataURI(
-          chainId,
-          address,
-        );
+      const contract = new Contract(
```
jnsdls, line 63: we're constructing this every time here now but it is immutable, I think we should construct it the first time we need it and then keep it around instead

`let contract;` above the FN

```js

if(!contract){
  contract = /*{... the implementation here}*/
}
```

jnsdls (approved):
overall LGTM :shipit: 

small nit about constructing the contract on every FN call but shippable regardless

packages/sdk/src/evm/common/metadata-resolver.ts
joaquim-verges: done

PR #1173, opened 2023-06-14
Clean up unnecessary exports in @sdk/src/evm/common & fix some final circular dependencies

kienngo98:
# Situation:
Everything inside `./packages/sdk/src/evm/common` is being exported to the final bundle (aka `export *`). This might partly contribute to the oversized final bundle.

# Solution:
Step 1: We need to remove all the methods (inside `./common/index.ts`) that are NOT being used OUTSIDE of `@thirdweb-dev/sdk`.
Step 2: Make sure that inside `packages/sdk` those methods above should be imported directly from the files containing them, and not via `./evm/common/index.ts`

# What to NOT remove from the `common` folder? 
-> Methods that are being used by other packages, for example:
- `getCachedAbiForContract()` is being used by `@thirdweb-dev/react-core`
- `fetchContractMetadataFromAddress()` is being used by `@thirdweb-dev/cli`
- `isContractDeployed()` is being used by `@thirdweb-dev/wallets`

-----
The following cir deps have been resolved as well
image

### DIFF
--- a/.changeset/sweet-carrots-wonder.md
+++ b/.changeset/sweet-carrots-wonder.md
@@ -1,0 +1,6 @@
+---
+"@thirdweb-dev/sdk": patch
+---
+
+- enable no-cycle eslint rule
+- various tree-shaking improvements
--- a/packages/sdk/.eslintrc.cjs
+++ b/packages/sdk/.eslintrc.cjs
@@ -1,9 +1,6 @@
 module.exports = {
   root: true,
   extends: ["thirdweb"],
-  rules: {
-    "import/no-cycle": "off",
-  },
   // allow all imports from within tests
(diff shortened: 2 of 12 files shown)


MananTank:
Hi @kienngo98, we currently have fixed all circular dependencies. circular dependencies of types is fine - as they will not be part of the bundle anyway. So no further efforts are required for fixing circular dependencies
 
You can run `dpdm` with `-T` to ignore the circular dependencies involving types

kienngo98:
@MananTank Understood. Thanks. This PR is mainly about cleaning up all the `import *` and `export *`. The circular deps is just a small part of it - and yeah this would be my last PR on the `circular deps` matter

PR #1188, opened 2023-06-16
[SDK] Code abstraction; import/export updates

kienngo98:
- Everything inside `./packages/sdk/src/evm/common` is being exported to the final bundle. This might partly contribute to the oversized final bundle.

- I looked for the methods/types that are being used ONLY within @sdk and commented them out. The ones being used by other packages like @auth, @react etc. remain intact.

- For the methods/types that are being used only within `./sdk`, I import them directly from the containing files, and not from `common/index.ts`.

Also I abstracted some dupped code.

Wondering how can I see the final bundle size after running `pnpm build` for every changes. Would be a nice DX.
image


jnsdls:
@MananTank could you take a look at this?

packages/sdk/src/evm/common/index.ts
```diff
@@ -1,90 +1,89 @@
...
+// export * from "./feature-detection/extractFunctionsFromAbi";
+// export * from "./feature-detection/extractEventsFromAbi";
+// export * from "./feature-detection/extractMinimalProxyImplementationAddress";
+export { resolveContractUriFromAddress } from "./feature-detection/resolveContractUriFromAddress";
+export { extractIPFSHashFromBytecode } from "./feature-detection/extractIPFSHashFromBytecode";
+// export * from "./feature-detection/fetchRawPredeployMetadata";
+// export * from "./feature-detection/fetchPreDeployMetadata";
+// export * from "./feature-detection/fetchExtendedReleaseMetadata";
+export { detectFeatures } from "./feature-detection/detectFeatures";
+// export * from "./feature-detection/getAllDetectedFeatures";
```
MananTank, line 23: `getAllDetectedExtensions` in this file is `@pubilc` but not export

MananTank:
Hey @joaquim-verges - Can you verify this PR? 
Is it okay to remove the exports of @internal functions ? 

there are also a few functions that are marked as @internal - but are still being used in different packages - should we mark it as @public then?

joaquim-verges:
Hmmm i would be careful with not exporting functions, how do we know they're not being used out there? This would break other people's code on update. 

Doing a pass on internal vs public would be good though. Also the contract roles change is good.

kienngo98:
@joaquim-verges, @MananTank  I made sure the commented-out exports are not being used anywhere else in the monorepo. If there are chances people using them, I can revert the changes on `packages/sdk/src/evm/common/index.ts`

kienngo98:
@MananTank @joaquim-verges I have reverted all the changes on the `export *` comments. Thanks for the feedback!

kienngo98:
image

Abstracted more contract roles

joaquim-verges (approved):
Looks good!

MananTank:
LGTM @kienngo98 Thanks!

PR #1189, opened 2023-06-16
[sdk] Add PURE annotations for improved tree-shaking

MananTank:


iketw:
Is there a lint rule we can add here so that when new schemas are added it forces the PURE command? Or add a comment in the Readme so that when we add new schemas people know to add it.

PR #1204, opened 2023-06-20
[React] Lazy load <ConnectModal />

kienngo98:
The component `<ConnectModal />` is being imported alongside `<ThirdwebProvider />`. However, it's not being used immediately nor visible by default. I don't see why we can't just lazy load this.
```
<WalletUIStatesProvider theme={theme}>
  <ThemeProvider theme={theme === "dark" ? darkTheme : lightTheme}>
    <ThirdwebProviderCore
      theme={theme}
      thirdwebApiKey={thirdwebApiKey}
      supportedWallets={wallets}
      {...restProps}
    >
      {children}
      <ConnectModal /> // <--------------------------------
    </ThirdwebProviderCore>
  </ThemeProvider>
</WalletUIStatesProvider>
```
Looking for your feedback ^

### DIFF
--- a/.changeset/tiny-llamas-complain.md
+++ b/.changeset/tiny-llamas-complain.md
@@ -1,0 +1,5 @@
+---
+"@thirdweb-dev/react": patch
+---
+
+(Code splitting) Lazy load <ConnectModal />
--- a/packages/react/src/evm/providers/thirdweb-provider.tsx
+++ b/packages/react/src/evm/providers/thirdweb-provider.tsx
@@ -5,10 +5,10 @@ import {
   WalletConfig,
 } from "@thirdweb-dev/react-core";
 import { WalletUIStatesProvider } from "./wallet-ui-states-provider";
-import { ConnectModal } from "../../wallet/ConnectWallet/ConnectModal";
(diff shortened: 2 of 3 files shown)


MananTank:
Hi @kienngo98 The idea is good - but the implementation is not quite right - `<ConnectModal/>` is always being rendered here - which defeats the purpose of lazy loading it. 

To properly implement lazy loading - you can move the return JSX from <ConnectModal/> to a Component in a different file and lazy load that - and conditionally render it using the `isWalletModalOpen` flag. You will also need to wrap that with a Suspense and a proper loading indicator

Let me know if you are up for it - I can also help here

MananTank:
Hi @kienngo98 I actually implemented it and it seems like there's not much benefit from doing this: It just reduces bundle size by 1KB ( from a total of 1MB ). This makes sense, because ConnectModal does not render much of it's own - it just renders the UI from wallet. 

Lazy loading ConnectModal is not really worth it here - considering that it introduces an additional loading state.

https://github.com/thirdweb-dev/js/assets/22043396/6c3789c4-5481-4005-9a52-5253d8d743bd

let me know what you think

kienngo98:
Hi @MananTank, really appreciate the feedback! I'll be working on this.

PR #1207, opened 2023-06-20
Add Commerce SDK

Marfuen:


packages/commerce-sdk/src/core/loyalty/index.ts
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Gift ERC20 loyalty tokens to a customer using their wallet address.
+ * 
+ * @description This function is used to gift loyalty tokens to a customer using their wallet address. The customer must have purchased and item from your store in order to receive the loyalty tokens. The function will check the order ID and verify that the customer has purchased an item from your store. If the customer has purchased an item from your store, the function will gift the customer the loyalty tokens.
+ * 
+ * @example
+ * ```
+ * const txHash = await rewardPoints({
+    nextApiRequest: {{next_api_request}},
```
jnsdls, line 16: we should not make this specific to next instead we should take in the specific data we need

packages/commerce-sdk/src/core/loyalty/index.ts
```diff
@@ -0,0 +1,213 @@
...
+    rewardAmount: 100,
+    tokenContractAddress: {{"contract_address"}},
+    webhookSecret: {{"webhook_secret"}},
+    chain: "goerli",
+    privateKey: {{"private_key"}},
```
jnsdls, line 21: is there something we can do here using our wallets package instead of expecting a pkey directly?

packages/commerce-sdk/src/core/loyalty/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  webhookSecret,
+  tokenContractAddress,
+  gaslessRelayerUrl,
+  chain,
+  rewardAmount,
+  privateKey,
+  shopifyAdminUrl,
+  shopifyAccessToken,
+}: LoyaltyRewardsParams) {
+  const secretKey = webhookSecret;
```
jnsdls, line 41: do we need to re-assign this?

packages/commerce-sdk/src/core/loyalty/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  const hmac = nextApiRequest.headers["x-shopify-hmac-sha256"];
+  const body = await getRawBody(nextApiRequest);
+  const hash = crypto
+    .createHmac("sha256", secretKey)
+    .update(body)
+    .digest("base64");
+
+    // Compare our hash to Shopify's hash
+    if (hash === hmac) {
+      const shopifyOrderId = nextApiRequest.headers["x-shopify-order-id"];
```
jnsdls, line 52: looks like all we need from the inboud request are the order ID + the hmac -> can we just let users pass us those? we could then still build helpers for next api routes / express / etc to grab them off of the request for the user, but I'd like to split that out

packages/commerce-sdk/src/core/loyalty/index.ts
```diff
@@ -0,0 +1,213 @@
...
+
+      let wallet = "";
+      try {
+        wallet =
+          itemsPurchased[0].customAttributes.find(
+            (e: any) => e.key === "wallet",
+          )?.value || "";
+      } catch (e) {
+        throw new Error("No wallet address found in order's custom attributes");
+      }
```
jnsdls, line 81: let's make *very* sure we document that this is how they have to attach the wallet address, also something to ask others (lazer / shopify) around to provide feedback if theres really no better way

packages/commerce-sdk/src/core/loyalty/index.ts
```diff
@@ -0,0 +1,213 @@
...
+ * Send tokens to specified wallet address.
+ * 
+ * @example
+ * ```javascript
+ * const txHash = await sendTokens({
+ *   wallet: "{{wallet_address}}",
+ *   tokenContractAddress: {{"contract_address"}},
+ *   amount: 100,
+ *   chain: "goerli",
+ *   privateKey: {{"private_key"}},
```
jnsdls, line 120: same as above RE can we do something with wallet package here?

jnsdls (commented):
good start left an initial review of things that stood out to me

PR #1210, opened 2023-06-21
update: Add zerion to useInstalledWallets

konojunya:
Add zerion to `useInstalledWallets`.
I should have included it in [this PR](https://github.com/thirdweb-dev/js/pull/1160), but I think the author forgot to include this PR.


packages/react/src/wallet/hooks/useInstalledWallets.ts
```diff
@@ -3,18 +3,20 @@ import { assertWindowEthereum } from "@thirdweb-dev/wallets";
...
   let isMetamaskInstalled = false;
   let isCoinbaseWalletInstalled = false;
+  let isZerionWalletInstalled = false;
 
   if (assertWindowEthereum(globalThis.window)) {
     isMetamaskInstalled = globalThis.window.ethereum?.isMetaMask;
     isCoinbaseWalletInstalled =
       globalThis.window.ethereum?.isCoinbaseWallet ||
-      globalThis.window.ethereum?.providers?.some((p) => p.isCoinbaseWallet) ||
-      false;
```
konojunya, line 12: Array.prototype.some always returns boolean. Therefore `|| false` is not necessary.
Since Array is an optional chaining, it may return `undefined`, but since everything uses `double not` before return, it will be a `Boolean` when used by the user.

joaquim-verges (approved):
lgtm

konojunya:
I'll check CI soon.

packages/react/src/wallet/hooks/useInstalledWallets.ts
konojunya: @joaquim-verges fixed it.
https://github.com/thirdweb-dev/js/pull/1210/commits/dd1583ea6cb0cb298f7671abdc8b34db94d41afe

jnsdls: can we do `!!` instead of `Boolean()` here? (just convention)

joaquim-verges: @konojunya any chance you could switch to `!!` instead of `Boolean` (and fix the conflicts)? thank you!

PR #1213, opened 2023-06-21
[RC/Wallets] Improve error messages for wc and auth

iketw:


packages/react-core/src/evm/contexts/ThirdwebAuthProvider.tsx
```diff
@@ -52,7 +52,7 @@ function ChangeActiveWalletOnAccountSwitch() {
       try {
         await switchAccount(data.address);
       } catch (err) {
-        console.warn(
+        console.debug(
```
iketw, line 55: No need to log a warn every time a user tries to switch network

packages/react-core/src/evm/contexts/ThirdwebAuthProvider.tsx
```diff
@@ -63,7 +63,7 @@ function ChangeActiveWalletOnAccountSwitch() {
       try {
         await logout();
       } catch (err) {
-        console.warn(`[Auth] Failed to logout with error:\n`, err);
```
iketw, line 66: No need to log a warn every time a user tries to disconnect

packages/wallets/src/lib/wagmi-core/errors.ts
```diff
@@ -29,7 +29,7 @@ export class RpcError<T = undefined> extends Error {
       throw new Error('"message" must be a nonempty string.');
     }
 
-    super(message);
+    super(`${message}. Cause: ${cause}`);
```
iketw, line 32: Showing the cause to the devs

PR #1215, opened 2023-06-21
Migration to WalletConnectV2

iketw:


iketw:
./release-pr

.changeset/sour-dingos-taste.md
```diff
@@ -0,0 +1,7 @@
+---
+"@thirdweb-dev/wallets": major
```
MananTank, line 2: Doesn't that also bump the `@thirdweb-dev/react` to major version?

hixus:
Thirdweb mobile sdk user here. Noticed this now and would like is there some beta npm packages or what is the expected  migration path if this comes at the same time as wc 1 is shutdown?

iketw:
./release-pr

iketw:
> Thirdweb mobile sdk user here. Noticed this now and would like is there some beta npm packages or what is the expected migration path if this comes at the same time as wc 1 is shutdown? We did some of the changes manually so wc2 somewhat works.

Hi @hixus , we are releasing our updated packages later today. The updates deprecate wcv1 and migrates wallets to use wcv2. For React Native specifically you shouldn't see any API changes, you should be able to use Rainbow and MM the same way, it will just use WCV2 in the background. 

If you want to test the package we'll release, you can get it here: `@thirdweb-dev/react-native@0.0.0-dev-c4fbd0b-20230627134115`

Feel free to ask any other questions, we know this migration is stressful

iketw:
./release-pr

packages/wallets/package.json
```diff
@@ -331,30 +331,27 @@
...
     "buffer": "^6.0.3",
     "cross-fetch": "^3.1.5",
     "crypto-js": "^4.1.1",
-    "eip1193-provider": "^1.0.1",
     "eth-provider": "^0.13.6",
     "eventemitter3": "^5.0.0",
     "magic-sdk": "^13.6.2",
     "web3-core": "1.5.2"
   },
+  "resolutions": {
```
jnsdls, line 349: resolutions on the package-level don't do anything with `pnpm` @iketw @MananTank - if you *need* these resolutions you will have to put them in the root package.json

jnsdls: (if you update your pnpm version to latest it will also warn you about this)

.changeset/sour-dingos-taste.md
```diff
@@ -0,0 +1,7 @@
...
+"@thirdweb-dev/react-native": patch
+"@thirdweb-dev/react": patch
+---
+
+[RN/Wallets/React] Migrate wallets to WCV2 and remove WCV1 support
```
jnsdls, line 7: can we not do this `[RN/Wallets/React]` stuff? it already puts this changeset in every package when it gets released based on the above metadata of which packages changed how

jnsdls: do it in PR titles sure but not in the changeset, unnecessary to do so

jnsdls (changes requested):

jnsdls (approved):

PR #1221, opened 2023-06-22
Omit marketplacev3 filter parameters that have no effect

mykcryptodev:
It is not possible to:
1. Filter auctions by the offeror address
2. Filter offers by the seller of the NFT

This PR makes it so that 
1. `marketplace.englishAuctions.getAll()` does not accept `offeror` as a parameter
2. `marketplace.offers.getAll()` does not accept `seller` as a parameter

Will link a corresponding PR for documentation updates shortly...


jnsdls:
@kumaryash90 could you check this out and see if it looks good to you?

joaquim-verges (approved):

PR #1232, opened 2023-06-23
[STORAGE] Use subdomain-based paths for our IPFS gateway

nessup:
This PR eliminates the use of `ipfs-{n}.thirdwebcdn.com` gateways. See https://github.com/thirdweb-dev/universal-gateway/pull/20 for a more detailed description of this change.

PR #1252, opened 2023-06-27
[STORAGE] Use origin isolated gateway

nessup:


packages/storage/src/common/urls.ts
```diff
@@ -6,9 +6,7 @@ import { GatewayUrls } from "../types";
 export const DEFAULT_GATEWAY_URLS: GatewayUrls = {
   // Note: Gateway URLs should have trailing slashes (we clean this on user input)
   "ipfs://": [
-    "https://ipfs-3.thirdwebcdn.com/ipfs/",
-    "https://ipfs-4.thirdwebcdn.com/ipfs/",
-    "https://ipfs-5.thirdwebcdn.com/ipfs/",
+    "https://ipfs-staging.thirdwebcdn.com/ipfs/",
```
joaquim-verges, line 9: how come we're using staging here feels off?

jnsdls: this is actually blocking, we cannot hard-code staging here, we can however now change this to ipfs-public?

jnsdls (changes requested):

jnsdls:
Closing in favor of #1232

PR #1253, opened 2023-06-27
[React] Stop using useActiveChain deprecated hook

rattrayd11:
Remove useActiveChain, update to useChain


nachoiacovino (approved):
good job!

PR #1255, opened 2023-06-27
[Wallets] handle passing RPC urls to SmartWallet config

joaquim-verges:


packages/wallets/src/evm/connectors/smart-wallet/index.ts
```diff
@@ -36,14 +36,14 @@ export class SmartWalletConnector extends Connector<SmartWalletConnectionArgs> {
...
   async initialize(personalWallet: EVMWallet) {
     const config = this.config;
-    const chain =
-      typeof config.chain === "string"
-        ? config.chain
-        : (config.chain as Chain).slug;
+    const originalProvider = getChainProvider(config.chain, {
+      thirdwebApiKey: config.thirdwebApiKey || DEFAULT_WALLET_API_KEY,
+    }) as providers.BaseProvider;
+    const chainId = (await originalProvider.getNetwork()).chainId;
```
jnsdls, line 42: oof so even if we have the chain upfront we always make another async request first?

PR #1257, opened 2023-06-28
v4? - functions based in viem

jnsdls:


jxom:
👀

jxom:
FYI - I think you guys will greatly benefit by using the [tree-shakable Actions](https://twitter.com/jakemoxey/status/1673846338067529730?s=20) internally. Will help a lot with keeping the bundle size slim for consuming code! For example, [`getBalance`](https://github.com/thirdweb-dev/js/blob/c0151f1783fd9049c61ea3b63a808175ec7e3764/packages/sdk-core/src/read/erc20/getBalance.ts#L45) itself only needs `readContract` from viem, so you can just import `readContract` and pass the `client` to it.

https://viem.sh/docs/clients/custom.html#tree-shaking

jnsdls:
> 👀

u're everywhere ser

jnsdls:
> 

ah this is *very* neat indeed. makes total sense

jnsdls:
> FYI - I think you guys will greatly benefit by using the [tree-shakable Actions](https://twitter.com/jakemoxey/status/1673846338067529730?s=20) internally. Will help a lot with keeping the bundle size slim for consuming code! For example, [`getBalance`](https://github.com/thirdweb-dev/js/blob/c0151f1783fd9049c61ea3b63a808175ec7e3764/packages/sdk-core/src/read/erc20/getBalance.ts#L45) itself only needs `readContract` from viem, so you can just import `readContract` and pass the `client` to it.
> 
> https://viem.sh/docs/clients/custom.html#tree-shaking

ah this is very neat indeed. makes total sense

Issue #1009, opened 2023-05-15
Missing import of FormatTypes


mahapo:
Importing of FormatTypes is not possible with

"@thirdweb-dev/sdk": "^3.10.14",
"ethers": "5.7.2",

![image](https://github.com/thirdweb-dev/js/assets/28730855/613d2ee0-e53e-4a9b-a85d-fabcd1e8d5b8)
![image](https://github.com/thirdweb-dev/js/assets/28730855/960c2e0c-e590-4445-82fe-2bdd60f0494c)
![image](https://github.com/thirdweb-dev/js/assets/28730855/e617b3d6-5ac8-4f45-a166-f2840a92d1e2)

mahapo:
Works with 
"@thirdweb-dev/auth": "3.0.15",
"@thirdweb-dev/sdk": "^3.10.14",
"@thirdweb-dev/wallets": "0.2.16",

joaquim-verges:
does it work if you explicitely add @ethersproject/units as a dependency in your package.json?

mahapo:
@joaquim-verges I tried with @ethersproject/units and with ethers. Both version 5.7.0 and still the error is still there.

Also with new version 1.0.2

joaquim-verges:
@mahapo removed the dependency in this file altogether here https://github.com/thirdweb-dev/js/pull/1277

once its merged you can pull it from `@thirdweb-dev/sdk@nigthly`

joaquim-verges:
The fix is now on latest stable. Let us know if that works for you @mahapo

Issue #1168, opened 2023-06-13
Cannot find models '...ethers/lib/utils'


backface:
I am building a Nuxt app and everything works as it should in dev mode. However, when generating a signature on the server  in the production build it fails with:

"Cannot find module '/app/.output/server/node_modules/ethers/lib/utils' imported from /app/.output/server/chunks/gen.mjs\nDid you mean to import ethers/lib/utils.js?"

What's wrong here? Could be a library/version mismatch?

backface:
> Hi @backface - Can you create a minimal repo and link it here so we can replicate this issue and fix it ASAP

Here is a nase of a Nuxt starter with signature minting which works for me in dev, but not in production build (I am using pnpm 7.17.0 and node v18.16.0)

https://github.com/backface/thirdweb-nuxt-test

MananTank:
Hi @backface - Looking at the package.json - it seems that you have not installed `ethers@^5`, which is required. Can you add it as a dependency and try again?

backface:
> Hi @backface - Looking at the package.json - it seems that you have not installed `ethers@^5`, which is required.
> 

as a dependency it gets pulled in on install - I have it in version 5.7.2 


> ```
> build: {
> transpile: ['ethers/lib/utils'],
> },
> ```

indeed that seems to work!

> Are you getting this error in dev?

No, not with pnpm, but with yarn I do get it as well

MananTank:
@backface So, does this fix your issue? Can I close this issue?

backface:
@MananTank No. With the transpile option set in now fails in 'run dev' witH

`Caught SyntaxError: The requested module '/_nuxt/node_modules/.pnpm/ethers@5.7.2/node_modules/ethers/lib/utils.js?v=5cdcadb1' does not provide an export named 'isAddress' (at null:33667:10)`

joaquim-verges:
@mahapo @backface fixed it in #1277

PR #1277, opened 2023-06-29
[SDK] remove ethers/lib/utils dependency

joaquim-verges:


MananTank:
Awesome! - This should fix https://github.com/thirdweb-dev/js/issues/1009 and https://github.com/thirdweb-dev/js/issues/1168

PR #1292, opened 2023-07-06
fix: handling authz errors from smart wallet connector

assimovt:
fixes FTD-1892

### DIFF
--- a/packages/wallets/src/evm/connectors/smart-wallet/lib/erc4337-signer.ts
+++ b/packages/wallets/src/evm/connectors/smart-wallet/lib/erc4337-signer.ts
@@ -65,34 +65,45 @@ export class ERC4337EthersSigner extends Signer {
 
   unwrapError(errorIn: any): Error {
     try {
+      let errorMsg = "Unknown Error";
+
       if (errorIn.error) {
-        const error = new Error(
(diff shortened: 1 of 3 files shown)


packages/wallets/src/evm/connectors/smart-wallet/lib/erc4337-signer.ts
```diff
@@ -65,34 +65,45 @@ export class ERC4337EthersSigner extends Signer {
...
-      }
-      if (errorIn.body && typeof errorIn.body === "object") {
+        errorMsg = `The bundler has failed to include UserOperation in a batch: ${errorIn.error}`;
+      } else if (errorIn.body && typeof errorIn.body === "string") {
         const errorBody = JSON.parse(errorIn.body);
-        let paymasterInfo: string = "";
-        let failedOpMessage: string | undefined =
-          errorBody?.error?.message || errorBody?.error?.data;
-        if (failedOpMessage?.includes("FailedOp") === true) {
+        const errorStatus = errorIn.status || "500";
```
joaquim-verges, line 74: maybe say unknown here as a default value too. don't want false positives

joaquim-verges (commented):
can you add a changeset?

PR #1295, opened 2023-07-06
Use WCv2 connector on mobile 

MananTank:


packages/wallets/src/evm/connectors/wallet-connect/index.ts
```diff
@@ -171,17 +171,36 @@ export class WalletConnectConnector extends WagmiConnector<
   }
 
   async disconnect() {
+    const cleanup = () => {
+      if (typeof window === "undefined") {
+        return;
+      }
+      for (var key in localStorage) {
+        if (key.startsWith("wc@2")) {
+          localStorage.removeItem(key);
```
iketw, line 180: This doesn't work in ReactNative, localStorage is not available.

packages/wallets/src/evm/connectors/wallet-connect/index.ts
```diff
@@ -171,17 +171,36 @@ export class WalletConnectConnector extends WagmiConnector<
   }
 
   async disconnect() {
+    const cleanup = () => {
+      if (typeof window === "undefined") {
```
iketw, line 175: There's a window object in RN, you can check for what you're using:
```
if (typeof localStorage !== "undefined") {
    .... do the work
}
```

iketw:
Thanks for adding the RN code! I added a comment about localStorage not being present in RN, other than that it worked as expected on RN.

PR #1299, opened 2023-07-06
[SDK] Create ERC721LoyaltyCard and ERC721UpdatableMetadata

nkrishang:


packages/sdk/src/evm/contracts/prebuilt-implementations/loyalty-card.ts
```diff
@@ -0,0 +1,390 @@
...
+ * ```javascript
+ * import { ThirdwebSDK } from "@thirdweb-dev/sdk";
+ *
+ * const sdk = new ThirdwebSDK("{{chainName}}");
+ * const contract = await sdk.getContract("{{contract_address}}", "nft-collection");
+ * ```
+ *
+ * @public
+ */
+export class LoyaltyCard extends StandardErc721<LoyaltyCardContract> {
```
joaquim-verges, line 44: I don't think we need a prebuilt class for this, we want to move away from specific prebuilt classes in favor of using SmartContract + extension.

The API for loyalty should just be `contract.erc721.loyalty.xyz()`

this makes it more straightforward for users (one way of doing things) and way more maintainable for us.

In a few versions of the SDK the goal is to remove all the prebuilt classes

packages/sdk/src/evm/core/classes/erc-721.ts
```diff
@@ -606,6 +609,52 @@ export class Erc721<
...
+   *
+   * @remarks Cancel loyalty card NFTs held by the connected wallet
+   *
+   * @example
+   * ```javascript
+   * // The token ID of the loyalty card you want to cancel
+   * const tokenId = 0;
+   *
+   * const result = await contract.erc721.cancel(tokenId);
+   * ```
```
joaquim-verges, line 625: add `@twfeature ERC721LoyaltyCard`

joaquim-verges (approved):
looks good ser! couple of tests would be nice but can be done later

PR #1307, opened 2023-07-08
Expose account contract in SmartWallet

nkrishang:


packages/wallets/src/evm/connectors/smart-wallet/index.ts
```diff
@@ -216,6 +237,22 @@ export class SmartWalletConnector extends Connector<SmartWalletConnectionArgs> {
...
+  /**
+   * Get the underlying account smart contract of the smart wallet.
+   * @returns the account smart contract of the smart wallet.
+   */
+  async getAccountContract(): Promise<Account<IAccountCore>> {
+    const signer = await this.getSigner();
+    const network = (await (await this.getProvider()).getNetwork()).chainId;
+    const sdk = ThirdwebSDK.fromSigner(signer, network);
+
+    if (await this.isDeployed()) {
```
joaquim-verges, line 249: This should be !isDeployed no?

packages/wallets/src/evm/connectors/smart-wallet/index.ts
```diff
@@ -216,6 +237,22 @@ export class SmartWalletConnector extends Connector<SmartWalletConnectionArgs> {
...
+      throw new Error("Account contract not deployed");
+    }
+    const contract = await sdk.getContract(await signer.getAddress());
+    return contract.account;
```
joaquim-verges, line 253: I'm actually tempted to return the whole contract here rather than contract.account. this will be super useful to call custom functions on accounts

packages/wallets/src/evm/connectors/smart-wallet/index.ts
```diff
@@ -216,6 +237,22 @@ export class SmartWalletConnector extends Connector<SmartWalletConnectionArgs> {
...
   }
 
+  /**
+   * Get the underlying account smart contract of the smart wallet.
+   * @returns the account smart contract of the smart wallet.
+   */
+  async getAccountContract(): Promise<Account<IAccountCore>> {
+    const signer = await this.getSigner();
+    const network = (await (await this.getProvider()).getNetwork()).chainId;
+    const sdk = ThirdwebSDK.fromSigner(signer, network);
```
joaquim-verges, line 247: Was there no way to get the contract from account API ? Might need to update the signer on it but otherwise it should work.

That way it inherits the options passed in like abi

PR #1308, opened 2023-07-08
[SDK] Base Router implementation for dynamic contracts

kumaryash90:


packages/sdk/src/evm/contracts/smart-contract.ts
```diff
@@ -335,6 +338,10 @@ export class SmartContract<
     return assertEnabled(this.detectAccount(), FEATURE_ACCOUNT);
   }
 
+  get baseRouter(): BaseRouterClass<BaseRouter> {
```
joaquim-verges, line 341: lets call this `extensions` 

and make the api
`contract.extensions.add("DirectListingLogic")`
`contract.extensions.update("DirectListingLogic")`
`contract.extensions.remove("DirectListingLogic")`

publisher as optional, default to deployer.thirdweb.eth

packages/sdk/src/evm/core/classes/base-router.ts
```diff
@@ -110,7 +110,7 @@ export class BaseRouterClass<TContract extends BaseRouter>
     },
   );
 
-  addExtensionWithAbi = /* @__PURE__ */ buildTransactionFunction(
+  addWithAbi = /* @__PURE__ */ buildTransactionFunction(
```
joaquim-verges, line 113: how can `extensionAbi` be optional here? you def need it.

also, can you add `addPublished(extensionName, publisher?)` ? that one should pull the abi and uri from publish metadata

packages/sdk/src/evm/core/classes/base-router.ts
```diff
@@ -0,0 +1,190 @@
...
+      });
+    },
+  );
+
+  addWithAbi = /* @__PURE__ */ buildTransactionFunction(
+    async (
+      extensionName: string,
+      extensionAddress: string,
+      extensionAbi?: ContractInterface,
+      extensionMetadataUri: string = "",
```
joaquim-verges, line 118: make this a metadata object, and use storage.upload() to get the URI rather than making ppl pass a URI.

think about the most convenient inputs for the user

packages/sdk/src/evm/contracts/smart-contract.ts
```diff
@@ -407,6 +407,17 @@ export class SmartContract<
     this.contractWrapper.updateSignerOrProvider(network);
   }
 
+  async onAbiUpdated(updatedAbi: ContractInterface): Promise<void> {
```
joaquim-verges, line 410: you don't need this at this level i think, i would put the `updateAbi()` function directly inside `contractWrapper` and make `this.abi = this.contractWrapper.abi` so it always reflects what's in contractWrapper. That way you don't need it to do callback style since you have this.contractWrapper everyhwere

packages/sdk/src/evm/core/classes/base-router.ts
```diff
@@ -12,25 +12,45 @@ import {
...
+  private onAbiUpdated: any;
 
-  constructor(contractWrapper: ContractWrapper<TContract>) {
+  constructor(
+    contractWrapper: ContractWrapper<TContract>,
+    contractInstance: any,
+    onAbiUpdated: any,
+  ) {
     this.contractWrapper = contractWrapper;
+    this.onAbiUpdated = onAbiUpdated.bind(contractInstance);
```
joaquim-verges, line 39: you don't need this

packages/sdk/src/evm/core/classes/base-router.ts
```diff
@@ -101,11 +121,43 @@ export class BaseRouterClass<TContract extends IBaseRouter>
...
+          ).abi;
+
+          const abiToAdd = this.filterAbiForAdd(extensionAbi, extension);
+
+          const updatedAbi = joinABIs([
+            AbiSchema.parse(this.contractWrapper.abi),
+            abiToAdd,
+          ]);
+
+          await this.onAbiUpdated(updatedAbi);
```
joaquim-verges, line 156: here, you just do `this.contractWrapper.updateAbi(updatedAbi)`

PR #1320, opened 2023-07-11
[Chains] Update chains: Linea Mainnet

kumaryash90:


packages/chains/data/additional/59144.mjs
```diff
@@ -0,0 +1,32 @@
...
+  chain: "Linea Mainnet",
+  shortName: "linea-mainnet",
+  chainId: 59144,
+  testnet: false,
+  nativeCurrency: {
+    name: "Ether",
+    symbol: "ETH",
+    decimals: 18,
+  },
+  rpc: ["https://rpc.linea.build"],
```
jnsdls, line 16: don't add this yet since it doesn't work (otherwise RPC edge will attempt to use this)

jnsdls (commented):
@kumaryash90 one small change then shipit

packages/chains/data/additional/59144.mjs
```diff
@@ -0,0 +1,32 @@
...
+  },
+  rpc: [],
+  explorers: [
+    {
+      name: "Linea Scan",
+      url: "https://lineascan.build",
+      standard: "",
+    },
+  ],
+  faucets: ["https://www.infura.io/faucet/linea"],
```
nachoiacovino, line 24: faucet for mainnet?

kumaryash90: lol, they filled it in the form

PR #1321, opened 2023-07-11
[WIP] gateway subdomains

joaquim-verges:


packages/storage/src/core/downloaders/storage-downloader.ts
```diff
@@ -15,6 +19,12 @@ import fetch from "cross-fetch";
  * @public
  */
 export class StorageDownloader implements IStorageDownloader {
+  private apiKey?: string;
+
+  constructor(options: IpfsDownloaderOptions) {
+    this.apiKey = options.apiKey;
```
iketw, line 25: why do we need to store this api key here?

packages/storage/src/types/upload.ts
```diff
@@ -53,6 +53,17 @@ export type IpfsUploaderOptions = {
   apiKey?: string;
 };
 
+/**
+ * @public
+ */
+export type IpfsDownloaderOptions = {
+  /**
+   * Optional apiKey to associate with the IpfsDownloader.
+   * You can get an API key here: https://thirdweb.com/settings
```
iketw, line 62: The url should be `https://thirdweb.com/dashboard/`, in case we want to change it later

PR #1322, opened 2023-07-12
[Storage] Use new upload server for uploads

adam-maj:


packages/storage/src/common/urls.ts
```diff
@@ -25,6 +25,12 @@ export const DEFAULT_GATEWAY_URLS: GatewayUrls = {
  */
 export const TW_IPFS_SERVER_URL = "https://upload.nftlabs.co";
 
+/**
+ * @internal
+ */
+export const TW_UPLOAD_SERVER_URL =
+  "https://storage-server-babv-am-pinata.zeet-nftlabs.zeet.app";
```
joaquim-verges, line 32: Reminder to update url before merging to main

packages/storage/src/common/urls.ts
```diff
@@ -25,6 +25,11 @@ export const DEFAULT_GATEWAY_URLS: GatewayUrls = {
  */
 export const TW_IPFS_SERVER_URL = "https://upload.nftlabs.co";
```
iketw, line 26: Can we then remove this const and the PINATA_IPFS_URL below?

adam-maj: Not sure if this is used externally is the reason I didn't use it. I have a vague memory of some other places using it outside the storage SDK (unsure if it's accurate).

dalmendray: Just searching in the whole repo should let you know right? I don't think we use it anywhere else

PR #1323, opened 2023-07-12
[SDK, React] mass rename thirdwebApiKey -> apiKey

joaquim-verges:


packages/chains/src/utils.ts
```diff
@@ -47,14 +45,9 @@ export function getChainRPCs(
     }
 
     // Replace API_KEY placeholder with value
-    if (thirdwebApiKey && rpc.includes("${THIRDWEB_API_KEY}")) {
-      processedRPCs.push(rpc.replace("${THIRDWEB_API_KEY}", thirdwebApiKey));
-    } else if (infuraApiKey && rpc.includes("${INFURA_API_KEY}")) {
-      processedRPCs.push(rpc.replace("${INFURA_API_KEY}", infuraApiKey));
-    } else if (alchemyApiKey && rpc.includes("${ALCHEMY_API_KEY}")) {
-      processedRPCs.push(rpc.replace("${ALCHEMY_API_KEY}", alchemyApiKey));
```
jnsdls, line 55: @joaquim-verges we use this in rpc edge :(

joaquim-verges: Ah good catch. I'll fix it tomorrow. Can still be here in chains package, just not passed from SDK.

PR #1324, opened 2023-07-12
[SDK] Add ERC6551 Interface

ciaranightingale:


packages/wallets/src/evm/connectors/token-bound-smart-wallet/index.ts
```diff
@@ -0,0 +1,73 @@
...
+        await super.initialize(personalWallet, factoryAddress);
+    }
+
+    protected defaultAccountInfo(): AccountContractInfo {
+        return {
+            execute: async (account, target, value, data) => {
+                return account.prepare("executeCall", [target, value, data]);
+            },
+            getNonce: async (account) => {
+                return account.call("nonce", []);
```
iketw, line 43: Dumb question but in the parent SmartWallet class, the calls here are: `execute` and `getNonce`, just making sure these are different on purpose

packages/wallets/src/evm/wallets/smart-wallet.ts
```diff
@@ -42,8 +41,8 @@ export class SmartWallet
     return "Smart Wallet" as const;
   }
 
-  constructor(options: WalletOptions<SmartWalletConfig>) {
-    super(SmartWallet.id, {
+  constructor(options: WalletOptions<SmartWalletConfig>, id?: string) {
+    super(id || SmartWallet.id, {
```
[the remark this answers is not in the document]
ciaranightingale: I added this to reflect these changes as well in `token-bound-smart-wallet.ts`:
```
constructor(options: WalletOptions<TokenBoundSmartWalletConfig>) {
        options.walletId = walletIds.tokenBoundSmartWallet;
        super(options
        );
    }
```

packages/wallets/src/evm/connectors/token-bound-smart-wallet/index.ts
```diff
@@ -0,0 +1,73 @@
...
+            paymasterAPI: config.paymasterAPI,
+            entryPointAddress: config.entryPointAddress,
+        }
+        super(smartWalletConfig);
+        this.config = config;
+    }
+
+    async initialize(personalWallet: EVMWallet) {
+        const factoryAddress = this.config.factoryAddress || "0x02101dfB77FDE026414827Fdc604ddAF224F0921";
+        await super.initialize(personalWallet, factoryAddress);
```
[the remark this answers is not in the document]
ciaranightingale: The behaviour I was trying to achieve here was to allow users to pass a factory address if they prefer but to provide the 6551 official address by default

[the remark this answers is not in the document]
ciaranightingale: I was following the same format as bundlerUrl etc

packages/wallets/src/evm/connectors/token-bound-smart-wallet/index.ts
ciaranightingale: Yeah in the official registry, the function signatures are different - they are of this format

ciaranightingale: Can you double check that these can be configured back if they so choose 

@joaquim-verges could you check this please - my assumption was to use the erc-6551 registry function signatures by default rather than the ones we have - is this the route we are choosing to go? SHould we have a `custom` option or something which uses the thirdweb defaults? I just wanted to keep things simple but I'm not sure

packages/wallets/src/evm/connectors/token-bound-smart-wallet/index.ts
```diff
@@ -0,0 +1,76 @@
...
+      createAccount: async (
+        factory,
+        owner,
+        {
+          implementation = this.config.implementation,
+          chainId = this.chainId,
+          tokenContract = this.config.tokenContract,
+          tokenId = this.config.tokenId,
+          salt = 1,
+        },
```
joaquim-verges, line 44: I don't think we need an extra parameter to this function here. We can just pass the data from `this.config` directly to the contract call

that way you don't have to add the ugly `optionalParam: any` on the interface

packages/wallets/src/evm/connectors/smart-wallet/types.ts
```diff
@@ -82,6 +82,7 @@ export type FactoryContractInfo = {
   createAccount: (
     factory: SmartContract,
     owner: string,
+    optionalParam?: any
```
joaquim-verges, line 85: we don't need this

packages/wallets/src/evm/connectors/token-bound-smart-wallet/types.ts
```diff
@@ -0,0 +1,51 @@
...
+  abi?: ContractInterface;
+  createAccount: (
+    factory: SmartContract,
+    owner: string,
+    tokenInfo: {
+      implementation: string;
+      chainId: Number;
+      tokenContract: string;
+      tokenId: Number;
+      salt?: Number;
```
joaquim-verges, line 26: don't need this here. just pass it in the `TokenBoundSmartWalletConfig`

packages/wallets/src/evm/connectors/token-bound-smart-wallet/types.ts
```diff
@@ -0,0 +1,51 @@
...
+    },
+  ) => Promise<Transaction>;
+  getAccountAddress: (
+    factory: SmartContract,
+    owner: string,
+    tokenInfo: {
```
joaquim-verges, line 32: same here

packages/wallets/src/evm/connectors/token-bound-smart-wallet/types.ts
```diff
@@ -0,0 +1,51 @@
...
+      implementation: string;
+      chainId: Number;
+      tokenContract: string;
+      tokenId: Number;
+      salt: Number;
+    },
+  ) => Promise<string>;
+};
+
+export type AccountContractInfo = {
```
joaquim-verges, line 42: what's diff here? should be able to re-use all the types from smart wallet

PR #1345, opened 2023-07-14
[service-utils] - rewrite

jnsdls:


packages/service-utils/src/cf-worker/index.ts
```diff
@@ -0,0 +1,150 @@
...
+  return {
+    secretKey,
+    clientId,
+    origin,
+    bundleId,
+    secretKeyHash,
+  };
+}
+
+export async function hashSecretKey(secretKey: string) {
```
assimovt, line 136: nice that these are here now per adapter 👍🏽

packages/service-utils/src/core/api.ts
```diff
@@ -0,0 +1,55 @@
...
+    message: string;
+  };
+};
+
+export async function fetchKeyMetadataFromApi(
+  clientId: string,
+  config: CoreServiceConfig,
+): Promise<ApiResponse> {
+  const { apiUrl, serviceScope, serviceApiKey } = config;
+  const url = new URL(`${apiUrl}/api/v1/keys`);
```
assimovt, line 39: its no longer /use?

packages/service-utils/src/core/authorize/index.ts
```diff
@@ -0,0 +1,152 @@
...
+        if ("updatedAt" in parsed) {
+          // we want to compare the updatedAt time to the current time
+          // if the difference is greater than the cacheTtl we want to ignore the cached data
+          const now = Date.now();
+          const diff = now - parsed.updatedAt;
+          const cacheTtl = cacheOptions.cacheTtlSeconds * 1000;
+          // only if the diff is less than the cacheTtl do we want to use the cached key
+          if (diff < cacheTtl * 1000) {
+            apiKeyMeta = parsed.apiKeyMeta;
+          }
```
assimovt, line 63: great you're handling it now 👏🏽

packages/service-utils/src/core/authorize/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      errorCode: clientAuth.errorCode,
+      authorized: false,
+      status: 401,
+      errorMessage: clientAuth.errorMessage,
+    };
+  }
+
+  // if we've made it this far we need to check service specific authorization
+  const serviceAuth = authorizeService(apiKeyMeta, serviceConfig, {
+    targetAddress: authData.targetAddress,
```
assimovt, line 135: cant there be a case with array of addresses? bundler passes an arr btw, what if we have service provide multiple contract addrs?

assimovt (commented):
awesome rewrite, much cleaner now and more abstract. left some comments

packages/service-utils/src/core/api.ts
jnsdls: yeah I was just wondering why it no longer works :D

jnsdls: fixing

packages/service-utils/src/core/authorize/index.ts
jnsdls: no we are comparing the targetAddress of the current request (per request) to the configuration of the service on the api key - I'm pretty confident this is correct

PR #1358, opened 2023-07-17
feat(chain): add iotex chain icon

DLHTX:


jnsdls:
hi @DLHTX, could you instead add this into `data/overrides` (otherwise it will get overridden by the automated syncing of this package)

DLHTX:
@jnsdls Hi I have added in my latest submission

PR #1362, opened 2023-07-18
[RN/Storage] Use storage.downloader in RN and pass the clientId

iketw:


packages/storage/src/core/downloaders/storage-downloader.ts
```diff
@@ -70,27 +70,27 @@ export class StorageDownloader implements IStorageDownloader {
       console.warn(`Retrying download with backup gateway URL: ${resolvedUri}`);
     }
 
+    let headers;
+    if (isTwGatewayUrl(resolvedUri)) {
+      if (this.secretKey) {
+        headers = { "x-secret-key": this.secretKey };
+      } else if (this.clientId) {
+        // @ts-ignore
+        if (globalThis.APP_BUNDLE_ID) {
```
jnsdls, line 79:
```suggestion
        if ("APP_BUNDLE_ID" in globalThis) {
```

jnsdls: check if they key exist In the object instead of trying to access directly. maybe even better if globalThis is defined first as well `typeof globalThis !== "undefined"` (should always be the case but just trying to be sure)

jnsdls (approved):
overall good but this might make it safer?

PR #1369, opened 2023-07-19
[Wallets] Add transfer & balance

adam-maj:


packages/wallets/src/evm/utils/currency.ts
```diff
@@ -0,0 +1,60 @@
...
+  let decimals = 18;
+  if (!isNativeToken(currencyAddress)) {
+    const erc20 = createErc20(provider, currencyAddress);
+    decimals = (await erc20.decimals()) as number;
+  }
+
+  return utils.parseUnits(BigNumber.from(price).toString(), decimals);
+}
+
+export async function fetchCurrencyValue(
```
joaquim-verges, line 44: these weren't exported already? We can just use the ones from the SDK no?

PR #1386, opened 2023-07-24
Service Utils Response Updates

farhanW3:
Updated the response messages on service utils to include more details and help user understand it.

### DIFF
--- a/packages/service-utils/src/core/authorize/client.ts
+++ b/packages/service-utils/src/core/authorize/client.ts
@@ -57,7 +57,7 @@ export function authorizeClient(
 
     return {
       authorized: false,
-      errorMessage: `The domain: ${origin}, is not authorized for this key. Please update your key permissions on the Thirdweb Dashboard`,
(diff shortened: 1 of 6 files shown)


packages/service-utils/src/core/authorize/client.ts
```diff
@@ -57,7 +57,7 @@ export function authorizeClient(
 
     return {
       authorized: false,
-      errorMessage: "The origin is not authorized for this key.",
+      errorMessage: `The domain: ${origin}, is not authorized for this key. Please update your key permissions on the Thirdweb Dashboard`,
```
joaquim-verges, line 60: lowercase `thirdweb dashboard` pls :)

in general we try to keep `thirdweb` lowercase everywhere

joaquim-verges (approved):
much better ty!

packages/service-utils/src/core/authorize/client.ts
farhanW3: cool. will remember now :)

PR #1391, opened 2023-07-25
[RN] Extend thirdweb SDK and pass the correct RN storage

iketw:


packages/react-native/src/evm/providers/thirdweb-provider.tsx
```diff
@@ -79,7 +80,12 @@ export const ThirdwebProvider = <
     <ThirdwebProviderCore
       supportedWallets={supportedWallets}
       storageInterface={
-        storageInterface || new ThirdwebStorage({ clientId: clientId })
+        storageInterface ||
+        new ThirdwebStorage({
+          clientId: clientId,
+          /// @ts-ignore
+          gatewayUrls: sdkOptions?.supportedChains,
```
joaquim-verges, line 87: this seems wrong

iketw: jeez, I was testing other props to see if typescript recognized them and left this :(

PR #1398, opened 2023-07-26
[SDK] Check process exists before accessing it

iketw:


packages/sdk/src/core/utils/process.ts
```diff
@@ -0,0 +1,15 @@
+export function getProcessEnv(key: string, defaultValue = "") {
+  if (typeof process !== undefined) {
```
jnsdls, line 2:
```suggestion
  if (typeof process !== "undefined") {
```

gotta compare to string value (typeof returns a string)

packages/sdk/src/core/utils/process.ts
```diff
@@ -0,0 +1,15 @@
...
+    if (process.env[key]) {
+      return process.env[key] as string;
+    }
+  }
+
+  return defaultValue;
+}
+
+export function setProcessEnv(key: string, value: string) {
+  if (typeof process !== undefined) {
```
jnsdls, line 12:
```suggestion
  if (typeof process !== "undefined") {
```

packages/sdk/src/evm/constants/addresses/getContractPublisherAddress.ts
```diff
@@ -1,13 +1,13 @@
...
 /**
  * @internal
  */
 export function getContractPublisherAddress() {
-  // eslint-disable-next-line turbo/no-undeclared-env-vars
-  if (process.env.contractPublisherAddress) {
-    // eslint-disable-next-line turbo/no-undeclared-env-vars
-    return process.env.contractPublisherAddress as string;
+  if (getProcessEnv("contractPublisherAddress")) {
+    return getProcessEnv("contractPublisherAddress");
```
jnsdls, line 10: this looks weird, why not pass the `ContractPublisher_address` below to the default value you added to the getEnv?

packages/sdk/src/evm/constants/addresses/getMultichainRegistryAddress.ts
```diff
@@ -1,13 +1,13 @@
...
 /**
  * @internal
  */
 export function getMultichainRegistryAddress() {
-  // eslint-disable-next-line turbo/no-undeclared-env-vars
-  if (process.env.multiChainRegistryAddress) {
-    // eslint-disable-next-line turbo/no-undeclared-env-vars
-    return process.env.multiChainRegistryAddress as string;
+  if (getProcessEnv("multiChainRegistryAddress")) {
+    return getProcessEnv("multiChainRegistryAddress");
```
jnsdls, line 10: same thing here, looks super weird

packages/sdk/src/evm/core/sdk.ts
```diff
@@ -2130,7 +2134,6 @@ export class ContractDeployer extends RPCConnectionHandler {
   }
 
   private hasLocalFactory() {
-    // eslint-disable-next-line turbo/no-undeclared-env-vars
-    return process.env.factoryAddress !== undefined;
+    return getProcessEnv("factoryAddress", undefined) !== undefined;
```
jnsdls, line 2137:
```suggestion
    return !!getProcessEnv("factoryAddress")
```

packages/sdk/test/evm/any-evm-keyless.test.ts
```diff
@@ -14,9 +17,11 @@ describe("Any EVM Keyless Deploy", async () => {
...
 
   async function deployTieredDrop() {
-    const mockPublisher = process.env.contractPublisherAddress;
-    process.env.contractPublisherAddress =
-      "0x664244560eBa21Bf82d7150C791bE1AbcD5B4cd7";
+    const mockPublisher = getProcessEnv(CONTRACT_PUBLISHER_ADDRESS_KEY);
+    setProcessEnv(
+      CONTRACT_PUBLISHER_ADDRESS_KEY,
+      "0x664244560eBa21Bf82d7150C791bE1AbcD5B4cd7",
+    );
```
jnsdls, line 24: I don't think you need to change it in tests since tests will always have `process.env`)

jnsdls (changes requested):
- the typeof checks are incorrect (need to check string)
- I don't think we should touch test code, only runtime code
- give we don't touch tests we don't need the `setProcessEnv` function (feels weird to ever set to env from runtime anyways)

packages/sdk/src/evm/constants/addresses/getContractPublisherAddress.ts
iketw: I like it

jnsdls (approved):
LGTM

PR #1402, opened 2023-07-26
Use auth token for CLI

Marfuen:


packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
+    domain: "localhost:3000",
+    address: walletAddress,
+  });
+
+  // In this case the state is the loggedIn object.
+  const ourState = encodeURIComponent(JSON.stringify(loggedIn));
+  const urlToOpen =
+    // `https://thirdweb.com/cli/login?from=cli&#${ourState}`;
+    // `https://thirdweb-www-git-mariano-api-keys-sign-in.thirdweb-preview.com/cli/login?from=cli&#${ourState}`;
+    `http://localhost:3000/cli/login?from=cli&#${ourState}`;
```
jnsdls, line 102: don't need `from=cli` if you already have `/cli` as the path, right? should be clear where we're from at that point

either:
`/login/cli?payload=<the payload>&state=<the separate state value>`
or:
`/login?from=cli&payload=<the payload>&state<the separate state value>`

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
+
+  let server: http.Server;
+  let loginTimeoutHandle: NodeJS.Timeout;
+  const timerPromise = new Promise<void>((resolve, reject) => {
+    loginTimeoutHandle = setTimeout(() => {
+      logger.error("Timed out waiting, please try again.");
```
jnsdls, line 108: does this error make it to the CLI output? if so let's provide more detail

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
+      switch (pathname) {
+        case "/auth/callback": {
+          if (query.token) {
+            const token = Array.isArray(query.token) ? query.token[0] : query.token;
+            const theirState = Array.isArray(query.state)
+              ? query.state[0]
+              : query.state;
+
+            // Check if the payload sent back is the same as the one we sent.
+            if (theirState !== JSON.stringify(loggedIn)) {
```
jnsdls, line 139: this needs to compare to the state not the login payload, cannot use login payload for state

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
+              );
+            } else {
+              // Save the token to the config file.
+              const configJson = JSON.parse(fs.readFileSync(credsConfigPath, "utf-8"));
+              fs.writeFileSync(credsConfigPath, JSON.stringify({
+                ...configJson,
+                token
+              }), {
+                encoding: "utf8",
+                mode: 0o600,
```
jnsdls, line 155: what's this mode do and why do we need it?

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
+              res.writeHead(400, { "Content-Type": "text/plain" }); // send a 400 response
+              res.end("Unauthorized request, state mismatch", () => {
+                finish(new Error("Unauthorized request, state mismatch"));
+              });
+              reject(
+                new Error(chalk.red("Unauthorized request, state mismatch")),
+              );
+            } else {
+              // Save the token to the config file.
+              const configJson = JSON.parse(fs.readFileSync(credsConfigPath, "utf-8"));
```
jnsdls, line 149: prob should just keep the token in its own file instead of mixed with other configuration? that way you can also just delete the file on logout and in this case you don't have to read first, just write the token directly to the file

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
-  } catch (error) {
-    console.log(error);
+    server.listen(8976);
+  });
+  if (props?.browser) {
+    ora(`Opening a link in your default browser: ${urlToOpen}\n`).start();
+    // Adding this timeout since it feels weird for the browser to open before the spinner.
+    setTimeout(async () => {
+      await open(urlToOpen);
+    }, 2000);
```
jnsdls, line 181: why the timeout?

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
-      process.exit(1);
-    }
-    cache.set("api-secret-key", response.apiSecretKey);
-    return response.apiSecretKey;
-  } catch (error) {
-    console.log(error);
+    server.listen(8976);
+  });
+  if (props?.browser) {
+    ora(`Opening a link in your default browser: ${urlToOpen}\n`).start();
```
jnsdls, line 177: this starts a spinner, do we ever "end" it? why a spinner in the first place?

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
+  const configJson = JSON.parse(fs.readFileSync(credsConfigPath, "utf-8"));
+
+  if (configExists && configJson.password) {
+    return configJson.password;
   }
+
+  const response = await prompts({
+    type: "invisible",
+    name: "password",
+    message: `Please enter a password to authenticate with the CLI`,
```
jnsdls, line 200: when will this get prompted? when user creates a wallet first time or when unlocking? can we make the message clearer?

packages/cli/src/cli/index.ts
```diff
@@ -302,16 +338,17 @@ const main = async () => {
     .option("--zksync", "Deploy on ZKSync")
     .option("-k, --key <key>", "API key to authorize usage")
```
jnsdls, line 339: rename this to `secret key`

packages/cli/src/cli/index.ts
```diff
@@ -302,16 +338,17 @@ const main = async () => {
...
       // If no key is passed in, prompt the user to login. If it is passed in, use it.
       if (!options.key) {
-        apiSecretKey = await loginUser(cache);
+        authToken = await loginUser(credsConfigPath, cliWalletPath);
       } else {
-        await validateKey(options.key);
-        apiSecretKey = options.key;
+        // Don't need to validate anymore? Also, should probably rely on the underlying service to throw that error.
+        // await validateToken(options.key);
+        authToken = options.key;
```
jnsdls, line 348: we'll want to support either auth token *or* secret key (secret key still makes sense for CI I think)

packages/sdk/src/evm/constants/urls.ts
```diff
@@ -173,6 +173,8 @@ export function getProviderFromRpcUrl(
         }
       } else if (sdkOptions?.secretKey) {
         headers["x-secret-key"] = sdkOptions?.secretKey;
+      } else if (sdkOptions?.authToken) {
+        headers["Authorization"] = `Bearer ${sdkOptions?.authToken}`;
```
jnsdls, line 177: yep nice

packages/sdk/src/evm/core/sdk.ts
```diff
@@ -262,11 +262,12 @@ export class ThirdwebSDK extends RPCConnectionHandler {
     options: SDKOptions = {},
     storage?: IThirdwebStorage,
   ) {
-    const apiKeyType = typeof window !== "undefined" ? "clientId" : "secretKey";
+    const apiKeyType = typeof window !== "undefined" ? "clientId" : options.authToken ? "authToken" : "secretKey";
```
jnsdls, line 265: I don't think we want to check `authToken` here. the option to pass in the `authToken` is probably more internal to our own usage of SDKs than to external usage of the SDK

packages/service-utils/src/core/authorize/index.ts
```diff
@@ -34,31 +35,31 @@ export async function authorize(
...
-  if (!enforceAuth && !clientId && !secretKeyHash) {
+  if (!enforceAuth && !clientId && !secretKeyHash && !authToken) {
     return {
       authorized: true,
       apiKeyMeta: null,
     };
   }
 
-  // if we don't have a client id at this point we can't authorize
-  if (!clientId) {
```
jnsdls, line 50: this file needs another look in its entirety now that we have auth token to consider

packages/storage/src/common/urls.ts
```diff
@@ -93,12 +93,17 @@ export function prepareGatewayUrls(
...
+  authToken?: string,
 ): GatewayUrls {
   const allGatewayUrls = {
     ...DEFAULT_GATEWAY_URLS,
     ...gatewayUrls,
   };
 
+  if (authToken && typeof window !== "undefined") {
+    throw new Error("Cannot use authToken in browser context");
+  }
```
jnsdls, line 105: why?

packages/storage/src/core/uploaders/ipfs-uploader.ts
```diff
@@ -301,8 +311,16 @@ export class IpfsUploader implements IStorageUploader<IpfsUploadBatchOptions> {
       );
     }
 
+    if ((this.authToken && this.clientId) || (this.authToken && this.secretKey)) {
+      throw new Error(
+        "Cannot use both auth token and client ID or secret key. Please use auth token or secret key for server-side applications or clientId for client-side applications.",
+      );
+    }
```
jnsdls, line 318: this feels wrong

jnsdls: auth token = either server or client
clientId = exclusively client
secretKey = exclusively server

packages/wallets/src/evm/connectors/smart-wallet/lib/http-rpc-client.ts
```diff
@@ -33,7 +34,15 @@ export class HttpRpcClient {
         );
       }
 
-      if (secretKey) {
+      if (authToken && clientId || authToken && secretKey) {
+        throw new Error(
+          "Cannot use both auth token and client ID or secret key. Please use auth token or secret key for server-side applications and clientId for client-side applications.",
+        );
+      }
```
jnsdls, line 41: same, feels wrong

packages/wallets/src/evm/connectors/smart-wallet/lib/paymaster.ts
```diff
@@ -40,7 +43,15 @@ class VerifyingPaymasterAPI extends PaymasterAPI {
         );
       }
 
-      if (this.secretKey) {
+      if ((this.authToken && this.clientId) || (this.authToken && this.secretKey)) {
+        throw new Error(
+          "Cannot use both auth token and client ID or secret key. Please use auth token or secret key for server-side applications and clientId for client-side applications."
+        );
+      }
```
jnsdls, line 50: feels wrong

pnpm-lock.yaml
```diff
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
```
jnsdls, line 6: please update your pnpm version :)

jnsdls (changes requested):
good start, let's review the auth library changes in person - rest I left a bunch of feedback for

packages/chains/src/utils.ts
```diff
@@ -76,6 +76,7 @@ export function getChainRPCs(
  * @param chain - The chain config to assemble RPC URLs from
  * @param clientId - The client id to use for the RPC URL
  * @param secretKey - The secret key to use for the RPC URL
+ * @param authToken - The auth token to use for the RPC URL
```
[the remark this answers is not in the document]
Marfuen: No idea why it's defined here, but yeah looks pointless.

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
   } else {
-    const apiKey = await createSession(cache);
-    return apiKey;
+    if (showLogs) {
+      console.log(
+        chalk.yellow(
+          "We did not find a session, please connect your wallet through our dashboard to continue",
+        ),
+      );
+    }
```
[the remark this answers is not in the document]
Marfuen: Yeah, I was only showing them when running the login function, so that if you have a session and this runs it won't show those logs everytime and just let you through.

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
-    cache.remove("api-secret-key");
+    const dirExists = fs.existsSync(credsConfigPath);
+    if (!dirExists) {
+      return;
+    }
+    const configJson = fs.readFileSync(credsConfigPath, "utf-8");
+    const parsedJson = JSON.parse(configJson);
+    parsedJson.password = "";
+    parsedJson.token = "";
+    fs.writeFileSync(credsConfigPath, JSON.stringify(parsedJson), "utf-8");
```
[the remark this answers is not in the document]
Marfuen: true

packages/cli/src/auth/index.ts
```diff
@@ -1,86 +1,243 @@
...
+  const { credsConfigPath, cliWalletPath } = props;
+
+  // Get or generate a localwallet.
+  const wallet = await getOrGenerateLocalWallet(credsConfigPath, cliWalletPath);
+  const walletAddress = await wallet.getAddress();
+  const auth = new ThirdwebAuth(wallet, "localhost:3000");
+
+  // Generate the login payload to pass to the dashboard.
+  const loggedIn = await auth.login({
+    domain: "localhost:3000",
```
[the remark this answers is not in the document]
Marfuen: yeah, the domain of the dashboard since the token will be generated from there

packages/cli/src/auth/index.ts
Marfuen: This makes sure that the permissions of the file being written is set to 600, to ensure only admins of that computer will be able to read/write from it.

packages/storage/src/core/uploaders/ipfs-uploader.ts
Marfuen: I thought we were saying that the auth token should be treated like a secret, so we shouldn't allow it client side?

packages/cli/src/auth/index.ts
Marfuen: To give the user some time to read what the text says before opening up the browser

packages/cli/src/auth/index.ts
Marfuen: yup agreed

packages/cli/src/auth/index.ts
Marfuen: Since this is waiting for the dashboard to respond, but I guess the spinner doesn't make sense on opening the link, I'll move it to a different message.

PR #1403, opened 2023-07-26
[Service Utils] : Origin Check Update

farhanW3:
updated Origin check to fallback to URL hostname if Origin/Referer not available

### DIFF
--- a/.changeset/hot-months-switch.md
+++ b/.changeset/hot-months-switch.md
@@ -1,0 +1,5 @@
+---
+"@thirdweb-dev/service-utils": patch
+---
+
+Updated Origin check to fallback to URL hostname if both Origin & Referer are not available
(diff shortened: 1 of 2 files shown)


jnsdls (commented):
the changes here make sense, please double check where we actually check the origin it won't reject every request because we will now *always* have an origin (that may not be allowed on the key)

farhanW3:
> the changes here make sense, please double check where we actually check the origin it won't reject every request because we will now _always_ have an origin (that may not be allowed on the key)

Yup, I will play around with

I think, we need to make sure we don't allow `*.thirdweb.com , *` in Allowed Domains otherwise it defeats the purpose. So it can either be `*` or `*.thirdweb.com, ipfscd.io (and more)`.

PR #1404, opened 2023-07-26
[Storage/RN] Read custom upload url from env

iketw:


packages/storage/src/common/urls.ts
```diff
@@ -39,6 +40,10 @@ export const DEFAULT_GATEWAY_URLS: GatewayUrls = {
  */
 export const TW_UPLOAD_SERVER_URL = "https://storage.thirdweb.com";
 
+export const CUSTOM_UPLOAD_SERVER_URL = getProcessEnv(
+  "CUSTOM_UPLOAD_SERVER_URL",
+);
```
jnsdls, line 45: why not do the defaultValue strategy here, why does the underlying consumer need to know about 2 values?

```js
export const TW_UPLOAD_SERVER_URL = getProcessEnv(
  "CUSTOM_UPLOAD_SERVER_URL",
"https://storage.thirdweb.com");
```

is kinda what I was thinking?

jnsdls (commented):
I think we can simplify?

packages/storage/src/common/urls.ts
iketw: I thought about it but we need to know if there's a custom url or not because we shouldn't pass the clientId/secretKey in case there's a custom url

jnsdls (approved):
@iketw add a changeset and let's go

PR #1411, opened 2023-07-28
[service-utils] - allow account level auth via jwt

jnsdls:


packages/service-utils/src/core/authorize/index.ts
```diff
@@ -13,36 +15,124 @@ export type AuthorizationInput = {
...
   if (!enforceAuth && !clientId && !secretKeyHash) {
     return {
       authorized: true,
       apiKeyMeta: null,
+      accountMeta: null,
+    };
+  }
+  // if we come in with a JWT then we only check the account is valid
+  if (jwt && hashedJWT) {
+    let accountMeta: AccountMetadata | null = null;
```
farhanW3, line 64: should this be outside the above if?

farhanW3: I think, it good. the spacing was weird for when checking for me on browser

packages/storage/src/core/uploaders/ipfs-uploader.ts
```diff
@@ -270,18 +270,27 @@ export class IpfsUploader implements IStorageUploader<IpfsUploadBatchOptions> {
...
+          "x-bundle-id",
+          (globalThis as any).APP_BUNDLE_ID as string,
+        );
+      }
+
+      // if we have a authorization token on global context then add that to the headers
+      if (typeof globalThis !== "undefined" && "TW_AUTH_TOKEN" in globalThis) {
+        xhr.setRequestHeader(
+          "authorization",
+          (globalThis as any).TW_AUTH_TOKEN as string,
```
iketw, line 290: Do we need to specify Bearer?

PR #1412, opened 2023-07-28
fix: postinstall check keeps adding to itself

jarrodwatts:
Post install now adds `thirdweb@latest generate`, but checks for `thirdweb generate`, so every time you run it it adds to itself

### DIFF
--- a/.changeset/happy-poets-hang.md
+++ b/.changeset/happy-poets-hang.md
@@ -1,0 +1,5 @@
+---
+"thirdweb": patch
+---
+
+fix: postinstall check keeps adding to itself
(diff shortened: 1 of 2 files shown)


jarrodwatts:
The post-install also breaks with API key changes - hangs deployment as it requires `--key` flag with API key.

jnsdls:
hi @jarrodwatts! 👋 

so if I understand this correctly this fixes the adding to itself but not the key issue you pointed out yet?

Marfuen (approved):
Looks good to me, in the future we might want to use regex to check, but this works for now

jarrodwatts:
Yea @jnsdls the `--key` thing is still an issue

PR #1417, opened 2023-07-31
[AuthService] return accountId from authorize call

joaquim-verges:


assimovt (approved):

assimovt:
@joaquim-verges this #1411 already has this change

PR #1429, opened 2023-08-02
[React] Allow passing secretKey for server-side rendering

iketw:


packages/react/src/evm/providers/thirdweb-provider.tsx
```diff
@@ -56,7 +56,6 @@ interface ThirdwebProviderProps<TChains extends Chain[]>
 export const ThirdwebProvider = <
   TChains extends Chain[] = typeof defaultChains,
 >({
-  clientId: clientId,
```
iketw, line 59: No need to do this here, clientId will be passed down as part of `{...restProps}`

packages/react-core/src/evm/providers/thirdweb-sdk-provider.tsx
```diff
@@ -92,7 +93,7 @@ const WrappedThirdwebSDKProvider = <TChains extends Chain[]>({
       // sdk from chainId
       sdk_ = new ThirdwebSDK(
         chainId,
-        { ...mergedOptions, clientId },
+        { ...mergedOptions, clientId, secretKey },
```
iketw, line 96: In ThirdwebSDK is where we create the storage obj, we need the secretKey

MananTank (approved):
LGTM

PR #1443, opened 2023-08-03
Switch to domain with good reputation

nessup:


jnsdls (approved):
ship it

joaquim-verges (approved):

packages/cli/src/deploy/app.ts
```diff
@@ -72,7 +72,7 @@ export async function deployApp(
     const uri = await upload(storage, distPath);
     return `${uri.replace(
       "ipfs://",
-      "https://ipfs-public.thirdwebcdn.com/ipfs/",
+      "https://gateway.ipfscdn.io/ipfs/",
```
joaquim-verges, line 75: this should be using your client id based gateway

joaquim-verges (changes requested):

joaquim-verges (approved):

PR #1448, opened 2023-08-04
chore: Add logHttpRequest helper func

arcoraven:


packages/service-utils/src/cf-worker/logging.ts
```diff
@@ -0,0 +1,37 @@
...
+  res,
+  isAuthed,
+  error,
+}: {
+  source: string;
+  req: Request;
+  res: ServerResponse;
+  isAuthed?: boolean;
+  error?: any;
+}) {
```
nessup, line 17: We need to add an arg to allow callers to override clientId

.changeset/slimy-pumas-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/service-utils": minor
+---
+
+[service-utils] Add logHttpRequest helper func
```
jnsdls, line 5: don't need the prefix (it's already scoped by the packages above)
```suggestion
Add logHttpRequest helper func
```

packages/service-utils/src/cf-worker/logging.ts
```diff
@@ -0,0 +1,39 @@
+import type { ServerResponse } from "http";
+import { extractAuthorizationData } from ".";
```
jnsdls, line 2: should extract this from `index.ts` as well otherwise we'll end up with cyclic dependencies

packages/service-utils/src/node/logging.ts
```diff
@@ -0,0 +1,36 @@
+import type { ServerResponse } from "http";
+import { AuthInput, extractAuthorizationData } from ".";
```
jnsdls, line 2: should extract this from `index.ts` as well otherwise we'll end up with cyclic dependencies

packages/service-utils/src/node/logging.ts
```diff
@@ -0,0 +1,36 @@
...
+      clientId: authorizationData.clientId,
+      isAuthed: !!isAuthed ?? null,
+      status: res.statusCode,
+    }),
+  );
+
+  if (error) {
+    // Log to a separate line. Logpush has a character limit per log line.
+    console.error("Request error:", error);
+  }
```
jnsdls, line 35: `node` will never run in workers -> there is no logpush

(probably still makes sense to do the same pattern in node as we do in cf so the output is consistent but just wanted to call it out)

jnsdls (commented):
non-blocking feedback

PR #1456, opened 2023-08-05
Allow styles on MediaRenderer video tags

mykcryptodev:
### Description
The MediaRenderer component currently does not provide a way to style the video tag that is produced by the component (it only allows for styling a div that wraps the video component). Therefore, it is not possible to change the objectFit property of the video tag.

This PR allows for users to apply styles directly to the video tag if desired as well as optionally hide the play button that overlays the video media.

### Use Case:
My NFT Marketplace has a banner at the top of the page that is rendered via the MediaRenderer. 
Screenshot 2023-08-05 at 10 57 18 AM

I want to allow videos to cover that entire banner.

The react code looks like this:
Screenshot 2023-08-05 at 10 55 49 AM

But since the MediaRender component renders the video tag with hardcoded styles, my video does not cover the div.
Screenshot 2023-08-05 at 10 58 52 AM

https://github.com/thirdweb-dev/js/assets/89474773/06934f02-f372-45d1-a1cc-697c672998fe

### Proposed Solution:
Allow the user to directly apply styles to the video tag with the existing styles prop on the MediaRenderer.

Here is what my NFT Marketplace looks like after implementing this change:

https://github.com/thirdweb-dev/js/assets/89474773/50822881-d1ed-43c4-843a-82faa0ed2f50

As an aside, I also wanted to optionally remove the play button that gets painted over the video. If the user specifies controls to be false, the play button gets hidden as well.

Open to feedback on this PR!

### DIFF
--- a/packages/react/src/evm/components/MediaRenderer.tsx
+++ b/packages/react/src/evm/components/MediaRenderer.tsx
@@ -139,6 +139,7 @@ const VideoPlayer = /* @__PURE__ */ React.forwardRef<
             zIndex: 1,
             transition: "opacity .5s",
             opacity: !poster ? 1 : playing ? 1 : 0,
+            ...style,
           }}
         />
         {poster && (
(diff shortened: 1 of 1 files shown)


MananTank:
@mykcryptodev You can set a class on the `MediaRenderer` and target it's child `video` element using that class in CSS file

mykcryptodev:
> @mykcryptodev You can set a class on the `MediaRenderer` and target it's child `video` element using that class in CSS file.
> 
> Also, I'm not sure why you are using `MediaRenderer` for banner, It is only meant to render NFT media files to render media of multiple types, If you just want to render a simple video, you should just use the `video` element

Thanks for the tip. I added !important to get it to work without this PR. Before we close, can you let me know if there is a way to hide the play button?

MananTank:
@mykcryptodev There's no way to hide the button by passing any prop, you can hide it with CSS, but I suggest just using `<video>` here

PR #1458, opened 2023-08-06
Add option to hide the [Request Testnet Funds] button

kien-ngo:
Currently there's no way to hide the [Request Testnet Fund] manually. I don't want that button in production. It confuses the average users. So maybe we should allow an option to hide it?
image

@MananTank Would you be able to take a look at this?

### DIFF
--- a/packages/react-native/src/evm/components/ConnectWallet.tsx
+++ b/packages/react-native/src/evm/components/ConnectWallet.tsx
@@ -19,6 +19,11 @@ export type ConnectWalletProps = {
    * render custom rows in the Connect Wallet Details modal
    */
   extraRows?: ConnectWalletDetailsProps["extraRows"];
+
+  /**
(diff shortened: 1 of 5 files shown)


kien-ngo:
On another note, I'm facing this issue when running the test in local using yalc - seems completely unrelated with my changes. How do I fix/by pass/ignore it?

I tried these but didn't work:
- pnpm install (both at root and inside packages/contract-js)
- pnpm build (both at root and inside packages/contract-js)

image

MananTank:
@kien-ngo - To Fix the yalc issue in testing, Try deleting node modules in the app you are using it and then running `yalc update && yarn`

kien-ngo:
@MananTank Thanks a lot! That fixed it. I can see the code working now
image

MananTank:
LGTM, We will merge this soon. Thanks @kien-ngo !

PR #1461, opened 2023-08-07
Firekeeper/xai and large chainids

0xFirekeeper:


packages/unity-js-bridge/src/thirdweb-bridge.ts
```diff
@@ -299,37 +299,35 @@ class ThirdwebBridge implements TWBridge {
...
     personalWallet: PossibleWallet = "localWallet",
   ) {
     if (!this.activeSDK) {
       throw new Error("SDK not initialized");
     }
-    if (chainId === 0) {
-      chainId = undefined;
-    }
     const walletInstance = this.walletMap.get(wallet);
+    const chainIdNumber = Number(chainId);
```
iketw, line 311: From the title of this PR it seems we want to deal with large numbers, wouldn't trying to convert the string to a number have the same effect here? Also, can you share an example of the chainId you're doing this fix for?

joaquim-verges: I think this might blow up with a big enough number too... did you test this?

0xFirekeeper: typescript Number is good enough for C# BigInteger, MAX_SAFE_INTEGER in ts is already much bigger than max int c#

0xFirekeeper: tested with xai's 47279324479

joaquim-verges (approved):

packages/unity-js-bridge/src/thirdweb-bridge.ts
iketw: I see, so the issue is that we're calling `connect` from C# with a BigInteger that JS doesn't support, in which case we need a string. And we're saying that it's fine to convert to Number in JS since MAX int in JS covers most of the values of BigInteger in C#, correct?

0xFirekeeper: we were sending a C# int, now we send a C# BigInteger, both of which should mostly be in the range of TS Number, so it's ok to convert to support > C# int in bridge

PR #1467, opened 2023-08-08
Add subpath exports for better Next.js app router support 

MananTank:
## Problem solved

Importing components like `ThirdwebProvider` in Next.js app router directly in a server component throws error saying that it's not marked with "use client"

To fix that, we can create a separate file that exports the component and has a "use client" directive on top


MananTank:
Don't know why but the dist folder in thirdweb-provider is missing

https://www.npmjs.com/package/@thirdweb-dev/react/v/0.0.0-dev-9fb0855-20230808164127?activeTab=code

image


But if check locally with `npm pack`, and `dist` folder is there in generated package zip

 
image

MananTank:
Nope, still not fixed

MananTank:
ok got it - it's the turbo.json issue

MananTank:
Next.js is showing this error for some reason when using below code: 

```ts
import { ThirdwebProvider } from '@thirdweb-dev/react/ThirdwebProvider';
```

image

PR #1469, opened 2023-08-08
refactor(urls): Simplify authentication strategy handling

jnsdls:
## Problem solved

- RPC secret key was not being passed if clientId even existed on options
- clientID was not being computed when secret key was passed
- clientID was not being passed as /path

## Changes made

- [ ] secret key will now be evaluated first and compute client Id
- [ ] clientId will always be added to the path

## How to test

### DIFF
--- a/packages/sdk/src/evm/constants/urls.ts
+++ b/packages/sdk/src/evm/constants/urls.ts
@@ -159,25 +159,53 @@ export function getProviderFromRpcUrl(
 ) {
   try {
     const headers: HeadersInit = {};
+    // will be used to make sure we don't cache providers with different auth strategies
+    let authStrategy = "none";
+
     if (isTwUrl(rpcUrl)) {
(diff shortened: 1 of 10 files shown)


packages/sdk/src/evm/constants/urls.ts
```diff
@@ -159,25 +159,55 @@ export function getProviderFromRpcUrl(
...
+        if (typeof window !== "undefined") {
+          throw new Error("Cannot use secretKey in browser context");
         }
-      } else if (sdkOptions?.secretKey) {
-        headers["x-secret-key"] = sdkOptions?.secretKey;
+        // this is on purpose because we're using the crypto module only in node
+        // try to trick webpack :)
+        const pto = "pto";
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        const crypto = require("cry" + pto);
```
jnsdls, line 177: don't ask...

PR #1472, opened 2023-08-09
Add sdk information in headers for various services

joaquim-verges:
## Problem solved

We dont have any data on which SDK versions are using our services currently.

## Changes made

- [x] Public API changes: none
- [x] Internal API changes: added `x-sdk` headers to rpc/storage/bundler calls

## How to test

- [x] Manual tests: inspect request headers


joaquim-verges (commented):
Thank you @MananTank !

packages/wallets/src/evm/connectors/smart-wallet/lib/http-rpc-client.ts
```diff
@@ -2,8 +2,12 @@ import { providers, utils } from "ethers";
 import { UserOperationStruct } from "@account-abstraction/contracts";
 import { deepHexlify } from "@account-abstraction/utils";
 import { isTwUrl } from "../../../utils/url";
+import pkg from "../../../../../package.json";
 
 const DEBUG = false;
+function isBrowser() {
+  return typeof window !== "undefined";
```
iketw, line 9: This is too simple imo, platforms like ReactNative will also return true here. I recently discovered that this is a thing:

`global?.navigator?.product === 'ReactNative'`, so we can use this to differentiate react-native, browser, node

packages/storage/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "@thirdweb-dev/tsconfig/base.json",
+  "extends": "@thirdweb-dev/tsconfig/sdk.json",
```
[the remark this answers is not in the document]
joaquim-verges: Needed the Json resolution flag

packages/wallets/src/evm/connectors/smart-wallet/lib/http-rpc-client.ts
joaquim-verges: Ooh that's nice, can you add it for me?

Def want to differentiate reqs from react native, thanks for catching that

joaquim-verges: hmmm that gives me weird names in browser though. I think i'll just consider react native if we have a bundle id

iketw: Yeah, looking deeper it seems that the feature is deprecated: 

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/product

iketw: It seems we can do something like this though:

```
function getEnvironment() {
  if (typeof navigator !== 'undefined' && navigator.product === 'ReactNative') {
    return 'React Native';
  } else if (typeof window !== 'undefined') {
    return 'React';
  } else {
    return 'Unknown';
  }
}
```

PR #1476, opened 2023-08-09
Improve missing client id error

nessup:


.changeset/modern-mirrors-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/service-utils": major
```
nachoiacovino, line 2: this should be a patch

nachoiacovino (changes requested):

packages/service-utils/src/core/authorize/index.ts
```diff
@@ -141,7 +141,7 @@ export async function authorize(
     return {
       authorized: false,
       status: 401,
-      errorMessage: "Missing clientId or secretKey.",
+      errorMessage: "Please include a client id or secret key with your request. Starting August 1st, 2023 API Keys are required to use thirdweb services. Learn more: https://blog.thirdweb.com/changelog/api-keys-to-access-thirdweb-infra",
```
nachoiacovino, line 144: shouldn't we link to the https://thirdweb.com/create-api-key like we do in other places?

PR #1501, opened 2023-08-16
moved res.json() to avoid getting JSON error in console

farhanW3:
## Problem solved

Solving the JSON error message. Moved `res.json()` after `res.ok` check


.changeset/hip-walls-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/storage": patch
+---
+
+moved res.json() to avoid getting JSON error in console
```
jnsdls, line 5: changesets should ideally be useful for end-users of the packages

farhanW3: Got it. Did it in a hurry, will keep in mind. Thanks

PR #1515, opened 2023-08-21
Try/catch parseSnapshotInputs

nachoiacovino:
## Problem solved

Make it so we throw instead of silently failing when parsing

## Changes made

- [x] try/catch parseSnapshotInputs


joaquim-verges:
@nachoiacovino let me know if we still need this, closing for now

PR #1520, opened 2023-08-22
feat(sdk): centralize contract reads

ElasticBottle:
## Problem solved

This PR introduces a basic pattern for centralizing contract reads within the SDK itself.

## Changes made

- [ ] Public API changes: `NONE`
- [ ] Internal API changes: Instead of calling `this.contractWrapper.readContract.fnName()`, we now call `this.contractWrapper.read("fnName", []);`

## How to test

- [ ] Automated tests: To be updated

##  Discussion

- Have a `call` function that already exists today. Is this something that we want to use?  (not typesafe auto completion though 😢 )
    - Not used anywhere. Seems to be exposed as an external API through the pre-built contracts.

## Notes

- We don't need things to be too clean. will be swapped out with viem items in a bit
- Typescript limitation means we have something like [contract-sales.ts](https://github.com/thirdweb-dev/js/pull/1520/files#diff-e9e0cc9e7ffcc3798766eeba7b78fa939dd0c53692ca0bcfd655d4dc043b318eR41) for generic `ContracttWrapper` (we cannot have partial inferred optional types. I might update the types as I centralize more of the reads, but I think this should tide us over for now till Viem in which case we probbaly have to write our own etherChain adapter or swap to abiType anyways.

### DIFF
--- a/packages/sdk/src/evm/core/classes/contract-sales.ts
+++ b/packages/sdk/src/evm/core/classes/contract-sales.ts
@@ -38,7 +38,11 @@ export class ContractPrimarySale<TContract extends IPrimarySale>
    * @twfeature PrimarySale
    */
   public async getRecipient(): Promise<Address> {
-    return await this.contractWrapper.call("", []);
+    const result = await this.contractWrapper.read<
+      IPrimarySale,
+      "primarySaleRecipient"
+    >("primarySaleRecipient", []);
+    return result;
   }
 
   /**
--- a/packages/sdk/src/evm/core/classes/contract-sales.ts
+++ b/packages/sdk/src/evm/core/classes/contract-sales.ts
@@ -17,13 +17,11 @@ import { Transaction } from "./transactions";
  * ```
  * @public
  */
-export class ContractPrimarySale<TContract extends IPrimarySale>
-  implements DetectableFeature
-{
+export class ContractPrimarySale implements DetectableFeature {
   featureName = FEATURE_PRIMARY_SALE.name;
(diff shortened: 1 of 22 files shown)


packages/sdk/src/evm/core/classes/contract-sales.ts
```diff
@@ -38,7 +38,7 @@ export class ContractPrimarySale<TContract extends IPrimarySale>
    * @twfeature PrimarySale
    */
   public async getRecipient(): Promise<Address> {
-    return await this.contractWrapper.readContract.primarySaleRecipient();
+    return await this.contractWrapper.call("", []);
```
joaquim-verges, line 41: use read

packages/sdk/src/evm/core/classes/contract-sales.ts
```diff
@@ -38,7 +38,11 @@ export class ContractPrimarySale<TContract extends IPrimarySale>
    * @twfeature PrimarySale
    */
   public async getRecipient(): Promise<Address> {
-    return await this.contractWrapper.readContract.primarySaleRecipient();
+    const result = await this.contractWrapper.read<
+      IPrimarySale,
+      "primarySaleRecipient"
+    >("primarySaleRecipient", []);
+    return result;
```
joaquim-verges, line 45: oof yeah that's ugly. how often do you think we'll need this workaround? not clear what's causing the type issue

joaquim-verges (approved):
type thing is a bit weird, but lets merge and iterate

PR #1521, opened 2023-08-23
[SDK] Add pagination to getOwned (NFTs)

kien-ngo:
## Problem solved

I think the getOwned function should allow to custom counts and pagination just like how getAll() does it. 
I'm not sure if this is the most efficient way to do it. Looking to learn ^^

## Changes made

## How to test


joaquim-verges (commented):
I think this approach is fine can you add a few unit tests for it?

kien-ngo:
Yup will get on this ^

.changeset/twelve-ravens-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"@thirdweb-dev/sdk": patch
+---
+
+Add pagination option to getOwned (Typescript SDK)
```
joaquim-verges, line 5:
```suggestion
Add pagination option to `getOwned()`
```

joaquim-verges:
thanks @kien-ngo !

kien-ngo:
Hi @joaquim-verges I just pushed an additional commit to fix the error on line 89 (see attached image).
image

Just have a question: Is there any way I can learn more about the lint issue as shown in the image? Like which file is causing it or which particular lint error is it?

packages/sdk/src/core/schema/QueryParams.ts
```diff
@@ -24,3 +24,30 @@ export type QueryAllParams = Exclude<
   z.input<typeof QueryAllParamsSchema>,
   undefined
 >;
+
+export const QueryOwnedParamsSchema = /* @__PURE__ */ (() =>
```
joaquim-verges, line 28: just looking at this vs `QueryAllParams` - any reason why they're different? feels like we could re-use the same no?

Also if its a new type, should be exported in the `index.ts` in the same folder

kien-ngo: @joaquim-verges The `start` in `QueryAllParams` is meant for the tokenId. Which is useful when querying all NFTs in the collection.

However for `getOwned()` the `start` is not quite applicable since we don't know if user has the exact tokenId, so instead it should be `page` - that's the way I think about it.

I'll check the exporting part now

kien-ngo: Done exporting: https://github.com/thirdweb-dev/js/pull/1521/commits/a9ff52fd19647c44026e847e0f11539131a5819a

joaquim-verges: `start` is not necessarily tokenId, its just where you want it to start from. The logic for getAll pagination is something like:
start = 0, count = 10
start = 10, count = 10
start = 20, count = 10
etc

its annoying to then switch your logic for getOwned() to
page = 0, count = 10,
page = 1, count = 10
page = 2, count = 10
...

let's keep it consistent, i think the `start` approach is fine (just take the last `start + count`). 

annoying to have 2 pagination logic in apps

kien-ngo:
Closing this PR to open a new one directly from `@thirdweb-dev/js` instead of my fork. Also changing the code according to Joaquim's suggestion

PR #1585, opened 2023-09-05
[Local Wallet] Fix bug in loadOrCreate

adam-maj:


packages/wallets/src/evm/wallets/local-wallet.ts
```diff
@@ -380,85 +380,85 @@ export class LocalWallet extends AbstractClientWallet<
 
 type DecryptOptions =
   | {
-    decrypt?: (message: string, password: string) => Promise<string>;
-    password: string;
-  }
+      decrypt?: (message: string, password: string) => Promise<string>;
+      password: string;
+    }
```
iketw, line 385: oof, do we have different settings here?

packages/wallets/src/evm/wallets/local-wallet.ts
```diff
@@ -80,7 +80,7 @@ export class LocalWallet extends AbstractClientWallet<
    * load saved wallet data from storage or generate a new one and save it.
    */
   async loadOrCreate(options: LoadOrCreateOptions) {
-    if (await this.getSavedData()) {
+    if (await this.getSavedData(options.storage)) {
```
iketw, line 83: When do we need to pass this down here? `this.getSavedData` uses the storage passed in when creating the wallet. Are we trying to load the wallet from a different storage instance?

adam-maj: The options has a `storage` option here, so if we're that's being passed in it should be overriding usage of storage everywhere, otherwise its confusing.

furqan had passed in a storage instance into the `loadOrCreate` function which wasn't being respected. Either it shouldn't be an option for the function or the storage option should prob get respect (I fixed it for web3 api though by just passing the storage into the constructor)

packages/wallets/src/evm/wallets/local-wallet.ts
adam-maj: sadge we must

packages/wallets/src/evm/wallets/local-wallet.ts
iketw: You're totally right here. Agree with this change, let's merge it. 

The question was more for us on why do we let options being passed in in this method, but unrelated with this PR

iketw (approved):

PR #1593, opened 2023-09-07
Better UX for expired CLI tokens + Enrich auth errors for better error handling

nessup:
### CLI authentication error UX
- This PR improves CLI: when the local wallet's token has expired, CLI will now re-ask the user to authenticate

### Standard Thirdweb error object
- Additionally, this PR sets us up for a standard error object throughout the thirdweb SDK
- Services like CLI do not have a granular understanding of why there are failures in auth flows
- This PR standardizes a format for errors

The big problem with JS errors is the lack of `code`, which would allow for duck typing on errors (and improving application behavior based on specific error codes.

This introduces `ThirdwebAuthError`. When the auth SDK throws an error, you'll get an error object with new properties:
- `code`: A hardcoded string that can be used for duck typing by upstream code
- `message`: Another hardcoded string, but unlike the regular `message` supplied by `Error`, you can use substitutions like `This error happened because of ${x}` and set the value of that
- `innerError`: This is a property that can be used to store an error that was encountered while throwing this one

## Changes made

## How to test

- [ ] Intentionally send an expired token from CLI (turn back the system clock). Notice CLI now has exposure to error codes and formatted messages

### DIFF
--- a/packages/auth/src/core/errors.ts
+++ b/packages/auth/src/core/errors.ts
@@ -1,44 +1,60 @@
-import { format } from 'util';
-
-export interface ThirdwebErrorConstructorArgs<Variables = undefined> {
+export interface ThirdwebCreateErrorConstructorArgs {
   code: string;
-  message: string
-  name?: string
+  message: string;
+}
+export interface ThirdwebErrorConstructorArgs<Variables = undefined> {
   innerError?: Error;
   variables: Variables;
 }
 
-function formatString(template: string, substitutions: { [key: string]: any }): string {
+function formatString(
+  template: string,
+  substitutions: { [key: string]: any }
+): string {
   return template.replace(/\$\{(\w+)\}/g, (_match, p1) => {
     return substitutions[p1] !== undefined ? substitutions[p1] : _match;
   });
 }
 
-export function createError<Variables>(options: Omit<ThirdwebErrorConstructorArgs<Variables>, 'variables'>) {
+type OverrideOptions<Variables> = Partial<
+  ThirdwebErrorConstructorArgs<Variables>
+>;
+type OverrideOptionsWithVariables<Variables> =
+  ThirdwebErrorConstructorArgs<Variables>;
+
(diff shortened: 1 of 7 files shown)


packages/auth/src/core/errors.ts
```diff
@@ -0,0 +1,116 @@
...
+export interface ThirdwebErrorConstructorArgs<Variables = undefined> {
+  innerError?: Error;
+  variables: Variables;
+}
+
+function formatString(
+  template: string,
+  substitutions: { [key: string]: any }
+): string {
+  return template.replace(/\$\{(\w+)\}/g, (_match, p1) => {
```
adam-maj, line 15: Would be good to leave a comment on the regex for what it does for people coming back after so it's clear

packages/auth/src/core/errors.ts
```diff
@@ -0,0 +1,116 @@
...
+      if (Error.stackTraceLimit !== 0) {
+        Error.captureStackTrace(this, ThirdwebError)
+      }
+    }
+  }
+
+  return ThirdwebError;
+}
+
+export const ThirdwebAuthErrors = {
```
adam-maj, line 66: Nice this is useful

packages/cli/src/auth/index.ts
```diff
@@ -56,24 +56,30 @@ export async function loginUser(
...
-      ora().warn(chalk.yellow("You are already logged out, did you mean to login?"));
+      if (showLogs) ora().warn(chalk.yellow("You are already logged out, did you mean to login?"));
       return;
     }
     fs.unlinkSync(credsConfigPath);
     fs.unlinkSync(tokenPath);
     // TODO: We can consider not removing this on logout later, once we want to implement teams. For now this wallet will be ephemeral.
     fs.unlinkSync(cliWalletPath);
-    ora().succeed(chalk.green("You have been logged out"));
+    if (showLogs) ora().succeed(chalk.green("You have been logged out"));
```
iketw, line 72: Lint is complaining again, can you add brackets?

PR #1624, opened 2023-09-18
[SDK] [Bug] Always update network if configured in fromSigner

adam-maj:
## Problem solved

- If you pass `network` to `ThirdwebSDK.fromSigner` it should always overwrite the network of the SDK and signer.
- Previously, if you pass `network` to `fromSigner`, but the `signer` passed in already has a configured `provider`, we wouldn't overwrite the network.
- Now we properly overwrite the network here.

### DIFF
--- a/packages/sdk/src/evm/core/sdk.ts
+++ b/packages/sdk/src/evm/core/sdk.ts
@@ -186,8 +186,12 @@ export class ThirdwebSDK extends RPCConnectionHandler {
   ): ThirdwebSDK {
     let signerWithProvider = signer;
     if (network) {
-      const provider = getChainProvider(network, options);
-      signerWithProvider = signer.connect(provider);
+      try {
+        const provider = getChainProvider(network, options);
+        signerWithProvider = signer.connect(provider);
+      } catch {
+        // We have to catch here because browser wallets throw when trying to override provider
+      }
     }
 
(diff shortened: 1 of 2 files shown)


packages/sdk/src/evm/core/sdk.ts
```diff
@@ -185,7 +185,7 @@ export class ThirdwebSDK extends RPCConnectionHandler {
     storage?: ThirdwebStorage,
   ): ThirdwebSDK {
     let signerWithProvider = signer;
-    if (network && !signer.provider) {
+    if (network) {
```
joaquim-verges, line 188: Careful with this, some signers throw if you try to override the provider. That's the case for most injected browser wallets which is why this check was here. We only override provider for signers like private key / kms / backend wallets etc which don't come with providers so we need to put one in. 

Network should probably be optional for this static method tbh

adam-maj: Gotcha - `network` is already optional - but the interface is a bit misleading. We could also catch the error - that might be the best approach

adam-maj: This led to a bug in web3api where network gets passed through but doesn't get respected

joaquim-verges: Try catch should be ok

PR #1626, opened 2023-09-18
[SDK] make api match for checking and adding admin

ciaranightingale:
## Problem solved

Updated the API so that the API when reading the signers on an account match the API for setting the signers

## Changes made

- [ ] Public API changes: makeAdmin -> isAdmin for adding, removing & updating signers on accounts

### DIFF
--- a/packages/sdk/src/evm/types/account.ts
+++ b/packages/sdk/src/evm/types/account.ts
@@ -31,7 +31,7 @@ export type SignerWithPermissions = {
 export const PermissionSnapshotSchema = /* @__PURE__ */ z.array(
   /* @__PURE__ */ z.object({
     signer: AddressOrEnsSchema,
-    makeAdmin: /* @__PURE__ */ z.boolean(),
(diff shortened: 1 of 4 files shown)


nachoiacovino (approved):
lgtm!

jnsdls:
note that this technically is a public API change (breaking)

packages/sdk/src/evm/types/account.ts
```diff
@@ -31,7 +31,7 @@ export type SignerWithPermissions = {
 export const PermissionSnapshotSchema = /* @__PURE__ */ z.array(
   /* @__PURE__ */ z.object({
     signer: AddressOrEnsSchema,
-    makeAdmin: /* @__PURE__ */ z.boolean(),
+    isAdmin: /* @__PURE__ */ z.boolean(),
```
iketw, line 34: As @jnsdls mentioned this is a breaking change. What are we matching this to?

PR #1648, opened 2023-09-25
[SDK] Perf improvements with Promise.all

kien-ngo:
## Don't merge this
I'm aware this is a lot of changes. I will probably split it up into smaller PRs so that we have more control

## Problem solved

We have a lot of Promise.all that are not calling the inner requests concurrently, rather sequencially.

## Changes made

- [ ] Public API changes: list the public API changes made if any
- [x] Internal API changes: explain the internal logic changes

## How to test

- [x] Automated tests: link to unit test file
- [ ] Manual tests: step by step instructions on how to test

Self note: The following Pormise.all in this file can be improved however it seems too complex dont wanna touch that rn
- [ ] packages\sdk\src\evm\core\classes\drop-erc1155-claim-conditions.ts
- [ ] packages\sdk\src\evm\common\claim-conditions\processClaimConditionInputs.ts
- [ ] packages\sdk\src\evm\common\any-evm-utils\computeDeploymentInfo.ts

PR #1678, opened 2023-09-28
[SDK] Performance improvement to various function for ERC721

kien-ngo:
## Problem solved

This is a subset of the changes from the draft PR below. Since https://github.com/thirdweb-dev/js/pull/1648 contains too many changes I will split it up into smaller PRs
## Changes made

## How to test


joaquim-verges (approved):
these are great, one thing i recommend is just to add a console.time() for some of these to see the before/after, just while you dev to see the delta. wonder how much we're saving here!

PR #1692, opened 2023-10-01
refactor(react): OTP Input Readability Enhancements 

etnlbck:
## Problem solved
Made readability enhancements to simplify reading.

## Changes made
Line 39 Prefer using nullish coalescing operator (`??`) instead of a ternary expression, as it is simpler to read.
Line 47, 100 Use concise character class syntax '\d' instead of '[0-9]'.


packages/react/src/components/OTPInput.tsx
```diff
@@ -36,15 +36,15 @@ export function OTPInput(props: {
             data-error={props.isInvalid}
             ref={(e) => (boxEls.current[i] = e)}
             key={i}
-            value={otp[i] === undefined ? "" : otp[i]}
+            value={otp[i] ?? ""}
```
etnlbck, line 39: Prefer using nullish coalescing operator (`??`) instead of a ternary expression, as it is simpler to read.

packages/react/src/components/OTPInput.tsx
```diff
@@ -36,15 +36,15 @@ export function OTPInput(props: {
...
             type="number"
             variant="outline"
             onPaste={(e) => {
               const pastedData = e.clipboardData.getData("text/plain");
               const newOTP = pastedData
                 .slice(0, props.digits)
                 .split("")
-                .filter((n) => /[0-9]/.test(n))
+                .filter((n) => /\d/.test(n))
```
etnlbck, line 47: Use concise character class syntax '\d' instead of '[0-9]'.

packages/react/src/components/OTPInput.tsx
```diff
@@ -97,7 +97,7 @@ export function OTPInput(props: {
                 value = value[value.length - 1];
               }
 
-              if (!/[0-9]/.test(value) && value !== "") {
+              if (!/\d/.test(value) && value !== "") {
```
etnlbck, line 100: Use concise character class syntax '\d' instead of '[0-9]'.

PR #1730, opened 2023-10-08
[Wallets] Add wallet analytics

joaquim-verges:
## Problem solved

Enable wallet analytics for app developers

## Changes made

- [ ] Public API changes: `analytics` can be explicitely enabled/disabled on each wallet constructor
- [ ] Internal API changes: fire and forget analytics call on connection

## How to test


packages/wallets/src/evm/utils/analytics.ts
```diff
@@ -0,0 +1,28 @@
+import { fetch } from "cross-fetch";
+
+const ANALYTICS_ENDPOINT = "https://c.thirdweb-dev.com/event";
```
joaquim-verges, line 3: TODO change this to prod

packages/wallets/src/evm/utils/analytics.ts
```diff
@@ -0,0 +1,28 @@
+import { fetch } from "cross-fetch";
```
jnsdls, line 1: don't need `cross-fetch` anymore we no longer support any environment that doesn't have native fetch

PR #1746, opened 2023-10-10
RPC Override Fix

farhanW3:
## Problem solved

RPC Override logic on SDK was not using the user provided RPC Url.

## Changes made

- [ ] Sync the order of data in `supportedChains` [], i.e., [`provided_network_data`, `user_rpc_override_data`]. It was mixed earlier.
- [ ] Updated the `filter` to have `===` instead of `!==`

## How to test

- [ ] Run `hardhat` locally using docker. Change the port to `8000` instead of `8545`
- [ ] Pass in `supportedChain` data with the new `rpc` Url, i.e., `http://localhost:8000`
- [ ] Call `sdk.wallet.getBalance()` -> This should get you the balance

Supported Chain (RPC Override) Data:

```
[
  {
    "name": "Localhost",
    "chain": "ETH",
    "rpc": ["http://localhost:8000"],
    "nativeCurrency": {
      "name": "Ether",
      "symbol": "ETH",
      "decimals": 18
    },
    "chainId": 1337,
    "shortName": "local",
    "slug": "localhost"
  }
]
```

### DIFF
--- a/packages/sdk/src/evm/constants/urls.ts
+++ b/packages/sdk/src/evm/constants/urls.ts
@@ -44,7 +44,7 @@ export function getChainProvider(
     options.supportedChains = [
       // @ts-expect-error - we know this is a chain and it will work to build the map
       network,
-      ...options.supportedChains.filter((c) => c.chainId !== network.chainId),
+      ...options.supportedChains.filter((c) => c.chainId === network.chainId),
     ];
   }
 
--- a/.changeset/pretty-toys-yawn.md
+++ b/.changeset/pretty-toys-yawn.md
@@ -1,0 +1,5 @@
+---
+"@thirdweb-dev/sdk": patch
+---
+
+Fix for RPC Override Implementation to get user provided Url and not default RPC Url for the chain
--- a/packages/sdk/src/evm/core/sdk.ts
+++ b/packages/sdk/src/evm/core/sdk.ts
@@ -724,13 +724,10 @@ export class ThirdwebSDK extends RPCConnectionHandler {
       walletAddress,
     );
 
-    const chainMap = chains.reduce(
-      (acc, chain) => {
-        acc[chain.chainId] = chain;
-        return acc;
-      },
-      {} as Record<number, Chain>,
-    );
+    const chainMap = chains.reduce((acc, chain) => {
+      acc[chain.chainId] = chain;
+      return acc;
+    }, {} as Record<number, Chain>);
 
     const sdkMap: Record<number, ThirdwebSDK> = {};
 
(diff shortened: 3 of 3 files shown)


packages/sdk/src/evm/constants/urls.ts
```diff
@@ -44,7 +44,7 @@ export function getChainProvider(
     options.supportedChains = [
       // @ts-expect-error - we know this is a chain and it will work to build the map
       network,
-      ...options.supportedChains.filter((c) => c.chainId !== network.chainId),
+      ...options.supportedChains.filter((c) => c.chainId === network.chainId),
```
joaquim-verges, line 47: This change doesn't make sense to me. Can you explain the logic?

farhanW3: with `!==` the `supportedChains ` property was not having the user provided RPC Override information and thus not working.

Changing this logic to be in-sync with [this](https://github.com/thirdweb-dev/js/pull/1746/files#diff-3c93eef5707ade0c30c25d6dd3ffd8bf75095e1d4d761decf5d9b4b7566607bbR890), thus updating it `===` gets the `supportChains` property value to be same, which is then used in the `buildDefaultMap()`

`buildDefaultMap()` implementation

```
function buildDefaultMap(options: SDKOptionsOutput) {
  return options.supportedChains.reduce(
    (previousValue, currentValue) => {
      previousValue[currentValue.chainId] = currentValue;
      return previousValue;
    },
    {} as Record<number, ChainInfo>,
  );
}
```

Since there was just one value (object) in `supportedChains` property, the above was just using the SDK provided value and not the user provided one.

joaquim-verges: I think the root cause is that we're using an array here and then reducing it as an object so any duplicates will get overriden.

I think this will work for now since the SDK assumes a single chain ATM, so this change is ok, but this logic is still fragile.

Will look into a proper fix for this

joaquim-verges: Your case is special since you're passing a network slug + custom supported chains, usually you would pass your custom chain directly as the network.

Let's merge this to unblock you and I'll do another pass at it

farhanW3: Cool thanks

joaquim-verges: Needs some unit tests badly

farhanW3: Cool.

So can I merge this? or you will do a proper fix and then merge it?

Please let me know. Will need to update Engine with the new SDK for it to work :)

farhanW3: okies. thank you will merge. 

yeah need UT

farhanW3: > Your case is special since you're passing a network slug + custom supported chains, usually you would pass your custom chain directly as the network.

Ahh got it. Maybe I can update it on Engine side then and we do not need to touch SDK..

PR #1809, opened 2023-10-19
[SDK] Fix publish

kumaryash90:
## Problem solved

Cleans up values related to old publish flow, whenever republishing an old contract.

For e.g. `isDeployableViaFactory` or `isDeployableViaFactory` have been deprecated, and exist only for backward compatibility. So these should be set to false for all new publishes.


packages/sdk/src/evm/core/classes/contract-publisher.ts
```diff
@@ -541,4 +545,19 @@ export class ContractPublisher extends RPCConnectionHandler {
...
+    extraMetadata: ExtraPublishMetadata,
+  ): ExtraPublishMetadata {
+    return {
+      ...extraMetadata,
+      isDeployableViaFactory: false,
+      isDeployableViaProxy: false,
+      factoryDeploymentData: {
+        ...extraMetadata.factoryDeploymentData,
+        implementationAddresses: {},
+        factoryAddresses: {},
```
joaquim-verges, line 559: we def don't use either of these?

kumaryash90: We use these when deploying, if these exist.

But the new publish relies on `deployType: autoFactory` etc, and doesn't have implementation/factory addresses because of deterministic deployments.

So, if an old published contract has not been republished, it will still get deployed. But, once republished, all the old fields will be reset in favour of new publish flow.

joaquim-verges:
@kumaryash90 can we merge this? if not lets close

PR #1821, opened 2023-10-20
[SmartWallet] implement multidimensional nonces for smart wallets

joaquim-verges:
## Problem solved

You can now chain send transactions without waiting for confirmations on smart wallets

## Changes made

## How to test

### DIFF
--- a/.changeset/famous-penguins-lay.md
+++ b/.changeset/famous-penguins-lay.md
@@ -1,0 +1,5 @@
+---
+"@thirdweb-dev/wallets": patch
+---
+
+Multidimensional nonces for smart wallets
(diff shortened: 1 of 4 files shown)


3Vis3:
The multidimensional nonce breaks compatibility reaching RPC endpoints using custom Paymaster/Bundlers, such as the [Particle Bundler](https://docs.particle.network/developers/account-abstraction/rpc/bundler-rpc).
I'd suggest using a sequential nonce inside sendTransaction if using a custom PaymasterAPI:

NOW:
`const multidimensionalNonce = randomNonce();`

FIX:
`const nonce = this.smartAccountAPI.getNonce();`

Let me know if it's all clear.

joaquim-verges:
@3Vis3 interesting, what error do you get from the particle bundler?

Multidimensional nonces are an important feature so that userOps can be send in quick succession without waiting for the bundler to submit the op onchain.

I get your fix but wonder if instead we should allow to control the behavior a bit more explicitly.

Feel free to put up a PR and we can discuss implementations there

PR #1852, opened 2023-10-26
update# Official brand upgrade, changed from opside to lumoz

cppfuns:
Official brand upgrade, changed from opside to lumoz

### DIFF
--- a/packages/chains/chains/51178.ts
+++ b/packages/chains/chains/51178.ts
@@ -4,8 +4,8 @@ export default {
   "chainId": 51178,
   "explorers": [
     {
-      "name": "OpsideTestnetInfo",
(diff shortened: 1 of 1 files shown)


jnsdls:
hi there @cppfuns.

This change sadly won't work here as this is generated code that you're manually updating and as such will get overwritten.

To make the change you're trying to make please do one of these 2 things:

1. Make the changes in a PR to: https://github.com/ethereum-lists/chains -> this will automatically then get updated in a subsequent build on our side.
2. Alternatively go here: https://thirdweb.com/solutions/chains and fill out the form -> we will then take the information you provide there and make the update in our system.

Thanks!

I'll be closing this as this code change won't actually work, please feel free to raise an issue if you have questions!

PR #1861, opened 2023-10-27
Initialize `@thirdweb-dev/crypto` and `@thirdweb-dev/merkletree`

jnsdls:
Required testing:
- local wallet on `node` and `browser`
- merkletree (allowlists for drops etc)

#### `/crypto`
- contains shared crypto functions that work across node + browsers
- goal is stability & speed
- 
#### `/merkletree`
- contains an initial optimized merkletree implementation

image

☝️ only leftover reference is a dev dependency that is used to verify the output of our `@thirdweb-dev/crypto` library is identical with `crypto-js`


packages/crypto/benchmark/keccak256.ts
```diff
@@ -0,0 +1,63 @@
...
+  .add("ethers@5", async () => {
+    // OK because this is benchmark code only
+    // eslint-disable-next-line no-restricted-globals
+    utils.keccak256(Buffer.from(stringValue));
+  })
+  .add("@noble/hashes", async () => {
+    uint8ArrayToHex(noble(stringValue));
+  })
+  .add("viem", async () => {
+    viem(stringValue as `0x${string}`);
```
jxom, line 55: This bench is kind of redundant as viem uses `@noble/hashes` (above bench) internally.

packages/crypto/README.md
```diff
@@ -0,0 +1,94 @@
...
+#### keccack256 -> Uint8Array
+
+```bash
+┌──────────────────────┬─────────┬────────────────────┬────────┬─────────┐
+│      Task Name       │ ops/sec │ Average Time (ns)  │ Margin │ Samples │
+├──────────────────────┼─────────┼────────────────────┼────────┼─────────┤
+│ @thirdweb-dev/crypto │ 72,533  │ 13786.719900000808 │ ±0.31% │ 100000  │
+│ ethers@v5            │ 71,891  │ 13909.94663000107  │ ±0.33% │ 100000  │
+│ @noble/hashes        │ 27,244  │ 36704.73226000136  │ ±0.19% │ 100000  │
+│ viem                 │ 27,115  │ 36879.07953999698  │ ±0.26% │ 100000  │
```
jxom, line 79: FYI, `@noble/hashes` can be faster, but it affects [auditability](https://github.com/paulmillr/noble-hashes/blob/main/benchmark/README.md).

packages/crypto/benchmark/keccak256.ts
jnsdls: this is fair, these benchmarks were more for internal communication and are very basic at the moment - I'll likely remove them entirely from the README because while we care about the performance of this stuff it is not the first thing we care about here and as such it shouldn't be the first big section in the README either

packages/crypto/README.md
jnsdls: Again fair on the benchmarks. Similar to above.

While audibility is laudable and wanted, it coming at such a significant price of performance (even with such basic benchmarks) is currently not acceptable.

All of our users are currently using js-sha3 under the hood via ethers@5 so it is acceptable to go with the same solution for the moment (but make it easier for us to switch it out / have that control) because it's no longer wrapped 3 levels down. The end goal is not to stick with it for the long run. :)

paulmillr: > significant price of performance

How many sha3 executions per second is "acceptable" for you? I don't think it matters for 99% of client apps AT ALL!

paulmillr: I've published `unrolled-nbl-hashes-sha3` 1 year ago, which is the same stuff from noble but with shitty auditability and it runs around js-sha3 and others. What for? It's useless.

jxom: Feel like the perf gain may _look_ large here, but hashing computation in the context of web3 libraries is generally constrained by other upstream dependant computation (ie. serializing, secp256k1, coding, network, etc). Hence, why I think they are misleading.

Issue #1868, opened 2023-10-29
useSwitchChain does not switch the activeChain for ThirdwebProvider


tommaay:
## Background
This issue is causing the `useContract` hook to be unusable. If I go from `Ethereum` chain to Polygon and pass a Polygon contract to `useContract`, the hook will fail because the `activeChain` in `ThirdwebProvider` is still `Ethereum`.  `activeChain` is currently passed as props to `ThirdwebProvider` and does not change.  

## Proposal
Set a state in `ThirdwebWalletProvider` that takes the `activeChain` value from props as the initial value. 
- When `useSwitchChain` is used to switch the chain, set the `activeChain` state to the chain that was called with `useSwitchChain` or a create a new hook like `useSwitchActiveChainForProvider`.

kien-ngo:
cc. @MananTank @joaquim-verges

I think it's a good idea to have a `useSwitchActiveChain` or `useSwitchChainProvider` so that users can conveniently navigate between networks in a multichain application.

tommaay:
I created a draft pr to show what it might look like: [pr](https://github.com/thirdweb-dev/js/pull/1869)

mertcankose:
> I created a draft pr to show what it might look like: [pr](https://github.com/thirdweb-dev/js/pull/1869)

same issue! I am waiting to approve your commit as soon as possible

tommaay:
> > I created a draft pr to show what it might look like: [pr](https://github.com/thirdweb-dev/js/pull/1869)
> 
> same issue! I am waiting to approve your commit as soon as possible

They won't be exposing a hook to handle the activeChain switch so we will have to do that in our apps

PR #1869, opened 2023-10-29
Set activeChain in Provider state and created useSwitchActiveChain hook

tommaay:
## Problem solved
- Added `const [activeChain, setActiveChain] = useState<Chain>(props.activeChain)` to `ThirdwebWalletProvider` and consume the activeChain value from local state instead of `props.activeChain` 
- Created `useSwitchActiveChain` to be able to switch the Provider `activeChain`

## Changes made

## How to test
Using the pr to publish an npm package to test the chanes

#1868

### DIFF
--- a/packages/react-core/src/core/hooks/wallet-hooks.ts
+++ b/packages/react-core/src/core/hooks/wallet-hooks.ts
@@ -187,6 +187,20 @@ export function useSwitchChain() {
 
 /**
  *
+ * @returns a method to connect the wallet to network/chain with given chainId and also change the activeChain in the <ThirdwebProvider/>
+ *
+ */
+export function useSwitchActiveChainForProvider() {
+  const context = useWalletContext();
+  invariant(
+    context,
+    "useSwitchActiveChainForProvider() must be used within a <ThirdwebProvider/>",
+  );
+  return context.switchActiveChainForProvider;
+}
+
+/**
+ *
  * @returns a method to set a connected wallet instance
  */
 export function useSetConnectedWallet() {
--- a/packages/react-core/src/core/providers/thirdweb-wallet-provider.tsx
+++ b/packages/react-core/src/core/providers/thirdweb-wallet-provider.tsx
@@ -92,6 +92,7 @@ type ThirdwebWalletContextData = {
   getWalletConfig: (walletInstance: WalletInstance) => WalletConfig | undefined;
   activeChainSetExplicitly: boolean;
   clientId?: string;
+  switchActiveChainForProvider: (chain: number) => Promise<void>;
 };
 
 const ThirdwebWalletContext = /* @__PURE__ */ createContext<
@@ -116,6 +117,8 @@ export function ThirdwebWalletProvider(
   const [signer, setSigner] = useState<Signer | undefined>(undefined);
   const [connectionStatus, setConnectionStatus] =
     useState<ConnectionStatus>("unknown");
+
+  const [activeChain, setActiveChain] = useState<Chain>(props.activeChain);
 
   const autoConnectTimeout = props.autoConnectTimeout || 15000;
 
@@ -261,6 +264,14 @@ export function ThirdwebWalletProvider(
       setSigner(_signer);
     },
     [activeWallet, storeLastActiveChainId],
+  );
+
+  const switchActiveChainForProvider = useCallback(
+    async (chainId: number) => {
+      await switchChain(chainId);
+      setActiveChain(chainId);
+    },
+    [switchChain],
   );
 
   const autoConnectTriggered = useRef(false);
@@ -513,13 +524,14 @@ export function ThirdwebWalletProvider(
         createWalletStorage: props.createWalletStorage,
         switchChain,
         setConnectedWallet: setConnectedWallet,
-        activeChain: props.activeChain,
+        activeChain: activeChain,
         chainToConnect,
         getWalletConfig: (walletInstance: WalletInstance) => {
           return walletInstanceToConfig.get(walletInstance);
         },
         activeChainSetExplicitly: props.activeChainSetExplicitly,
         clientId: props.clientId,
+        switchActiveChainForProvider,
       }}
     >
       {props.children}


joaquim-verges:
Thank you for this @tommaay ! I do agree with @kien-ngo, its a good feature to have.

We'll be looking at this first thing on Monday. cc @MananTank @iketw

MananTank:
@tommaay Thanks for the PR! 

However, there are a few issues with this PR

* `activeChain` is only updated inside the `ThirdwebWalletContext` - but `activeChain` is not updated in `ThirdwebSDKProviderWrapper` context. `activeChain` must be updated at the root context to update the `activeChain` everywhere

* creating a `activeChain` state from the `activeChain` prop passed from `ThirdwebProvider` makes it "disconnected" from the prop - if the `activeChain` prop is updated on `ThirdwebProvider`, it will NOT update the `activeChain` state. To fix this, you would to create an effect that updates the state when the `activeChain` prop is updated, and there few more considerations like memoizing the prop to avoid treating the "same" prop object as different on consecutive re-renders.

This is unnecessarily complicated, the simple solution for changing the activeChain is to just wrap the ThirdwebProvider in a component and put a state above it 

```jsx
import { type Chain, Mumbai } from "@thirdweb-dev/chains"

function AppWithProviders() {
  const [activeChain, setActiveChain] = useState<Chain>(Mumbai);
  return <ThirdwebProvider activeChain={activeChain}> <App /> </ThirdwebProvider>
}

```

tommaay:
> @tommaay Thanks for the PR!
> 
> However, there are a few issues with this PR
> 
> * `activeChain` is only updated inside the `ThirdwebWalletContext` - but `activeChain` is not updated in `ThirdwebSDKProviderWrapper` context. `activeChain` must be updated at the root context to update the `activeChain` everywhere
> * creating a `activeChain` state from the `activeChain` prop passed from `ThirdwebProvider` makes it "disconnected" from the prop - if the `activeChain` prop is updated on `ThirdwebProvider`, it will NOT update the `activeChain` state. To fix this, you would to create an effect that updates the state when the `activeChain` prop is updated, and there few more considerations like memoizing the prop to avoid treating the "same" prop object as different on consecutive re-renders.
> 
> This is unnecessarily complicated, the simple solution for changing the activeChain is to just wrap the ThirdwebProvider in a component and put a state above it
> 
> ```js
> import { type Chain, Mumbai } from "@thirdweb-dev/chains"
> 
> function AppWithProviders() {
>   const [activeChain, setActiveChain] = useState<Chain>(Mumbai);
>   return <ThirdwebProvider activeChain={activeChain}> <App /> </ThirdwebProvider>
> }
> ```

I understand that but what is the point of exposing a useSwitchChain hook if it's not going to switch the activeChain? Create a local state at the root level or a context provider just to handle this scenario seems like an anti pattern when it should be handled already.

MananTank:
@tommaay - `useSwitchChain` is for changing the network/chain in user's wallet. 

You would use it when the user's wallet is not connected to the `activeChain` set on the `ThirdwebProvider` 

`activeChain` declares what chain your app works on, `useSwitchChain` helps you to get the user's wallet on that chain

tommaay:
> @tommaay - `useSwitchChain` is for changing the network/chain in user's wallet.
> 
> You would use it when the user's wallet is not connected to the `activeChain` set on the `ThirdwebProvider`
> 
> `activeChain` declares what chain your app works on, `useSwitchChain` helps you to get the user's wallet on that chain

There should also be the ability to switch the `activeChain` for the app. Our app has contracts on different chains and we want to be able to create a contract instance with `useContract` for the `activeChain`. Sure there are hacky ways to do that but this is not a good developer experience. The team already has another solution using Privy and I thought I could get them to switch to Thirdweb but I don't think I can recommend that with this approach.

MananTank:
Hey @tommaay - This is not a hacky workaround, The example I suggested does exactly what you want with your proposed `useSwitchActiveChain` hook. Let me know if I'm missing something here

tommaay:
> Hey @tommaay - This is not a hacky workaround, The example I suggested does exactly what you want with your proposed `useSwitchActiveChain` hook. Let me know if I'm missing something here

Sorry I meant to say not intuitive because of the `useSwitchChain` hook.

tommaay:
@MananTank Are you familiar with Remix apps? I am trying to use the context the root file but it's not working for some reason. 

`Error: Error: useWalletContext() can only be used inside <ThirdwebProvider />`


```
import { createContext, useContext, useState } from "react";
import {
  ThirdwebProvider,
  coinbaseWallet,
  embeddedWallet,
  metamaskWallet,
  rainbowWallet,
  walletConnect,
  zerionWallet,
  useSwitchChain,
} from "@thirdweb-dev/react";

const defaultChain = 1;

export const ActiveChainContext = createContext<{
  activeChain: number;
  switchActiveChain: (chain: number) => Promise<void>;
}>({ activeChain: defaultChain, switchActiveChain: () => Promise.resolve() });

export const ActiveChainProvider: React.FC<{ children: React.ReactNode; clientId: string }> = ({
  children,
  clientId,
}) => {
  const [activeChain, setActiveChain] = useState<number>(defaultChain);

  const switchChain = useSwitchChain();

  const switchActiveChain = async (chain: number) => {
    await switchChain(chain);
    setActiveChain(chain);
  };

  return (
    <ActiveChainContext.Provider
      value={{
        activeChain,
        switchActiveChain,
      }}
    >
      <ThirdwebProvider
        activeChain={activeChain}
        clientId={clientId}
        supportedWallets={[
          metamaskWallet(),
          coinbaseWallet(),
          rainbowWallet(),
          walletConnect(),
          zerionWallet(),
          embeddedWallet(),
        ]}
      >
        {children}
      </ThirdwebProvider>
    </ActiveChainContext.Provider>
  );
};

export const useActiveChain = () => {
  const context = useContext(ActiveChainContext);
  if (context === undefined) {
    throw new Error("useActiveChain must be used within a ActiveChainProvider");
  }
  return context;
};
```

`root.tsx`
```
export async function loader() {
  return {
    ENV: {
      THIRDWEB_CLIENT_ID: process.env.THIRDWEB_CLIENT_ID,
    },
  };
}

export default function App() {
  const { ENV } = useLoaderData<typeof loader>();

  return (
    <html lang="en">
      <head>
        <meta charSet="utf-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1" />
        <Meta />
        <Links />
      </head>
      <body>
        <ActiveChainProvider clientId={ENV.THIRDWEB_CLIENT_ID!}>
          <div className="p-2 flex justify-end">
            <ConnectWallet
              switchToActiveChain={true}
              className="hover:bg-brand bg-white transition-all ease-in-out duration-200 text-sm hover:text-white rounded-full px-4 py-2"
            />
          </div>
        </ActiveChainProvider>
        <Outlet />
        <ScrollRestoration />
        <Scripts />
        <LiveReload />
      </body>
    </html>
  );
}
```

tommaay:
@MananTank Never mind I got it to work

MananTank:
@tommaay 

After discussing it with the team, we think a built-in hook is more convenient for changing the `activeChain` on `ThirdwebProvider`. We will add the necessary code changes to cover all the edge cases soon.

tommaay:
> @tommaay
> 
> After discussing it with the team, we think a built-in hook is more convenient for changing the `activeChain` on `ThirdwebProvider`. We will add the necessary code changes to cover all the edge cases soon.

Sweet thanks @MananTank!

PR #1871, opened 2023-10-30
[SDK] Reduce bundle size by lazy loading some large modules

kien-ngo:
## Problem solved

-51kb (14% reduction)

## Changes made

## How to test


packages/sdk/src/evm/core/sdk.ts
```diff
@@ -9,7 +9,7 @@ import {
   PREBUILT_CONTRACTS_MAP,
   getContractTypeForRemoteName,
 } from "../contracts";
-import { SmartContract } from "../contracts/smart-contract";
```
joaquim-verges, line 12: dont think you need to do an alias here for the type. since the real import below is dynamic it should not clash

kien-ngo: got it, will update

kien-ngo: @joaquim-verges Lint issue if they have the same name :(
image

joaquim-verges: Hm ok my bad then!

PR #1885, opened 2023-11-01
[SmartWallet] Improved `useSmartWallet` hook

joaquim-verges:
## Problem solved

New API for conveniently connecting smart wallets without UI components:

Example with metamask:

```tsx
const { connect } = useSmartWallet(metamaskWallet(), {
    factoryAddress: factoryAddress,
    gasless: true,
});

const onClick = async () => {
    // nothing to do here, all handled internally
    await connect(); 
}
```

Example with localWallet:

```tsx
const { connect } = useSmartWallet(localWallet(), {
    factoryAddress: factoryAddress,
    gasless: true,
});

const onClick = async () => {
    // function to 'load' the local wallet before using it
    await connect({
        connectPersonalWallet: async (w) => {
            await w.generate();
            await w.connect();
        }
    });
}
```

Example with embeddedWallet:

```tsx
const { connect } = useSmartWallet(embeddedWallet(), {
    factoryAddress: factoryAddress,
    gasless: true,
});

const onClick = async () => {
    // function to 'auth' the embedded wallet before using it
    await connect({
        connectPersonalWallet: async (w) => {
            const authResult = await w.authenticate({ strategy: "google" });
            await w.connect({ authResult });
        }
    });
}
```

## Changes made

## How to test

### DIFF
--- a/packages/react-native/src/evm/wallets/connectors/embedded-wallet/embedded-connector.ts
+++ b/packages/react-native/src/evm/wallets/connectors/embedded-wallet/embedded-connector.ts
@@ -59,6 +59,9 @@ export class EmbeddedWalletConnector extends Connector<EmbeddedWalletConnectionA
       await this.switchChain(options.chainId);
     }
 
+    if (this.email) {
+      saveConnectedEmail(this.email);
+    }
     return this.getAddress();
   }
 
@@ -144,7 +147,6 @@ export class EmbeddedWalletConnector extends Connector<EmbeddedWalletConnectionA
     email: string;
   }): Promise<SendEmailOtpReturnType> {
     this.email = options.email;
-    saveConnectedEmail(options.email);
     return sendVerificationEmail({
       email: options.email,
       clientId: this.options.clientId,
--- a/packages/react-native/src/evm/wallets/connectors/embedded-wallet/embedded-connector.ts
+++ b/packages/react-native/src/evm/wallets/connectors/embedded-wallet/embedded-connector.ts
@@ -67,6 +67,7 @@ export class EmbeddedWalletConnector extends Connector<EmbeddedWalletConnectionA
     switch (strategy) {
       case "email_verification": {
         return await this.validateEmailOTP({
+          email: params.email,
           otp: params.verificationCode,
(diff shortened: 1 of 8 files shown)


packages/react-native/src/evm/wallets/hooks/useSmartWallet.ts
```diff
@@ -0,0 +1,99 @@
...
+        clientId: context.clientId,
+      });
+    },
+    [context.clientId],
+  );
+
+  return {
+    connect,
+    predictAddress,
+    sendVerificationEmail,
```
joaquim-verges, line 97: hmmm its a bit weird to have this here since its very embeddedWallet specific... I feel like in that case its ok to let ppl grab this from `useEmbeddedWallet()` or the static method directly no?

packages/react-native/src/evm/wallets/connectors/embedded-wallet/embedded-connector.ts
```diff
@@ -59,6 +59,9 @@ export class EmbeddedWalletConnector extends Connector<EmbeddedWalletConnectionA
       await this.switchChain(options.chainId);
     }
 
+    if (this.email) {
+      saveConnectedEmail(this.email);
```
joaquim-verges, line 63: was this needed?

iketw: I was doing it on sendverification but at that point the wallet is not connected so doing it after a successful connection.

I'm saving the email here because on auto-connection the email is not in the connect params and we want to show it in the UI

packages/react-native/src/evm/wallets/hooks/useSmartWallet.ts
iketw: oof, yeah, I was testing it and mechanically added it, my bad.

packages/react-native/src/evm/wallets/connectors/embedded-wallet/embedded-connector.ts
joaquim-verges: sg

PR #1887, opened 2023-11-01
[SDK] Add pagination to ERC721's `getAllOwners` method

kien-ngo:
## Problem solved

Currently this method has no limit on how many items one can query, which create excessive RPC calls. Adding this option should provide users some control over their RPC usage, and should potentially improve the performance of the `useOwnedNFTs` hook

## Changes made

## How to test


packages/sdk/src/evm/core/classes/erc-721.ts
```diff
@@ -435,7 +435,7 @@ export class Erc721<
     } else {
       const [address, allOwners] = await Promise.all([
         walletAddress || this.contractWrapper.getSignerAddress(),
-        this.getAllOwners(),
+        this.getAllOwners(queryParams),
```
joaquim-verges, line 438: you're already applying the query params below, I don't think its expected to paginate both all owners and then the tokens of the given wallet. it will break your pagination logic - in fact this should have broken your tests

kien-ngo: checking, ty

kien-ngo: Hi @joaquim-verges sorry but I don't see what I'm doing wrong here. 
Basically `getAllOwner` and the tokenIds array are targeting the same set of items filtered by the queryParams.

So on L438 I'm fetching the owners for, let's say, index 0 -> 10, and then `ownedTokens` will also be filtered so that it contains only NFTs from index 0 - 10. Finally we can merge the owners into the nft metadata and they should perfectly match.

I have manually done the test on a collection so not sure what I'm missing here

kien-ngo:
Just wanna credit a community member for pointing out what I couldn't see in my code 💀 
image

packages/sdk/src/evm/core/classes/erc-721.ts
kien-ngo: Abort! my brain stopped functioning for a while

jnsdls (approved):
LGTM

PR #1895, opened 2023-11-02
[SDK] Enable proxy deployments for account factories

kumaryash90:
## Problem solved

Enable proxy deployments for account factories


packages/sdk/src/evm/common/any-evm-utils/computeDeploymentInfo.ts
```diff
@@ -223,6 +224,21 @@ export async function encodeConstructorParamsForImplementation(
...
+        const extensions = constructorParamMap
+          ? constructorParamMap["_extensions"].value
+          : [];
+
+        return extensions;
+      } else if (p.name && p.name.includes("entrypoint")) {
+        const chainId = (await provider.getNetwork()).chainId;
+        const entrypoint = getEntrypointByChainId(chainId);
+
+        if (entrypoint === constants.AddressZero) {
```
joaquim-verges, line 237: this is fine for now,

in the next iteration of this would be nice to publish the Entrypoint contract so we can deploy one ourselves when needed

joaquim-verges (approved):

PR #1908, opened 2023-11-08
[React Native] Option to hide [Switch to Personal Wallet] button

kien-ngo:
## Problem solved

## Changes made

## How to test


packages/react-native/src/evm/components/ConnectWallet.tsx
```diff
@@ -69,6 +69,13 @@ export type ConnectWalletProps = {
    * ```
    */
   displayBalanceToken?: Record<number, string>;
+
+  /**
+   * Hide the "switch to Personal wallet" option in the wallet modal which is shown when wallet is connected to either Smart Wallet or Safe
```
iketw, line 74: Please, don't mention Safe in the comments here since it's not available in RN :)

packages/react-native/src/evm/components/ConnectWalletDetails/WalletDetailsButton.tsx
```diff
@@ -36,6 +36,12 @@ export type ConnectWalletDetailsProps = {
    * ```
    */
   displayBalanceToken?: Record<number, string>;
+  /**
+   * Hide the "switch to Personal wallet" option in the dropdown which is shown when wallet is connected to either Smart Wallet or Safe
```
iketw, line 40: same comment about Safe

iketw (commented):
Thanks again for adding this! I'll get your branch and test this out in RN

packages/react-native/src/evm/components/ConnectWalletDetails/WalletDetailsButton.tsx
kien-ngo: Ty let me update it

iketw:
@kien-ngo tested this in RN and it's working as expected. To be honest, I'm not sure this should live in the `ConnectWallet` component but in the `smartWallet` config and then inject it down to the UI. We can merge this for now though

PR #1918, opened 2023-11-09
[CLI] fixed url redirect for pre-built deploy command

saminacodes:
## Problem solved

Fixed deploy pre-built URL leading to the wrong dashboard

## Changes made

- changed URL to /thirdweb.eth/ instead of /dashboard/


iketw (approved):
Thanks for fixing this @saminacodes !! Tested on my end and it correctly opens the deploy page :)

PR #1935, opened 2023-11-13
[SDK] Resolve diamond pattern facets abi

kumaryash90:
## Problem solved

Fetch and merge ABIs for facets (extensions) of diamond pattern contracts


packages/sdk/src/evm/common/plugin/getCompositePluginABI.ts
```diff
@@ -68,6 +68,22 @@ export async function getCompositeABI(
 
       // get ABIs of extension contracts
       pluginABIs = await getPluginABI(plugins as any[], provider, storage);
+    } else {
+      // check if the contract is diamond pattern
+      const dimaondAbi = [
+        "function facets() external view returns (tuple(address,bytes4[])[])",
```
joaquim-verges, line 74: we should be detecting this feature instead of always hitting the RPC to check this. Otherwise you're adding an RPC call to every getContract()

kumaryash90: The diamond entrypoint doesn't have anything specific besides the fallback function. So tried this way to test it.

However, looking into directly reading storage slot like we do for resolving proxies.

joaquim-verges (approved):
im ok with this - lmk when you merge, we'll need to update the contracts api with it

PR #1951, opened 2023-11-15
[React-core] Fix caching issue with useTotalCirculatingSupply

kien-ngo:
## Problem solved
The code below doesn't work beecause currently they all share the same cacheKey, so as the result all of them will return the cached data from the first `useTotal...` (tokenId #0)

This fix adds the `tokenId` as the extra cacheKey
```
const { contract: nftContract } = useContract(
  "0x8a41F6......9829659526"
);
const { data: token0 } = useTotalCirculatingSupply(nftContract, 0);
const { data: token1 } = useTotalCirculatingSupply(nftContract, 1);
const { data: token2 } = useTotalCirculatingSupply(nftContract, 2);
const { data: token3 } = useTotalCirculatingSupply(nftContract, 3);

console.log({
  token0: token0?.toString(),
  token1: token1?.toString(),
  token2: token2?.toString(),
  token3: token3?.toString(),
});
```

## Changes made

## How to test

### DIFF
--- a/packages/react-core/src/evm/utils/cache-keys.ts
+++ b/packages/react-core/src/evm/utils/cache-keys.ts
@@ -182,12 +182,12 @@ export const cacheKeys = {
           ),
         totalCirculatingSupply: (
           contractAddress: RequiredParam<ContractAddress>,
-          tokenId?: RequiredParam<BigNumberish>,
+          tokenId?: BigNumberish,
         ) =>
           createContractCacheKey(contractAddress, [
             "query",
             "totalCirculatingSupply",
-            tokenId ?? "erc721",
+            tokenId ?? "0",
           ]),
         totalCount: (contractAddress: RequiredParam<ContractAddress>) =>
           createContractCacheKey(contractAddress, ["query", "totalCount"]),
--- a/packages/react-core/src/evm/utils/cache-keys.ts
+++ b/packages/react-core/src/evm/utils/cache-keys.ts
@@ -182,10 +182,12 @@ export const cacheKeys = {
           ),
         totalCirculatingSupply: (
           contractAddress: RequiredParam<ContractAddress>,
+          tokenId?: RequiredParam<BigNumberish>,
         ) =>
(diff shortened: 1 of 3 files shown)


packages/react-core/src/evm/utils/cache-keys.ts
```diff
@@ -182,10 +182,12 @@ export const cacheKeys = {
           ),
         totalCirculatingSupply: (
           contractAddress: RequiredParam<ContractAddress>,
+          tokenId?: RequiredParam<BigNumberish>,
         ) =>
           createContractCacheKey(contractAddress, [
             "query",
             "totalCirculatingSupply",
+            tokenId ?? "erc721",
```
jnsdls, line 190:
```suggestion
            tokenId ?? 0,
```

tokenId will be a BigNumberIsh so I would simply go with `0` here for erc721 case (erc721 can only ever be one token so the "tokenId" would be `0` if it had one, seems reasonable)

also just a note for additional learning, no need to make this change here:
- the _correct_ hierarchy for the query key (cache key) here would be `contract -> query -> tokenId -> totalCirculatingSupply
- in this case it's OK to leave it as is as we actually never invalidate cache for  individual tokens anywhere today, but in general the principle should be "least significant -> "most significant"

jnsdls (approved):
small notes, overall looks good and good find!

packages/react-core/src/evm/utils/cache-keys.ts
```diff
@@ -182,10 +182,12 @@ export const cacheKeys = {
           ),
         totalCirculatingSupply: (
           contractAddress: RequiredParam<ContractAddress>,
+          tokenId?: RequiredParam<BigNumberish>,
```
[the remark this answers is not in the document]
kien-ngo: ty ser

packages/react-core/src/evm/utils/cache-keys.ts
kien-ngo: 🙏 🙏

PR #1953, opened 2023-11-15
Override signMessage for SmartWallet class

nkrishang:
## Problem solved

Uses EIP-712 signature in `signMessage` for the updated EIP-1271 signature verification in smart wallet contracts (https://github.com/thirdweb-dev/contracts/pull/572)

## Changes made

    - **Before:** `SmartWallet.signMessage` directly uses ethers-js `signer.signMessage` to sign the message hash of the data to sign (i.e. `signMessage(ethers.utils.hashMessage(data))`)
    - **Now:** `SmartWallet.signMessage` signes EIP-712 typed data of the data to sign mixed with the smart wallet's domain separator. If this yields in invalid signature, the function falls back to the legacy method of signing.

## How to test

    - https://linear.app/thirdweb/issue/PLAT-1313/test-jam-[smartwallet][auth]-sign-in-with-smart-wallet


packages/wallets/src/evm/wallets/smart-wallet.ts
```diff
@@ -76,6 +78,90 @@ export class SmartWallet extends AbstractClientWallet<
...
+    const isDeployed = await connector.isDeployed();
+    if (!isDeployed) {
+      console.log(
+        "Account contract not deployed yet. Deploying account before signing message",
+      );
+      const tx = await erc4337Signer.sendTransaction({
+        to: await this.getAddress(),
+        data: "0x",
+      });
+      await tx.wait();
```
joaquim-verges, line 98: you can just use `connector.deployIfNeeded()` here

packages/wallets/src/evm/wallets/smart-wallet.ts
```diff
@@ -76,6 +78,90 @@ export class SmartWallet extends AbstractClientWallet<
...
+    const AccountMessage = [{ name: "message", type: "bytes" }];
+
+    /**
+     * We first try to sign the EIP-712 typed data i.e. the message mixed with the smart wallet's domain separator.
+     * If this fails, we fallback to the legacy signing method.
+     */
+    try {
+      const signature = await (
+        signer as providers.JsonRpcSigner
+      )._signTypedData(
```
joaquim-verges, line 118: can we use `signTypedDataInternal()` instead ? already handles a lot of this for you and its better to centralize this logic in one place

packages/wallets/src/evm/wallets/smart-wallet.ts
```diff
@@ -76,6 +78,90 @@ export class SmartWallet extends AbstractClientWallet<
...
+            [utils.hashMessage(message)],
+          ),
+        },
+      );
+
+      if (!signature) {
+        throw new Error("Failed to sign message");
+      }
+
+      const isValid = await checkContractWalletSignature(
```
joaquim-verges, line 138: I don't think it makes sense to validate the signature here, will also slow down the process quite a bit

nkrishang: I don't think there's another way to check whether to fallback to legacy method of signing. 

So this seems necessary to maintain backward compatibility with existing smart wallets out there.

packages/wallets/src/evm/wallets/smart-wallet.ts
nkrishang: I tried importing it but that didn't work; I'll try again.

joaquim-verges (approved):

PR #1962, opened 2023-11-16
Turkish Language & Cleaning up the language code

iketw:
### Combined Locales
Created a locales object to link each language code with its data.

### Unified Locale Function
Made a getLocale function. It combines the base language data with any extra bits you add. This means you don't have to write the same code for each language.

### Streamlined Language Functions
en, es, ja, tr etc now just use getLocale. It's cleaner and makes adding more languages easy. (Hopefully.) I would make them JSON files instead of separate ts files though. But I didn't want to overstep any boundaries.

## Added Turkish Language
Created by @selcukemiravci and @Visual917


iketw:
@Visual917 thanks again. I added the translation for ConnectWallet and added `tr` to the `locales` object.

I tested this with my RN app and it works great. 

Let me know if it looks good and I'll merge it :)

Visual917:
@iketw Awesome! Thank you and sorry I missed a spot ;)

jnsdls:
waiting for #1964 to merge -> will fix resulting conflicts -> merge

PR #1967, opened 2023-11-16
[sdk, chains] Add Arb Sepolia to defaultChains

kien-ngo:
## Problem solved

With future chains coming to the @chains package this PR might be irrelevant, however for now some users are requesting ArbSep to be one of the default chains (rightfully so) - so that they don't have to import it from the `@chains` package.

## Changes made

## How to test


packages/sdk/src/evm/constants/chains/ChainId.ts
```diff
@@ -16,6 +16,7 @@ export enum ChainId {
   OptimismGoerli = 420,
   Arbitrum = 42161,
   ArbitrumGoerli = 421613,
+  ArbitrumSepolia = 421614,
```
joaquim-verges, line 19: no need to add it here. this `ChainId` enum is actually deprecated. we should be using the chains package instead

packages/sdk/src/evm/constants/addresses/CONTRACT_ADDRESSES.ts
```diff
@@ -107,6 +106,14 @@ export const CONTRACT_ADDRESSES: Record<
     twRegistry: TWRegistry_address,
     twBYOCRegistry: AddressZero,
   },
+  [ChainId.ArbitrumSepolia]: {
```
joaquim-verges, line 109: no need for this, this is handled with the anyEVM flow

packages/sdk/src/evm/constants/chains/ChainId.ts
kien-ngo: Hmm the build would fail if I don't add it, since this enum is being referenced in other places. Should we keep it for now?
image

kien-ngo: Oh wait nvm I can just remove them altogether

.changeset/fifty-berries-appear.md
```diff
@@ -0,0 +1,6 @@
+---
+"@thirdweb-dev/chains": patch
+"@thirdweb-dev/sdk": patch
```
jnsdls, line 3: this no longer has any sdk cahnegs
```suggestion
```

PR #2018, opened 2023-11-28
[Embedded] Enforce encryption key for custom jwt

iketw:
## Problem solved

```
const embedded = useEmbeddedWallet();

embedded.connect({
  jwt: "<token>",
  encryptionKey: "<my-key>",
  strategy: "jwt",
});
```

## Changes made

## How to test


iketw:
@b-bot when you say it doesn't connect, I think you're expecting it to work with the same jwt but if you were using it without jwt before you used our custom encryption key. Can you try extracting the following encryptionKey from your jwt and passing that as `encryptionKey` ?

```
export function extractSubFromJwt(jwt: string): string {
  const parts = jwt.split(".");
  if (parts.length !== 3) {
    throw new Error("Invalid JWT format.");
  }

  const encodedPayload = parts[1];
  if (!encodedPayload) {
    throw new Error("Invalid JWT format.");
  }
  const decodedPayload = Buffer.from(encodedPayload, "base64").toString("utf8");
  return decodedPayload.sub;
}

// The call this function to pass it as encryptionKey
...
encryptionKey: extractSubFromJwt(jwt)
...
```

Issue #2002, opened 2023-11-23
Error Encountered When Running Code with `vite-node`


RSamaium:
Hello,

I'm reaching out to report an issue I've encountered while using `vite-node`. When I try to execute my code using `vite-node`, I'm facing an error that prevents the code from running correctly. This issue does not occur when I run the same code with standard Node.js.

### Environment
- Node.js version: 18+
- Vite-node version: Latest
- Operating system: Linux
- ThirdWeb SDK: Latest

### Steps to Reproduce

https://stackblitz.com/edit/stackblitz-starters-b7pyzu?file=index.js

To test:
- `node index.js` : works
- `npm start` or `npx vite-node index.js`: not works

### Actual Behavior

```
Error: missing revert data in call exception; Transaction reverted without a reason string [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (data="0x", transaction={"to":"0xfc575B22033dA2d86a52d8167B4fEd375c79f0Ab","data":"0x3b1475a7","accessList":null}, error={"reason":"missing response","code":"SERVER_ERROR","requestBody":"[{\"method\":\"eth_call\",\"params\":[{\"to\":\"0xfc575b22033da2d86a52d8167b4fed375c79f0ab\",\"data\":\"0x3b1475a7\"},\"latest\"],\"id\":43,\"jsonrpc\":\"2.0\"}]","requestMethod":"POST","serverError":{},"url":"https://mumbai.rpc.thirdweb.com/"}, code=CALL_EXCEPTION, version=providers/5.7.2)
    at Logger.makeError (/home/projects/stackblitz-starters-b7pyzu/node_modules/@ethersproject/logger/src.ts/index.ts:269:28)
    at Logger.throwError (/home/projects/stackblitz-starters-b7pyzu/node_modules/@ethersproject/logger/src.ts/index.ts:281:20)
    at checkError (/home/projects/stackblitz-starters-b7pyzu/node_modules/@ethersproject/providers/src.ts/json-rpc-provider.ts:66:16)
    at StaticJsonRpcBatchProvider.eval (/home/projects/stackblitz-starters-b7pyzu/node_modules/@ethersproject/providers/src.ts/json-rpc-provider.ts:642:20)
    at Generator.throw (<anonymous>)
    at rejected (/home/projects/stackblitz-starters-b7pyzu/node_modules/@ethersproject/providers/lib.esm/json-rpc-provider.js:17:65) {
[...]

PR #2085, opened 2023-12-14
feat: ethers `skipFetchSetup` support

hsynlms:
## Problem solved

This fixes https://github.com/thirdweb-dev/js/issues/2002

It may not be the clearest solution but the quick one. If the global environment variable `TW_SKIP_FETCH_SETUP ` is set, the Thirdweb SDK will assign the value of it to the `skipFetchSetup` option. When ethers.js v6 support arrives, we need to revert this commit. This is also commented on the source as a todo.

## Changes made

## How to test

- [x] Manual tests: set `TW_SKIP_FETCH_SETUP` to true and the issue explained in the related issue number #2002 will be gone

### DIFF
--- a/packages/sdk/src/evm/constants/urls.ts
+++ b/packages/sdk/src/evm/constants/urls.ts
@@ -286,12 +286,22 @@ export function getProviderFromRpcUrl(
           }
 
           // Otherwise, create a new provider on the specific network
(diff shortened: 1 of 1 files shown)


MananTank:
Awesome, @hsynlms Thank you very much for this! LGTM 🚀 

We will merge this shortly

PR #2106, opened 2023-12-20
feat(auth): Next.js App Router support

### LINKED ISSUE #2022: When will Thirdweb give support for Next.js App Router?
Next.js App Router became stable since [May 4, 2023](https://nextjs.org/blog/next-13-4).
However Thirdweb Auth SDK has no support for App Router according to [that document](https://portal.thirdweb.com/auth/server-frameworks/next#getting-started).

Is there any development roadmap?

hsynlms:
This feature brings Next.js App Router support to the Thirdweb Auth package and fixes #2022.

## Problem solved

Thirdweb Auth Next generates API handlers for Next.js Pages Router apps. But it is still not fully compatible with the new Next.js App Router.

Next.js App Router has become stable since v13.4.

In this PR, I kept existing router support which is Pages Router and added compatibility with App Router.

## Example usage

```ts
// file: /app/api/auth/[...thirdweb]/route.ts

// Next.js - Pages Router
import { ThirdwebAuth } from '@thirdweb-dev/auth/next'
import { PrivateKeyWallet } from '@thirdweb-dev/auth/evm'

export const { ThirdwebAuthHandler, getUser } = ThirdwebAuth({
  domain: '<your_app_domain>',
  wallet: new PrivateKeyWallet('<your_wallet_private_key>'),
})

export {
  ThirdwebAuthHandler as GET,
  ThirdwebAuthHandler as POST
}

// Next.js - App Router
import { ThirdwebAuthAppRouter } from '@thirdweb-dev/auth/next'
import { PrivateKeyWallet } from '@thirdweb-dev/auth/evm'

export const { ThirdwebAuthHandler, getUser } = ThirdwebAuthAppRouter({
  domain: '<your_app_domain>',
  wallet: new PrivateKeyWallet('<your_wallet_private_key>'),
})

export {
  ThirdwebAuthHandler as GET,
  ThirdwebAuthHandler as POST
}
```

## Changes made

## How to test

- [x] Manual tests: You are now able to use Next.js App Router route handler for Thirdweb Auth.


hsynlms:
Btw, I also brought `skipFetchSetup` development here (https://github.com/thirdweb-dev/js/pull/2085) because whenever it's missing, the RPC provider throws below error:

```
Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.7.2)
     at Logger.makeError (webpack-internal:///(rsc)/../../node_modules/.pnpm/@ethersproject+logger@5.7.0/node_modules/@ethersproject/logger/lib.esm/index.js:240:23)
     at Logger.throwError (webpack-internal:///(rsc)/../../node_modules/.pnpm/@ethersproject+logger@5.7.0/node_modules/@ethersproject/logger/lib.esm/index.js:249:20)
     at JsonRpcProvider.eval (webpack-internal:///(rsc)/../../node_modules/.pnpm/@ethersproject+providers@5.7.2/node_modules/@ethersproject/providers/lib.esm/json-rpc-provider.js:513:27)
     at Generator.throw (<anonymous>)
     at rejected (webpack-internal:///(rsc)/../../node_modules/.pnpm/@ethersproject+providers@5.7.2/node_modules/@ethersproject/providers/lib.esm/json-rpc-provider.js:34:40)
     at runNextTicks (node:internal/process/task_queues:60:5)
     at listOnTimeout (node:internal/timers:540:9)
     at process.processTimers (node:internal/timers:514:7) {
   reason: 'could not detect network',
   code: 'NETWORK_ERROR',
   event: 'noNetwork'
 }
```

hsynlms:
The below document and code example need to be updated accordingly.
https://portal.thirdweb.com/auth/server-frameworks/next
https://github.com/thirdweb-example/thirdweb-auth-next

joaquim-verges:
Thanks for the PR! We'll take a look shortly

packages/auth/src/next/helpers/user.ts
```diff
@@ -1,31 +1,23 @@
...
-  if (typeof req.cookies.get === "function") {
-    return req.cookies.get(cookie);
-  }
+import {
+  ThirdwebAuthContext,
+  ThirdwebAuthUser,
+} from "../types";
 
-  return (req.cookies as any)[cookie];
+export function getCookie(cookie: string): string | undefined {
```
adam-maj, line 14: Instead of supporting just the app router format, we need to support both the app router & the old pages directory - so we need to check for both cases here

adam-maj (changes requested):
Thanks a lot for putting up this PR, super helpful!

As far as approach - rather than replacing the support for auth in `pages` directory with `app` directory as it is here, we should support both `pages` (old code) and `app` (new code) since there are users still using both.

The best path here would likely just to create separate exports for each of them - ie the old export could be called `ThirdwebAuthPages` and the new one could be called `ThirdwebAuthApp` or something like `ThirdwebAuth` and `ThirdwebAuthRouter`, etc.

So essentially, would just add back the old code and change the export name, and then keep your new code as a separate folder with another export.

Feel free to make this change - or I'll get to it after my current focus (might take a few days/week+)

hsynlms:
@adam-maj sure. I will do it 'cause I really need this feature asap :)

packages/auth/src/next/helpers/user.ts
```diff
@@ -84,14 +69,12 @@ export async function getUser<
...
 
-  if (!ctx.callbacks?.onUser) {
-    return authenticatedUser;
-  }
-
-  const data = await ctx.callbacks.onUser(authenticatedUser, req);
-  if (!data) {
-    return authenticatedUser;
+  if (req && ctx.callbacks?.onUser) {
+    const data = await ctx.callbacks.onUser(authenticatedUser, req);
```
hsynlms, line 73: @adam-maj why do we need to pass the request object to the callback functions?

packages/auth/src/next/routes/logout.ts
```diff
@@ -1,45 +1,51 @@
...
   );
 
-  return res.status(200).json({ message: "Successfully logged out" });
+  response.cookies.set({
+    name: activeCookie,
+    value: '',
+    domain: ctx.cookieOptions?.domain,
+    path: ctx.cookieOptions?.path || "/",
+    sameSite: ctx.cookieOptions?.sameSite || "none",
+    expires: new Date(Date.now() + 5 * 1000),
```
hsynlms, line 45: @adam-maj I wonder why setting the cookie expiry to 5 seconds instead of deleting the cookie at that time?

hsynlms:
@adam-maj There are a few boring things to cover Pages and App Routers support in the same package. For more information see the below reasons list.

I suggest that it'd be better to create a new package (`next-legacy`) for Pages Router. So the `next` package will only be compatible with App Router and the below list will make no sense then.

- The only way to have the request object in App Router is route handlers. So most functions require the request object but there is no need. For example in the `getUser` and `ThirdwebAuthConfig` callback functions.
- We need to put a lot of TODOs to describe the things that need to be done when Pages Router is deprecated.
- It seems hard to give support for both routers in the same `ThirdwebAuthHandler` function. Because the returned async function signature varies with each router.
- Changing function names by adding suffixes (like `ThirdwebAuthPages`, `ThirdwebAuthApp`, etc.,)  makes things harder to maintain and I think is not good for the developer experience.

packages/auth/src/next/helpers/user.ts
adam-maj: Users should be able to access the request object on the callback functions in case they need to run their own business logic with anything else on headers/body/etc.

adam-maj:
> @adam-maj There are a few boring things to cover Pages and App Routers support in the same package. For more information see the below reasons list.
> 
> I suggest that it'd be better to create a new package (`next-legacy`) for Pages Router. So the `next` package will only be compatible with App Router and the below list will make no sense then.
> 
> * The only way to have the request object in App Router is route handlers. So most functions require the request object but there is no need. For example in the `getUser` and `ThirdwebAuthConfig` callback functions.
> * We need to put a lot of TODOs to describe the things that need to be done when Pages Router is deprecated.
> * It seems hard to give support for both routers in the same `ThirdwebAuthHandler` function. Because the returned async function signature varies with each router.
> * Changing function names by adding suffixes (like `ThirdwebAuthPages`, `ThirdwebAuthApp`, etc.,)  makes things harder to maintain and I think is not good for the developer experience.

Thanks for the update & comment! Given that many (from what I've seen, most) Next + Auth users are still using the `pages` directory because there are many people that haven't migrated over/haven't learned `app` router, I don't want to deprecate the pages directory support - as both are fully valid in Next 13 itself.

Instead of handling the same routes in one `ThirdwebAuthHandler`, I'd suggest just having 2 separate handler functions - like `ThirdwebAuth` and `ThirdwebAuthAppRouter` (this being the non-default). 

You should be able to keep `ThirdwebAuth` exactly the same as before. Then `ThirdwebAuthAppRouter` can change the implementation in all the ways you describe above to make it conducive to the `/app` router format. These can both ship from the same package, and we'll be sure to document it so it's clear what to use when.

With that setup, should be able to merge in the PR and get it all build ASAP

packages/auth/src/next/helpers/user.ts
hsynlms: I see. So no need to pass it in App Router because users always have access to headers, cookies via next/headers package.

hsynlms:
Hi @adam-maj 

We cannot keep the `ThirdwebAuth` function the same. Here is why:

- The defined inner async function `ThirdwebAuthHandler` will have no idea which router is requested. But we can return two handlers (`ThirdwebAuthHandler` and `ThirdwebAuthAppRouterHandler`) and there will be no breaking changes for the current implementations.
- The function returns an object that contains the `getUser` function. But this function will vary with each router. So the signatures will not be the same. The old one requires a request object but the App Router's `getUser` function will not ([because](https://github.com/thirdweb-dev/js/pull/2106#discussion_r1448363614)). In the end, there will be two `getUser` functions returned (`getUser` and `getAppRouterUser`).

So as a result, the auth handler and getting user functions will be named specifically for the App Router. When the Pages Router is deprecated it will be hard to rename new App Router functions (`ThirdwebAuthAppRouterHandler` and `getAppRouterUser`) to the default ones (`ThirdwebAuthHandler` and `getUser`) 'cause that will be a breaking change.

If you guys would like to do this in the same package, it seems that there will be two `ThirdwebAuth` functions for both routers.

adam-maj:
> Hi @adam-maj
> 
> We cannot keep the `ThirdwebAuth` function the same. Here is why:
> 
> * The defined inner async function `ThirdwebAuthHandler` will have no idea which router is requested. But we can return two handlers (`ThirdwebAuthHandler` and `ThirdwebAuthAppRouterHandler`) and there will be no breaking changes for the current implementations.
> * The function returns an object that contains the `getUser` function. But this function will vary with each router. So the signatures will not be the same. The old one requires a request object but the App Router's `getUser` function will not ([because](https://github.com/thirdweb-dev/js/pull/2106#discussion_r1448363614)). In the end, there will be two `getUser` functions returned (`getUser` and `getAppRouterUser`).
> 
> So as a result, the auth handler and get user function will be named specifically for the App Router. When the Pages Router is deprecated it will be hard to rename new App Router functions (`ThirdwebAuthAppRouterHandler` and `getAppRouterUser`) to the default ones (`ThirdwebAuthHandler` and `getUser`) 'cause that will be a breaking change.
> 
> If you guys would like to do this in the same package, it seems that there will be two `ThirdwebAuth` functions.

Yes - this is what I was referring to above - not keeping the same `ThirdwebAuth` function and try to case on the logic below - I mean have two fully separate `ThirdwebAuth` functions exported top-level from the package.

The old `ThirdwebAuth` function would be the same (support pages).

Then the new function would be a separate export `ThirdwebAuthAppRouter` - and you can build it fully with the concerns of Next 13 app router - it should be completely untethered. So there should be two separate `ThirdwebAuth` functions (the app router one can be explicitly named differently. And we'll cover how to use this in the docs!

adam-maj:
@hsynlms Thanks - this looks perfect! Have you tested both the router and pages versions? If not I can help with this, an I'll merge ASAP right after.

hsynlms:
@adam-maj I rebuilt the package and tested it with my App Router app. It works! But let's double check and ensure it works properly. So I need your tests too :)

adam-maj:
Btw @hsynlms one question - what's the reasoning behind the abstract wallets change in `packages/wallets/src/evm/wallets/abstract.ts` - was this intentional?

adam-maj:
@hsynlms Just tested for both app router and pages and works like a charm! Pending merge conflict resolution and the last question, should be good to go and I'll merge 😄

hsynlms:
> @hsynlms Just tested for both app router and pages and works like a charm! Pending merge conflict resolution and the last question, should be good to go and I'll merge 😄

Sure! See the below comment. That's why I edited the `packages/wallets/src/evm/wallets/abstract.ts` file.
https://github.com/thirdweb-dev/js/pull/2106#issuecomment-1864418796

packages/auth/src/next/routes/logout.ts
hsynlms: @adam-maj ping

hsynlms:
@adam-maj I think the PR is ready to go!

adam-maj:
@hsynlms Perfect, good to go - thanks for putting this together!

adam-maj (approved):

Issue #2113, opened 2024-01-02
[SmartWallet] Enable gassless option per transaction.


alko89:
Currently gasless option has to be enabled in the `SmartWalletConfig`. A great improvement would be to allow disabling this option per transaction, with optional config:

```typescript
const wallet = new SmartWallet(config);

await wallet.executeBatch([tx1, tx2], { gassless: false })

await wallet.send(tx, { gassless: false })
```

joaquim-verges:
thanks for the PR @alko89 ! will get this in next week

PR #2117, opened 2024-01-03
feat: allow enabling/disabling gasless tx on demand

alko89:
## Problem solved

Updates #2113

Allow enabling/disabling gasless tx on demand.

## Changes made

- [x] Added optional config to transaction calls - `execute`, `send`, `sendBatch`, `executeBatch`, `sendRaw`, `executeRaw`, `sendBatchRaw`, `executeBatchRaw` and `deploy`
- [x] Removed paymaster API flag initialization, which should always initialize the API now and change the condition in `createUnsignedUserOp` to use the `gasless` flag instead.
- [ ] [Discussion] Since paymaster API is now always initialized it should probably not be optional anymore?

## How to test

- Initialize a smart wallet with gasless enabled or disabled
- Pass opposite `gasless` flag to transaction


joaquim-verges:
thanks for this @alko89 ! I've done some changes to the critical path in #2098 - so we'll need to merge those. Feel free to do those changes on top of my branch, otherwise I can tackle it.

alko89:
@joaquim-verges doesn't seem the MRs collide much on specific parts, but I can update.

BTW, should I change the paymaster API to not be optional anymore? This change should ensure it will always be defined, but didn't want to introduce to many changes.

joaquim-verges:
@alko89 don't change it just yet, will be part of another refactor for custom paymasters.

The way you have it is fine

PR #2124, opened 2024-01-07
feat: allow enabling/disabling gasless tx on demand

alko89:
## Problem solved

Updates #2113

Allow enabling/disabling gasless tx on demand.

## Changes made

- [x] Added optional config to transaction calls - `execute`, `send`, `sendBatch`, `executeBatch`, `sendRaw`, `executeRaw`, `sendBatchRaw`, `executeBatchRaw` and `deploy`
- [x] Removed paymaster API flag initialization, which should always initialize the API now and change the condition in `createUnsignedUserOp` to use the `gasless` flag instead.

## How to test

- Initialize a smart wallet with gasless enabled or disabled
- Pass opposite `gasless` flag to transaction

### DIFF
--- a/packages/wallets/src/evm/connectors/smart-wallet/index.ts
+++ b/packages/wallets/src/evm/connectors/smart-wallet/index.ts
@@ -5,11 +5,12 @@ import { getVerifyingPaymaster } from "./lib/paymaster";
 import { create4337Provider } from "./lib/provider-utils";
 import {
   AccountContractInfo,
-  BatchData,
   FactoryContractInfo,
   ProviderConfig,
   SmartWalletConfig,
   SmartWalletConnectionArgs,
+  TransactionOptions,
+  UserOpOptions,
 } from "./types";
 import { ENTRYPOINT_ADDRESS } from "./lib/constants";
 import { EVMWallet } from "../../interfaces";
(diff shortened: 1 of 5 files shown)


alko89:
cc @joaquim-verges

packages/wallets/src/evm/connectors/smart-wallet/index.ts
```diff
@@ -556,7 +576,9 @@ export class SmartWalletConnector extends Connector<SmartWalletConnectionArgs> {
       this.accountApi.createUnsignedUserOp(
         this.aaProvider.httpRpcClient,
         tx,
-        batchData,
+        {
+          batchData
```
joaquim-verges, line 580: will also need the gasless flag here. the estimation would be different if its gasless vs not, might as well let users pass the config to estimateX

packages/wallets/src/evm/connectors/smart-wallet/index.ts
```diff
@@ -221,8 +230,9 @@ export class SmartWalletConnector extends Connector<SmartWalletConnectionArgs> {
    */
   async executeBatch(
     transactions: Transaction<any>[],
+    config?: { gasless?: boolean },
```
joaquim-verges, line 233: would be nice to have an exported type for this

joaquim-verges (commented):
thank you @alko89 - left a couple of comments, but can also merge and take it over next week

packages/wallets/src/evm/connectors/smart-wallet/index.ts
alko89: Ohh, thought gas usage is mostly the same. Will update

PR #2139, opened 2024-01-10
Resolve beacon from beacon-proxy

kumaryash90:
Check if the contract is a beacon-proxy, resolve beacon contract from beacon-slot, and then find out implementation address from beacon.

Ref: https://eips.ethereum.org/EIPS/eip-1967#beacon-contract-address


packages/sdk/src/evm/common/feature-detection/resolveContractUriFromAddress.ts
```diff
@@ -52,6 +52,12 @@ export async function resolveImplementation(
   }
 
   // check other proxy types
+  const beacon = await getBeaconFromStorageSlot(address, provider); // if the contract is BeaconProxy
+  if (beacon && beacon !== constants.AddressZero) {
+    // In case of a BeaconProxy, it is setup as BeaconProxy --> Beacon --> Implementation
+    // Hence we replace the proxy address with Beacon address, and continue further resolving below
+    address = beacon as string;
```
joaquim-verges, line 59: why do you need `as string` here?

packages/sdk/src/evm/common/feature-detection/resolveContractUriFromAddress.ts
```diff
@@ -52,6 +52,12 @@ export async function resolveImplementation(
   }
 
   // check other proxy types
+  const beacon = await getBeaconFromStorageSlot(address, provider); // if the contract is BeaconProxy
```
joaquim-verges, line 55: would be nice to parallelize this with `fetchBytecode` - you can promise all both

packages/sdk/src/evm/common/feature-detection/resolveContractUriFromAddress.ts
kumaryash90: fixed

packages/sdk/src/evm/common/feature-detection/resolveContractUriFromAddress.ts
kumaryash90: Updated

packages/sdk/src/evm/common/feature-detection/resolveContractUriFromAddress.ts
```diff
@@ -36,7 +36,11 @@ export async function resolveImplementation(
   address: string,
   provider: providers.Provider,
 ): Promise<{ address: string; bytecode: string }> {
-  const bytecode = await fetchBytecode(address, provider);
+  const fetched = await Promise.all([
```
joaquim-verges, line 39: its cleaner to destructure here instead of `fetched[0]` to access the values, you can do

`const [bytecode, beacon] = await Promise.all(....);`

joaquim-verges (approved):
lg

PR #2154, opened 2024-01-11
alpha

jnsdls:


packages/thirdweb/src/abi/lib/getFunctionSelector.ts
```diff
@@ -0,0 +1,23 @@
...
+const FN_SELECTOR_CACHE = /*@__PURE__*/ new Map<string, Hash>();
+
+export function getFunctionSelector(fn: string | AbiFunction): Hash {
+  const normalizedFn = getFunctionSignature(fn);
+  // if we have already computed the function selector before, return it
+  if (FN_SELECTOR_CACHE.has(normalizedFn)) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return FN_SELECTOR_CACHE.get(normalizedFn)!;
+  }
+  const functionSelector = slice(hash(normalizedFn), 0, 4);
```
jxom, line 20: Could you use `toFunctionSelector` here?

https://viem.sh/docs/utilities/toFunctionSelector#tofunctionselector

packages/thirdweb/src/abi/lib/normalizeSignature.ts
```diff
@@ -0,0 +1,65 @@
+export function normalizeSignature(signature: string): string {
```
jxom, line 1: do we need to export this in viem?

packages/thirdweb/src/abi/lib/getFunctionSelector.ts
jnsdls: I believe previously I was using a different hash function here - very well possible this can now just use that util function directly yes - will look into updating this

packages/thirdweb/src/abi/lib/normalizeSignature.ts
jnsdls: yep, I believe this is currently not exported by viem or abitype (I'll double check again on my side as well)

packages/thirdweb/src/abi/lib/getFunctionSelector.ts
jnsdls: had to update viem version but yes that's helpful!

packages/thirdweb/comparison/setup.md
```diff
@@ -0,0 +1,106 @@
...
+const client = createPublicClient({
+  chain: mainnet,
+  transport: http(),
+});
+
+const supply = await client.readContract({
+  address: "0x...",
+  abi,
+  functionName: "totalSupply",
+});
```
jxom, line 82: maybe for sake of comparison alignment

```suggestion
const contract = getContract({ address: '0x...', abi, client })
const supply = await contract.read.totalSupply()
```

jnsdls: thanks!

packages/thirdweb/src/react/ui/TransactionButton/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { Spinner } from "../components/Spinner.js";
+import {
+  useActiveAccount,
+  useActiveWalletChainId,
+  useSwitchActiveWalletChain,
+} from "../../providers/wallet-provider.js";
+import { useSendTransaction } from "../../hooks/contract/useSend.js";
+
+export type TransactionButtonProps = React.PropsWithChildren<
+  ButtonProps & {
```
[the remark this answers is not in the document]
jnsdls: oh? how come?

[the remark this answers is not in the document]
MananTank: They are just for internal styling props - it's very restricted. we can just expose the style and className instead

jxom:
I am curious if the benchmarks bust the LRU caches per iteration (or iterations are impure) to reflect a fair and real world comparison?

jnsdls:
we're #2356 now

jnsdls:
> I am curious if the benchmarks bust the LRU caches per iteration (or iterations are not pure) to reflect a fair and real world comparison?
> 
> Update: we added this to `isAddress` recently, but just realised we didn’t update benchmarks to reflect - we can update that on our side.

this is something I need to take a look at, good shout. We'll definitely want to do it (at least in the) "cold cache" benches

PR #2197, opened 2024-01-18
[react, react-native] ConnectWallet Details Modal UI, hide testnet faucet link by default

MananTank:


packages/react-native/src/evm/components/ConnectWallet.tsx
```diff
@@ -50,7 +50,12 @@ export type ConnectWalletProps = {
...
    *
-   * The default is `false`
+   * By default it is `true`, If you want to show the "Request Testnet funds" link when user is connected to a testnet, set this prop to `false`
+   *
+   * @example
+   * ```tsx
+   * <ConnectWallet hideTestnetFaucet={false} />
+   * ```
    */
   hideTestnetFaucet?: boolean;
```
iketw, line 60: Since this is true by default, should we rename this to `showTesnetFaucet`  instead?

MananTank: Good point, but let's try to avoid having a breaking change atm

PR #2213, opened 2024-01-24
Winston/embedded wallet v2

ElasticBottle:
## Problem solved

Adds basic api for embedded wallet operations

## Changes made

- [ ] Public API changes: TBD
- [ ] Internal API changes: 

## How to test

- [ ] Automated tests: NONE
- [ ] Manual tests: linked via `yalc` to `kitchen-sink-v2` in the `embedded-wallet` repo


packages/thirdweb/package.json
```diff
@@ -147,6 +147,7 @@
     "fuse.js": "7.0.0",
     "mipd": "0.0.5",
     "qrcode": "1.5.3",
+    "secrets.js-34r7h": "2.0.2",
```
jnsdls, line 150: consider using https://github.com/privy-io/shamir-secret-sharing instead, has more recent audits and also is half the size + more modern codebase (ts, tree shaking works etc)

jnsdls: alternatively we can also write it ourselves and get it audited

packages/thirdweb/src/wallets/embedded/core/authentication/error.ts
```diff
@@ -0,0 +1,36 @@
+export const USER_ABORT_AUTHENTICATION_ERROR_MESSAGE =
+  "User aborted the authentication flow";
+
+/**
+ * Error thrown when user aborts the authentication flow
+ */
+export class UserAbortError extends Error {
```
jnsdls, line 7: these errors are nice but custom errors also bloat bundle size (so let's not over-do it)

packages/thirdweb/src/wallets/embedded/core/base-url.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export const setBaseUrl = (url: string) => {
+  baseUrl = url;
+};
+/**
+ * @internal
+ */
+export const getBaseUrl = () => {
+  return baseUrl;
+};
```
jnsdls, line 19: ideally we set this via .env, no? I don't think there should be a public api for this

packages/thirdweb/src/wallets/embedded/core/wallet/index.ts
```diff
@@ -0,0 +1,286 @@
...
+ * import { embeddedWallet } from "thirdweb/wallets";
+ *
+ * const storage = createManagedStorage({})
+ * const wallet = await embeddedWallet({
+ *  storage,
+ * });
+ *
+ * ```
+ */
+export const embeddedWallet = async (arg: { storage: StorageType }) => {
```
jnsdls, line 26: this does nothing async, so it can be sync?

PR #2220, opened 2024-01-25
[SDK] Add erc1155.transferBatch

0xFirekeeper:
## Problem solved

Exposes erc1155.transferBatch API 

## Changes made

- [ ] Added transferBatch method to erc1155 standard
- [ ] Added test

## How to test

- [ ] custom.test.ts

### DIFF
--- a/packages/sdk/src/evm/core/classes/erc-1155.ts
+++ b/packages/sdk/src/evm/core/classes/erc-1155.ts
@@ -273,6 +273,40 @@ export class Erc1155<
   );
 
   /**
+   * Transfer multiple NFTs
+   * 
+   * @remarks Transfer multiple NFTs from the connected wallet to another wallet.
+   * 
+   * @example
+   * ```javascript
+   * // Address of the wallet you want to send the NFT to
+   * const toAddress = "{{wallet_address}}";
+   * // The token IDs of the NFTs you want to send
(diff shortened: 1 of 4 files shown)


packages/sdk/src/evm/core/classes/erc-1155.ts
```diff
@@ -272,6 +272,40 @@ export class Erc1155<
...
+   * ```
+   * 
+   * @twfeature ERC1155BatchTransferable
+   */
+  transferBatch = /* @__PURE__ */ buildTransactionFunction(
+    async (
+      to: AddressOrEns,
+      tokenIds: BigNumberish[],
+      amounts: BigNumberish[],
+      data: BytesLike = [0],
```
joaquim-verges, line 298: can you add an optional `from` here? if not passed in, defaults to current signer

0xFirekeeper: we don't do that for `transfer`, should I add it there too

0xFirekeeper: added [here](https://github.com/thirdweb-dev/js/pull/2220/commits/80c5e080efe8c8092c02a42be730232e2129efe1)

joaquim-verges (approved):

PR #2235, opened 2024-01-30
[Service-Utils] Usage Tracker Updates

farhanW3:
## Problem solved

Adding Engine to Usage tracker to allow for tracking of events from engine when being used.

## Changes made

- Updated UsageEventSchema to include `engine` enum & new column/properties to store more data.


packages/service-utils/src/cf-worker/usage.ts
```diff
@@ -52,6 +53,15 @@ const usageEventSchema = z.object({
...
   sdkPlatform: z.string().optional(),
   productName: z.string().optional(),
+  transactionValue: z.string().optional(),
+  pathname: z.string().optional(),
+  contractAddress: z.string().optional(),
+  errorCode: z.string().optional(),
+  httpStatusCode: z.string().optional(),
+  functionName: z.string().optional(),
+  extension: z.string().optional(),
+  retryCount: z.string().optional(),
```
arcoraven, line 63: number

arcoraven (approved):

packages/service-utils/src/cf-worker/usage.ts
farhanW3: updated

PR #2300, opened 2024-02-12
Resolve implementation update

kumaryash90:
- adds another minimal proxy pattern
- updates `resolveImplementation` to return original address if implementation address returns 0x

### DIFF
--- a/packages/sdk/src/evm/common/feature-detection/resolveContractUriFromAddress.ts
+++ b/packages/sdk/src/evm/common/feature-detection/resolveContractUriFromAddress.ts
@@ -80,15 +80,15 @@ export async function resolveImplementation(
           address: implementationAddress,
           bytecode: implBytecode,
         };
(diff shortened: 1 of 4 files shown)


packages/sdk/src/evm/common/feature-detection/resolveContractUriFromAddress.ts
```diff
@@ -71,10 +71,26 @@ export async function resolveImplementation(
...
+        );
+        return {
+          address: implementationAddress,
+          bytecode: implBytecode,
+        };
+      } catch (e) {
+        if (e instanceof Error) {
+          if (
+            !e.message.includes(
+              `Contract at ${implementationAddress} does not exist on chain`,
```
joaquim-verges, line 87: can you explain this change? what do we return if that's the error we hit?

joaquim-verges (approved):

PR #2424, opened 2024-03-07
chore: update size benchmark

jxom:
Reproduce with `bun run size` on Viem.

---

## PR-Codex overview
This PR optimizes the size of the `viem` package in the benchmarks folder by reducing its gzipped size from 35 KB to 3.78 KB. The `createPublicClient` function was replaced with `createClient`.

### Detailed summary
- Reduced `viem` package size from 35 KB to 3.78 KB
- Replaced `createPublicClient` with `createClient` in benchmarks README.md

### DIFF
--- a/packages/thirdweb/benchmarks/README.md
+++ b/packages/thirdweb/benchmarks/README.md
@@ -8,8 +8,8 @@ Imported bundle size to make a basic RPC call:
 
 - thirdweb: 6.1 KB gzipped
     - `import { createThirdwebClient, getRpcClient, defineChain }`
-- viem: 35 KB gzipped
-    - `import { createPublicClient, http }`
+- viem: 3.78 KB gzipped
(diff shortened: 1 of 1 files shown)


jnsdls:
@jxom is this using `size-limit`? I believe the numbers we have came from pkg-size.dev (uses `esbuild`) and I believe `size-limit` by default uses webpack?

_means I should re-check ourselves using size-limit too_

Regardless thanks for the PR - merging!

jxom:
Think `esbuild` might include some [weird side-effects](https://github.com/ai/size-limit/issues/351#issuecomment-1953069120). Not sure if that might improve your benchmark?

jnsdls:
@jxom sg, I've separately also updated the `thirdweb` size (which was also smaller using size-limit) - we want to be accurate with this so I think now the methodologies are the same at least :)

PR #2516, opened 2024-03-20
chore: update viem bench

jxom:
Updating Viem's benchmarks. Removed Viem from the summary to avoid underemphasising the improvements in performance to the Thirdweb SDK. 

---

## PR-Codex overview
This PR updates `viem` to version `2.8.14`, improving `encode transfer` benchmark performance significantly.

### Detailed summary
- Updated `viem` to version `2.8.14`
- Improved `encode transfer` benchmark performance
- Updated `@babel/core` to version `7.24.1`

> The following files were skipped due to too many changes: `pnpm-lock.yaml`

### DIFF
--- a/packages/thirdweb/benchmarks/encode-tx.ts
+++ b/packages/thirdweb/benchmarks/encode-tx.ts
@@ -60,6 +60,11 @@ const ethers6Contract = new ethers.Contract(
   ethers.getDefaultProvider(LOCAL_RPC),
 );
 
+const transfer_viem = viem.prepareEncodeFunctionData({
+  abi: ABI,
+  functionName: "transfer",
+})
+
 function randomBigint() {
   return BigInt(Math.floor(Math.random() * 1000));
 }
@@ -84,8 +89,7 @@ group("encode transfer (warm cache)", () => {
 
   bench("viem", async () => {
     viem.encodeFunctionData({
-      abi: ABI,
-      functionName: "transfer",
+      ...transfer_viem,
       args: [VITALIK_WALLET, randomBigint()],
     });
   });
@@ -126,8 +130,7 @@ group("encode transfer (cold cache)", () => {
 
   bench("viem", async () => {
     viem.encodeFunctionData({
-      abi: ABI,
-      functionName: "transfer",
+      ...transfer_viem,
       args: [VITALIK_WALLET, randomBigint()],
     });
   });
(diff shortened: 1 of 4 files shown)


jnsdls:
Thanks, will take a look tonight or tomorrow. I think we'll have to update some things to make the latest viem version work outside of benchmarks.

packages/thirdweb/benchmarks/encode-tx.ts
```diff
@@ -126,8 +130,7 @@ group("encode transfer (cold cache)", () => {
 
   bench("viem", async () => {
     viem.encodeFunctionData({
-      abi: ABI,
-      functionName: "transfer",
+      ...transfer_viem,
```
joaquim-verges, line 133: hmmm that's not really cold cache is it?

jxom: this is equivalent to [`transfer` on the thirdweb bench](https://github.com/thirdweb-dev/js/pull/2516/files/f90ffdb4ff9a71be8ca031d6d0c7514de865f14b#diff-3ba235fd736bcbaf2740c128e25e522f307f933328aba1fe3e8126b8a4b3eacbR112) in this suite.

joaquim-verges: right but one is compiled ahead of time, the other is computed at runtime. When we say cold cache, generally that means you start from nothing. In your case the user would have to already have done some work right?

jxom: From the perspective of benchmarking, I'd expect "cold cache" to refer to the invalidation of internal caches where the computation (between I/O) is happening. Doesn't matter how you generate your input vectors. If we are "starting from nothing", we shouldn't even load in the compiled vectors. ;P

joaquim-verges: maybe a better word is 'cold start'? Don't think its fair to do the runtime prepare work in both benchmarks here.

jnsdls:
@jxom I did some more perf work and incorporated these updates as part of #2520 - the new pattern in `viem` is still a bit faster than what we can hit for pure encode, I kept it in the benchmarks all the same (showing viem is faster in pure encode performance)

jxom:
Cool! Could use `prepareEncodeFunctionData` in your prepare utils? I kind of nerd sniped myself to add that into Viem this morning. Open to feedback if there’s something missing there.

jnsdls:
> Cool! Could use `prepareEncodeFunctionData` in your prepare utils? I kind of nerd sniped myself to add that into Viem this morning. Open to feedback if there’s something missing there.

possible, I haven't looked to deep yet but at initial glance it basically does the same thing as what I did in my `prepareMethod` call, there's a couple of small differences but yeah likely I can align it

jnsdls:
@jxom closing this since benchmarks are updated already - I'm going to start looking into removing some of the hand-written stuff that now exists in viem and will provide u feedback on those if I run into things separately

as always, appreciate the work you do and the discussions we have!

PR #2525, opened 2024-03-21
[SDK] Native privateKeyAccount implementation

gregfromstl:
## Problem solved

Implements `privateKeyAccount` and a number of its dependencies  While this moves many viem dependencies to the SDK, much of the typing is still dependent on viem given the viem type system's complexity. Future PRs will continue to migrate viem dependencies.

## Changes made

- [ ] `privateKeyAccount` now returns a `PrivateKeyAccount` object, which extends `Account` to guarantee `signTransaction` and adds a `publicKey`
- [ ] Adds `sign` utility fn
- [ ] Adds `signTransaction` action
- [ ] Adds `signatureToHex` utility fn
- [ ] Adds `signMessage` utility fn
- [ ] Adds `signTypedData` utility

## How to test

- [ ] Changes are covered by unit tests taken from viem to ensure feature parity and correctness

---

## PR-Codex overview
This PR enhances private key account support and adds conditional test suites based on environment variables. 

### Detailed summary
- Improved private-key account handling
- Conditional test execution based on environment variables
- Added functions for signing transactions and messages
- Updated test timeouts and error handling
- Refactored test descriptions for clarity

> The following files were skipped due to too many changes: `packages/thirdweb/src/wallets/interfaces/wallet.ts`, `packages/thirdweb/src/utils/signatures/sign.ts`, `packages/thirdweb/src/react/core/hooks/contract/useEstimate.ts`, `packages/thirdweb/src/utils/signatures/signature-to-hex.test.ts`, `packages/thirdweb/src/transaction/actions/estimate-gas-cost.test.ts`, `packages/thirdweb/src/utils/signatures/sign-message.ts`, `packages/thirdweb/src/utils/signatures/signature-to-hex.ts`, `packages/thirdweb/src/transaction/actions/sign-transaction.ts`, `packages/thirdweb/src/transaction/actions/sign-transaction.test.ts`, `packages/thirdweb/src/utils/signatures/sign-message.test.ts`, `packages/thirdweb/src/utils/signatures/sign-typed-data.ts`, `packages/thirdweb/src/adapters/viem.test.ts`, `packages/thirdweb/src/adapters/viem.ts`, `packages/thirdweb/test/src/typed-data.ts`, `packages/thirdweb/src/wallets/private-key.test.ts`, `packages/thirdweb/src/wallets/utils/getTokenBalance.ts`, `packages/thirdweb/src/utils/signatures/sign-typed-data.test.ts`, `packages/thirdweb/src/wallets/private-key.ts`


jxom:
Viem already exports these. What’s the point of copy & pasting the source code?

joaquim-verges:
> Viem already exports these. What’s the point of copy & pasting the source code?

we want to have low level control on signing to be able to inject other methods to handle the private keys like SSS, MPC and other explorations. This is a first step towards that.

jxom:
I mean `toAccount` in Viem is agnostic, and this is what teams are already doing for AA (permissionless.js, Biconomy, Alchemy, etc). Is there something it doesn’t achieve which we could add in if necessary?

jnsdls:
I think there are a couple of things that are leftover from the merge process I did, will clean it up for you today and then do a proper review

gregfromstl:
@jnsdls going to add some stuff to this later tonight to fix #2603 so no need for a full review yet

gregfromstl:
Okay should be ready for you now @jnsdls

jnsdls:
> Okay should be ready for you now @jnsdls

taking a look at this today!

packages/thirdweb/src/transaction/actions/sign-transaction.ts
```diff
@@ -0,0 +1,44 @@
...
+ * });
+ * ```
+ */
+export function signTransaction({
+  transaction,
+  privateKey,
+  // TODO: Setup more advanced transaction result typing
+}: SignTransactionOptions): Hex {
+  // We don't need to worry about sidecars yet since we're on an older version of viem
+  // TODO: Add eip4844 type check when upgrading viem past 2.8.0
```
jnsdls, line 35: I believe we've done this already? In general yeah we'll have to keep up to date because we'll want to keep updating

packages/thirdweb/src/wallets/interfaces/wallet.ts
```diff
@@ -45,6 +46,7 @@ export type Wallet<TWalletId extends WalletId = WalletId> = {
 export type Account = {
   // REQUIRED
   address: Address;
+  type?: "local" | "json-rpc";
```
jnsdls, line 49: hmmm, this basically aligns our `Account` type with viem's account type which I both like and dislike.

I'd prefer to keep the details of `account` internal to it instead of on the public api if possible.

packages/thirdweb/src/wallets/private-key.test.ts
```diff
@@ -0,0 +1,65 @@
+import { expect, test } from "vitest";
+import { parseEther, parseGwei } from "viem/utils";
```
jnsdls, line 2: I believe we have our own parsing functions for these, should use those

packages/thirdweb/src/wallets/private-key.ts
```diff
@@ -57,12 +64,65 @@ export type PrivateKeyAccountOptions = {
...
  */
-export function privateKeyAccount(options: PrivateKeyAccountOptions): Account {
-  if (!options.privateKey.startsWith("0x")) {
-    options.privateKey = "0x" + options.privateKey;
-  }
-  const viemAccount = privateKeyToAccount(options.privateKey as Hex);
-  return viemToThirdwebAccount(viemAccount, options.client);
+export function privateKeyAccount(
+  options: PrivateKeyAccountOptions,
+): LocalAccount {
```
jnsdls, line 69: unless we absolutely *have* to have different types for different types of accounts I'd love to be able to avoid this here

.changeset/empty-chairs-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+"thirdweb": minor
```
jnsdls, line 2: does this add any new APIs to the public API? - If so `minor` is appropriate, if not this should be `patch` (if it's only internals)

jnsdls (changes requested):
biggest challenge is the change in account type, I'd love to avoid that if we can

.changeset/empty-chairs-yawn.md
gregfromstl: If we have the new account type yes, otherwise no. Will update based on what we decide to do there.

packages/thirdweb/src/wallets/interfaces/wallet.ts
gregfromstl: Agree keeping it internal is ideal, but an issue came up re: this type attribute when I was playing around with the latest release this weekend. When calling `createWalletClient`, we only pass the address to viem because our accounts don't line up with theirs 1:1, and when that function only gets an address it defaults to a JsonRpcAccount. 
 This causes issues in certain viem wallet actions that branch based on the account type. I specifically noticed it in `getAddresses` where it can't fetch the addresses from rpc because it's a local account.

jnsdls: gotcha, this makes sense - for this useCase can we add a "hidden" flag on the account? Since the only use-case for this inside the adapter we should be able to then "check" for that hidden flag inside the adapter -> it solves the problem?

.changeset/empty-chairs-yawn.md
jnsdls: makes sense

packages/thirdweb/src/wallets/interfaces/wallet.ts
jnsdls: basically what I'm thinking is we can add a `Symbol("accountType")` key on the account, which is completely internal so nobody will be able to access it (won't leak to public api) but inside the adapter code we can reference it by looking up for the same symbol

jnsdls: let me know if that makes sense

packages/thirdweb/src/transaction/actions/sign-transaction.ts
gregfromstl: Ok cool, will just add a check for 4844 sidecars then

packages/thirdweb/src/wallets/interfaces/wallet.ts
gregfromstl: Yeah that makes sense. And should we just generate the other fields viem expects for an account (publicKey, source) within the adapter?

gregfromstl: actually publicKey might need to be another hidden field since privateKey doesn't stick around to generate it

jnsdls: so maybe `privateKey` itself is the hidden field which can server both as the flag and the data required inside the adapter

gregfromstl: Are there any security concerns with that? Otherwise that makes sense

jnsdls: I don't... think there are? Worth thinking through a couple of scenarios but if you have the pkey available *to put* there it means its already available in memory => if you're somehow attacking it you could already find it

gregfromstl: Yeah, was jw if third party libraries would be able to somehow access it and create downstream vulnerabilities unintentionally if they knew the symbol key. Tried and the only way I could get it was by converting the account to a string and pulling it from there, in which case the library would have to be either intentionally malicious or extremely determined to get to the raw private key to use it somehow.

packages/thirdweb/src/wallets/interfaces/wallet.ts
```diff
@@ -42,6 +42,9 @@ export type Wallet<TWalletId extends WalletId = WalletId> = {
   getConfig: () => CreateWalletArgs<TWalletId>[1];
 };
 
+// Symbols to track account origins internally for downstream adapters
+// eslint-disable-next-line better-tree-shaking/no-top-level-side-effects
+export const accountKeySymbol: unique symbol = Symbol("key");
```
jnsdls, line 47: instead of exporting this directly can we export a (still marked as `@internal`) function to access the key (which can return Hex | undefined)? - basically I don't want anyone to accidentally start using this symbol for anything but this specific purpose

jnsdls (commented):
one final thing, rest seems pretty good!

packages/thirdweb/src/wallets/interfaces/wallet.ts
gregfromstl: Copy that

jnsdls (approved):
LGTM!

0xFirekeeper:
Better than a CV

PR #2587, opened 2024-03-29
fix(cli): fix execution on windows and when prettier is not installed

jnsdls:


packages/thirdweb/src/cli/bin.ts
```diff
@@ -23,21 +23,39 @@ async function main() {
...
+        } catch {
+          bunAvailable = false;
         }
-      } catch {
-        bunAvailable = false;
       }
-      const runner = bunAvailable ? "bunx" : "npx";
+      let runner = "npx";
+
+      switch (true) {
```
joaquim-verges, line 44: Funky syntax but ok

packages/thirdweb/src/cli/commands/generate/generate.ts
```diff
@@ -330,3 +331,18 @@ function lowercaseFirstLetter(str: string) {
 function eventNameToPreparedEventName(name: string) {
   return `${lowercaseFirstLetter(name)}Event`;
 }
+
+const printedPrettierWarning = false;
+
+async function prettifyCode(code: string, options: Options) {
+  try {
+    const { format } = await import("prettier/standalone.js");
```
joaquim-verges, line 339: Nice

joaquim-verges (approved):

packages/thirdweb/src/cli/bin.ts
jnsdls: pretty standard tbh :)

PR #2588, opened 2024-03-29
fix: 721/1155 claimTo with custom currency

joaquim-verges:
## PR-Codex overview
The focus of this PR is to fix the claimTo function in both ERC1155 and ERC721 extensions to handle the case where the currency is not the native token address.

### Detailed summary
- Updated `claimTo` function in ERC1155 and ERC721 extensions to return 0 if currency is not native token address
- Added check for currency type before calculating total price


jnsdls (approved):
LGTM

PR #2632, opened 2024-04-02
feat: Wallet connection improvements

MananTank:
## PR-Codex overview
This PR improves wallet connection handling by changing the `accountsChanged` event to `accountChanged` and emitting a new `Account` object. 

### Detailed summary
- Renamed `accountsChanged` event to `accountChanged` event
- Emit new `Account` object instead of creating it in the connection manager
- Enhanced WalletConnect connection handling
- Updated types and interfaces for better clarity

### DIFF
--- a/packages/thirdweb/src/wallets/injected/index.ts
+++ b/packages/thirdweb/src/wallets/injected/index.ts
@@ -200,10 +200,15 @@ async function onConnect(
   }
 
   function onAccountsChanged(accounts: string[]) {
-    if (accounts.length === 0) {
-      onDisconnect();
+    if (accounts[0]) {
+      const newAccount = {
+        ...account,
+        address: getAddress(accounts[0]),
+      };
+
+      emitter.emit("accountChanged", newAccount);
     } else {
-      emitter.emit("accountsChanged", accounts);
+      onDisconnect();
     }
   }
 
--- a/packages/thirdweb/src/wallets/wallet-connect/index.ts
+++ b/packages/thirdweb/src/wallets/wallet-connect/index.ts
@@ -19,7 +19,6 @@ import {
   getRpcUrlForChain,
 } from "../../chains/utils.js";
 import type { Chain } from "../../chains/types.js";
-import { ethereum } from "../../chains/chain-definitions/ethereum.js";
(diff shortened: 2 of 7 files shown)


packages/thirdweb/src/wallets/wallet-connect/index.ts
```diff
@@ -521,3 +499,45 @@ async function getRequestedChainsIds(): Promise<number[]> {
...
+
+  if (options.chain) {
+    rpcMap[options.chain.id] = getRpcUrlForChain({
+      chain: options.chain,
+      client: options.client,
+    });
+  }
+
+  // limit optional chains to 10
+  const optionalChains = (options?.optionalChains || []).slice(0, 10);
```
joaquim-verges, line 522: was this a limitation from WC? or our choice?

packages/thirdweb/src/wallets/injected/index.ts
```diff
@@ -200,10 +200,15 @@ async function onConnect(
   }
 
   function onAccountsChanged(accounts: string[]) {
-    if (accounts.length === 0) {
-      onDisconnect();
+    if (accounts[0]) {
+      const newAccount = {
+        ...account,
+        address: getAddress(accounts[0]),
```
joaquim-verges, line 206: is there an assumption that the 'new' account will always be accounts[0] ?

also kind of weird to spread the old account functions and just override the address, but i guess in this case it'll work

joaquim-verges (approved):
looks fine, few questions.

also is this a feat: or a refactor: ?

packages/thirdweb/src/wallets/wallet-connect/index.ts
MananTank: Wallet connection does not work if send too many optional chains ( this happens in our dashboard - we also limit the optional chains in v4 )

packages/thirdweb/src/wallets/injected/index.ts
MananTank: Yes - the first one is the "active" account if the wallet has multiple accounts ( in case of metamask ) 

Re: the spread: We can also simply mutate the `address` property - but if we do that -  the account object will not trigger a re-render in react - so we clone the object and change the address since that's the only change we need to make

PR #2666, opened 2024-04-03
Fix getAddresses call for accounts converted to viem

gregfromstl:
`eth_accounts` rpc calls return an empty array for local accounts. This PR allows any converted account to use viem's `getAddresses` method by using the account's original supplied address.

Note: The unit test must use a remote chain since the test chain has a hardcoded response for `eth_accounts`.

---

## PR-Codex overview
This PR focuses on fixing the `getAddresses` method for accounts converted to `viem`.

### Detailed summary
- Fixed `getAddresses` method in `viem.ts` for converted accounts
- Added handling for `eth_accounts` method in `viem.ts`
- Imported `mainnet` chain definition in `viem.test.ts`
- Added test for getting address on live chain in `viem.test.ts`


jnsdls (approved):
Good catch, thx!

jnsdls:
Hm, maybe we can also shortcut the eth_chainId calls? Since we know which chain the user explicitly passed?

packages/thirdweb/src/adapters/viem.test.ts
```diff
@@ -62,4 +63,16 @@ describe("walletClient.toViem", () => {
     expect(txHash).toBeDefined();
     expect(txHash.slice(0, 2)).toBe("0x");
   });
+
+  test("should get address on live chain", async () => {
```
jnsdls, line 67: This may fail on fork test runs, let's see, if it does we should skip this test if TW_SECRET_KEY env var isn't set

gregfromstl:
> Hm, maybe we can also shortcut the eth_chainId calls? Since we know which chain the user explicitly passed?
> 
> Less of a fix and more of an optimization though.
> 
> Let me know what you think @gregfromstl

Great idea, can open a small PR for that right now

gregfromstl:
> > Hm, maybe we can also shortcut the eth_chainId calls? Since we know which chain the user explicitly passed?
> > Less of a fix and more of an optimization though.
> > Let me know what you think @gregfromstl
> 
> Great idea, can open a small PR for that right now

On second look, I don't think we can since viem's wallet client has the ability to add and switch chains after creation

PR #2866, opened 2024-04-26
Fix exception on chain with unknown chainId

MananTank:
## PR-Codex overview
The focus of this PR is to fix an exception when passing an unknown `chainId` to `ConnectButton`.

### Detailed summary
- Fixed exception handling in `NetworkSelector.tsx`
- Improved error handling in `utils.ts`
- Added `createChainMetadata` function for creating chain metadata

### DIFF
--- a/.changeset/khaki-hounds-drum.md
+++ b/.changeset/khaki-hounds-drum.md
@@ -1,0 +1,5 @@
+---
+"thirdweb": patch
+---
+
+Fix exception on passing chain with unknown chainId to ConnectButton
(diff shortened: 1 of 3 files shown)


MananTank:
creating a dev build for now - need to test this further / modify

MananTank:
we should also add the icon to `Chain` object type - a user can pass the icon to `defineChain` but it does not make it's way to `Chain`

jnsdls (approved):
I really don't like standardizing on the `chainmetadata` type rather than on the `chain` type (and adding anything it's missing in there)

that said this should be an internal change we can make later so it should not be a blocker

PR #2879, opened 2024-04-29
Add `onConnect` prop on `AutoConnect`

MananTank:
## PR-Codex overview
This PR adds an `onConnect` callback prop to `AutoConnect` component for successful wallet connection handling.

### Detailed summary
- Added `onConnect` prop to `AutoConnect` component
- Updated `ConnectButton` and `ConnectEmbed` to call `onConnect` callback
- Modified callback parameter name to `wallet` for consistency

### DIFF
--- a/packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
+++ b/packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
@@ -205,9 +205,15 @@ export function AutoConnect(props: AutoConnectProps) {
             message: `AutoConnect timeout : ${timeout}ms limit exceeded.`,
           });
 
-          connect(activeWallet);
-          if (onConnect) {
-            onConnect(activeWallet);
+          // connected wallet could be activeWallet or smart wallet
+          const connectedWallet = await connect(activeWallet);
+
+          if (!connectedWallet) {
+            setConnectionStatus("disconnected");
+          }
+
+          if (onConnect && connectedWallet) {
+            onConnect(connectedWallet);
           }
         } catch (e) {
           console.error("Failed to auto connect last active wallet");
--- a/packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
+++ b/packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
@@ -109,6 +109,19 @@ export type AutoConnectProps = {
    * />
    */
   accountAbstraction?: SmartWalletOptions;
+
+  /**
+   * Callback to be called on successful auto-connection of last active wallet. The callback is called with the connected wallet
+   *
+   * ```tsx
+   * <AutoConnect
+   *  onConnect={(wallet) => {
+   *    console.log("auto connected to", wallet)
+   *  }}
+   * />
+   * ```
+   */
+  onConnect?: (wallet: Wallet) => void;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
```diff
@@ -193,6 +206,9 @@ export function AutoConnect(props: AutoConnectProps) {
           });
 
           connect(activeWallet);
```
gregfromstl, line 208: Should we wait till after `connect` has resolved before calling `onConnect`?

packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
```diff
@@ -193,6 +206,9 @@ export function AutoConnect(props: AutoConnectProps) {
...
+          if (onConnect) {
```
gregfromstl, line 209: Should `onConnect` be called outside the try-catch so the user can provide their own error handling?

MananTank: error will be thrown before the wallet is connected, so I'm not sure how would `onConnect` handle the error here?

packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
MananTank: Good point - await-ing is now necessary because `connect` is not simply setting is active ( which was the case earlier ) - but now it may or may not connect a smart wallet instead, so need to await it and pass the returned wallet to `onConnect` - Good catch !

MananTank: Done

packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
gregfromstl: Just thinking if the user passes a function that errors to `onConnect` and doesn't handle it, the `connect` try-catch will be triggered and they'll get an error "Failed to auto connect last active wallet" which misleads them to think `connect` failed instead of `onConnect`

packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
```diff
@@ -192,7 +205,16 @@ export function AutoConnect(props: AutoConnectProps) {
...
+          // connected wallet could be activeWallet or smart wallet
+          const connectedWallet = await connect(activeWallet);
+
+          if (!connectedWallet) {
+            setConnectionStatus("disconnected");
+          }
+
+          if (onConnect && connectedWallet) {
+            onConnect(connectedWallet);
+          }
```
jnsdls, line 217: would be easier to read here if this was instead:

```js
if(connectedWallet){
  if(onConnect){
    onConnect(connectedWallet);
  }
} else { 
  setConnectionStatus("disconnected");
}
```

or if we can return early here that could also do the trick, negation as the "first" option just always makes things difficult to read

jnsdls (approved):
one style-nit, but otherwise lgtm

packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
MananTank: ok - let me wrap it in try-catch

MananTank: done

packages/thirdweb/src/react/core/hooks/connection/useAutoConnect.tsx
MananTank: done

PR #2899, opened 2024-04-30
Fix event duplication in useContractEvents

gregfromstl:
## Problem solved

There was a bug where if the events filters passed to `useContractEvents` updated every render (which in most naive use cases they do), duplicate events are repeatedly added to the events list until a new block is mined. This was because every time a new poller spawned, it didn't know what block the previous poller had left off on, and so it just added the events from the current block.

## Changes made

- [ ] Internal API changes: The latestBlockNumber is now tracked with a ref and optionally passed to the poller
- [ ] Updated docs to be more descriptive

### DIFF
--- a/packages/thirdweb/src/rpc/watchBlockNumber.ts
+++ b/packages/thirdweb/src/rpc/watchBlockNumber.ts
@@ -32,12 +32,13 @@ function createBlockNumberPoller(
   client: ThirdwebClient,
   chain: Chain,
   overPollRatio?: number,
+  latestBlockNumber?: bigint,
 ) {
   let subscribers: Array<(blockNumber: bigint) => void> = [];
   let blockTimesWindow: number[] = [];
 
   let isActive = false;
-  let lastBlockNumber: bigint | undefined;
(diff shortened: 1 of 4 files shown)


packages/thirdweb/src/rpc/watchBlockNumber.ts
```diff
@@ -49,8 +49,6 @@ function createBlockNumberPoller(
   async function poll() {
     // stop polling if there are no more subscriptions
     if (!isActive) {
-      lastBlockNumber = undefined;
-      lastBlockAt = undefined;
```
joaquim-verges, line 53: just checking we don't need this anymore? I know @jnsdls added this to fix a bug when subscribing after a few blocks that would re-play a bunch of events.

that said there is a unit test for it now so I think we should be good!

jnsdls: yeah if the unit tests are happy then I'm happy!

joaquim-verges (approved):

PR #2900, opened 2024-04-30
Create wallet ID constants file

gregfromstl:
Doesn't cover all cases, will incrementally replace uses of wallet IDs with these constants

---

## PR-Codex overview
This PR updates wallet constants and uses them consistently across the codebase for popular wallets like Coinbase and MetaMask.

### Detailed summary
- Updated wallet constants for Coinbase, MetaMask, Rainbow, and Zerion
- Replaced hardcoded wallet IDs with constants in various files
- Improved consistency and readability in wallet-related code

### DIFF
--- a/legacy_packages/wallets/test/smart-wallet-integration.test.ts
+++ b/legacy_packages/wallets/test/smart-wallet-integration.test.ts
@@ -1,7 +1,7 @@
 /* eslint-disable better-tree-shaking/no-top-level-side-effects */
 import { SmartWallet } from "../src/evm/wallets/smart-wallet";
 import { LocalWallet } from "../src/evm/wallets/local-wallet";
-import { ArbitrumSepolia } from "@thirdweb-dev/chains";
+import { BaseSepoliaTestnet } from "@thirdweb-dev/chains";
 import { ThirdwebSDK, SmartContract } from "@thirdweb-dev/sdk";
 import { checkContractWalletSignature } from "../src/evm/connectors/smart-wallet/lib/check-contract-wallet-signature";
 import { describe, it, expect, beforeAll } from "vitest";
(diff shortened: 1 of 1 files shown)


packages/thirdweb/src/wallets/create-wallet.ts
```diff
@@ -62,9 +63,9 @@ export function createWallet<const ID extends WalletId>(
      * COINBASE WALLET VIA SDK
      * -> if no injected coinbase found, we'll use the coinbase SDK
      */
-    case "com.coinbase.wallet": {
+    case COINBASE: {
       return coinbaseWalletSDK(
-        creationOptions as CreateWalletArgs<"com.coinbase.wallet">[1],
+        creationOptions as CreateWalletArgs<typeof COINBASE>[1],
```
joaquim-verges, line 68: maybe we just do `typeof X` everywhere rather than having another const

MananTank:
I'm not really sure if it's useful to export a subset of wallet ids as constants. Could be more confusing than simply using the wallet id as string for library user

gregfromstl:
> I'm not really sure if it's useful to export a subset of wallet ids as constants. Could be more confusing than simply using the wallet id as string for library user

The idea is mainly for internal use so we don't have the same manually typed string everywhere that could break with a typo. A user could theoretically import them but they shouldn't need to

jnsdls:
> > I'm not really sure if it's useful to export a subset of wallet ids as constants. Could be more confusing than simply using the wallet id as string for library user
> 
> The idea is mainly for internal use so we don't have the same manually typed string everywhere that could break with a typo. A user could theoretically import them but they shouldn't need to

for internal use this makes sense to me, I don't think we should export them from the `/exports` at all (so they will not be publicly available, I don't think that is useful)

jnsdls (approved):

PR #2921, opened 2024-05-04
Adds use client directives

gregfromstl:
## PR-Codex overview
This PR adds the 'use client' directive to all client-only components.

### Detailed summary
- Added 'use client' directive to all client-only components.

> The following files were skipped due to too many changes: `packages/thirdweb/src/react/web/ui/ConnectWallet/Modal/screen.tsx`, `packages/thirdweb/src/react/web/wallets/in-app/InAppWalletSelectionUI.tsx`, `packages/thirdweb/src/react/web/ui/ConnectWallet/Modal/InjectedConnectUI.tsx`, `packages/thirdweb/src/react/web/wallets/in-app/InAppWalletOTPLoginUI.tsx`

### DIFF
--- a/packages/thirdweb/src/exports/react.ts
+++ b/packages/thirdweb/src/exports/react.ts
@@ -1,5 +1,3 @@
-"use client";
-
 export { darkTheme, lightTheme } from "../react/web/ui/design-system/index.js";
 export type {
   Theme,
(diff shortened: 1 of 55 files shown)


packages/thirdweb/src/exports/react.ts
```diff
@@ -1,3 +1,5 @@
+"use client";
```
jnsdls, line 1: do we *need* it here because not everything that's exported from here needs to be client only?

jnsdls (commented):
one question, rest looks good to me

packages/thirdweb/src/exports/react.ts
gregfromstl: you're right, meant to remove that one

jnsdls (approved):

PR #2951, opened 2024-05-07
Fix(Marketplace): Increase gas on writes

gregfromstl:
## PR-Codex overview
This PR increases gas on MarketplaceV3 writes for various auction and listing operations to account for router calls.

### Detailed summary
- Increased gas on MarketplaceV3 writes for bid, buyout, cancel auction, create listing, update listing, execute sale, and create auction operations
- Updated `stringify` import in `TransactionButton` component

### DIFF
--- a/packages/thirdweb/src/react/web/ui/TransactionButton/index.tsx
+++ b/packages/thirdweb/src/react/web/ui/TransactionButton/index.tsx
@@ -7,6 +7,7 @@ import {
 } from "../../../../transaction/actions/wait-for-tx-receipt.js";
 import type { PreparedTransaction } from "../../../../transaction/prepare-transaction.js";
 import type { TransactionReceipt } from "../../../../transaction/types.js";
+import { stringify } from "../../../../utils/json.js";
 import {
   useActiveAccount,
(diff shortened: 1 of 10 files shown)


packages/thirdweb/src/react/web/ui/TransactionButton/index.tsx
```diff
@@ -165,9 +166,9 @@ export function TransactionButton(props: TransactionButtonProps) {
 
           if (onTransactionConfirmed) {
             const receipt = await doWaitForReceipt(result);
-            if (receipt.status === "reverted")
+            if (receipt.status !== "success")
```
joaquim-verges, line 169: what other status is there? just curious

joaquim-verges (approved):

packages/thirdweb/src/react/web/ui/TransactionButton/index.tsx
gregfromstl: There's a dropped and replaced status I think but apparently viem doesn't include that so I'm gonna undo this

PR #2953, opened 2024-05-08
Fix blockNumber inconsistencies

gregfromstl:
Fixes inconsistent block numbers when fetching events.

Prevents errors from inconsistent `eth_blockNumber` return values, `fromBlock` now waits for the latest block in `getContractEvents` to catch up if within a reasonable range.

`getContractEvents` will now properly handle `blockRange` interaction with `fromBlock` and `toBlock` given they are both inclusive (i.e. it will only return logs for the number of blocks specified in `blockRange`).

---

## PR-Codex overview
This PR enhances event handling in the `thirdweb` package by fixing block number inconsistencies, ensuring inclusive block ranges, and optimizing block retrieval.

### Detailed summary
- Fixed inconsistent block numbers
- Ensured inclusive block ranges for `toBlock` and `fromBlock`
- Improved handling of block range interactions
- Optimized block retrieval logic for latest block number

### DIFF
--- a/.changeset/witty-pumpkins-do.md
+++ b/.changeset/witty-pumpkins-do.md
@@ -4,6 +4,6 @@
 
 Fixes inconsistent block numbers when fetching events.
 
-Prevents errors from inconsistent `eth_blockNumber` return values, `fromBlock` now reverts back to "latest" when greater than the latest block.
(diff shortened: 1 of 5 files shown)


gregfromstl:
@joaquim-verges let me know what you think of these changes, it's a bit of a hacky fix but we're seeing some inconsistencies between the poller's block number, the block number fetched in here, and the block number `eth_getLogs` is accepting so this was the best way to keep them in sync that I could think of

PR #3024, opened 2024-05-14
Add BSC mainnet and testnet

kien-ngo:


kien-ngo:
Damn looks like I hit the size limit. @jnsdls any chance we can increase the "floor" for this one?

jnsdls (approved):

PR #3051, opened 2024-05-16
Fix v4 marketplace calls

gregfromstl:
Adds gas boost for V4 marketplace writes as done in V5

---

## PR-Codex overview
This PR enhances gas estimates on marketplace transactions by introducing `CallOverridesWithExtraGas`. 

### Detailed summary
- Added `CallOverridesWithExtraGas` to improve gas estimates on marketplace writes
- Updated transaction context options to include `overrides` with extra gas
- Modified `TransactionContext` to use `CallOverridesWithExtraGas`
- Implemented extra gas handling in transaction overrides
- Added `extraGas` property to overrides in marketplace transaction methods


legacy_packages/sdk/src/evm/core/classes/internal/marketplace/marketplace-auction.ts
```diff
@@ -254,6 +254,9 @@ export class MarketplaceAuction {
             startTime: BigNumber.from(listingStartTime),
           } as IMarketplace.ListingParametersStruct,
         ],
+        overrides: {
+          extraGas: 50_000n,
```
jnsdls, line 258: I believe the `_` notation breaks some older bundlers, to be super safe can we do `BigInt(50000)` everywhere here

legacy_packages/sdk/src/evm/core/classes/transactions.ts
```diff
@@ -449,13 +452,18 @@ export class Transaction<
...
-    const overrides: CallOverrides = { ...gasOverrides, ...this.overrides };
+    const overrides: CallOverridesWithExtraGas = { ...gasOverrides, ...this.overrides };
 
     // First, if no gasLimit is passed, call estimate gas ourselves
     if (!overrides.gasLimit) {
       overrides.gasLimit = await this.estimateGasLimit();
     }
 
+    if (overrides.extraGas) {
+      const originalLimit = await overrides.gasLimit;
```
joaquim-verges, line 463: do we need to await again here?

legacy_packages/sdk/src/evm/core/classes/internal/marketplace/marketplace-direct.ts
```diff
@@ -485,6 +503,9 @@ export class MarketplaceDirect {
       return Transaction.fromContractWrapper({
         contractWrapper: this.contractWrapper,
         method: "cancelDirectListing",
+        overrides: {
+          extraGas: BigInt(50000),
```
joaquim-verges, line 507: nit: use BigNumber.from(50000) here since that's what we use everywhere in v4

also did you verify that fixes the issue in v4? reason im asking is because in v4 we do have already a gas bump for dynamic contracts IIRC

jnsdls: or just `50000` since it later goes into `BigNumber.from()` anyways

legacy_packages/sdk/src/evm/core/classes/transactions.ts
gregfromstl: We technically shouldn't but TS requires it because `gasLimit` type on the overrides object is `Promise<BigNumber> | BigNumber`

legacy_packages/sdk/src/evm/core/classes/internal/marketplace/marketplace-direct.ts
gregfromstl: Haven't confirmed it yet, trying to now

PR #3064, opened 2024-05-17
feat: In-app wallets for react native v5

joaquim-verges:
## PR-Codex overview
This PR introduces in-app wallet support for React Native, updates authentication modules, and refactors file paths.

### Detailed summary
- Added in-app wallet support for React Native
- Updated authentication modules and interfaces
- Refactored file paths for consistency

> The following files were skipped due to too many changes: `packages/thirdweb/src/extensions/erc721/write/sigMint.ts`, `packages/thirdweb/src/utils/jwt/encode-jwt.ts`, `packages/thirdweb/src/wallets/in-app/implementations/lib/auth/passkeys.ts`, `packages/thirdweb/src/extensions/erc20/write/sigMint.ts`, `packages/thirdweb/src/wallets/in-app/implementations/lib/auth/index.ts`, `packages/thirdweb/src/wallets/in-app/native/helpers/auth/cognitoAuth.ts`, `packages/thirdweb/src/extensions/airdrop/write/airdropERC20WithSignature.ts`, `packages/thirdweb/src/extensions/airdrop/write/airdropERC721WithSignature.ts`, `packages/thirdweb/src/extensions/airdrop/write/airdropERC1155WithSignature.ts`, `packages/thirdweb/src/wallets/in-app/core/lib/in-app-wallet-calls.ts`, `packages/react-native-adapter/package.json`, `packages/thirdweb/src/extensions/erc4337/account/common.ts`, `packages/thirdweb/src/wallets/in-app/web/types.ts`, `packages/thirdweb/src/wallets/in-app/native/helpers/wallet/recoveryCode.ts`, `packages/thirdweb/src/wallets/in-app/implementations/lib/core/in-app-wallet.ts`, `packages/thirdweb/src/wallets/in-app/native/helpers/constants.ts`, `packages/thirdweb/package.json`, `packages/thirdweb/src/wallets/in-app/native/helpers/auth/middleware.ts`, `packages/thirdweb/src/wallets/in-app/native/helpers/wallet/encryption.ts`, `packages/thirdweb/src/wallets/in-app/native/helpers/wallet/creation.ts`, `packages/thirdweb/src/wallets/in-app/core/authentication/type.ts`, `packages/thirdweb/src/wallets/in-app/native/helpers/storage/local.ts`, `packages/thirdweb/src/wallets/in-app/core/authentication/index.ts`, `packages/thirdweb/src/wallets/in-app/web/lib/web-connector.ts`, `packages/thirdweb/src/wallets/in-app/native/helpers/wallet/retrieval.ts`, `packages/thirdweb/src/wallets/in-app/native/native-connector.ts`, `packages/thirdweb/src/wallets/in-app/native/helpers/api/fetchers.ts`, `packages/thirdweb/src/wallets/in-app/native/auth.ts`, `pnpm-lock.yaml`


packages/thirdweb/package.json
```diff
@@ -166,8 +166,18 @@
     "viem": "2.10.9"
   },
   "peerDependencies": {
+    "@aws-sdk/client-lambda": "3.549.0",
+    "@aws-sdk/credential-providers": "3.549.0",
+    "amazon-cognito-identity-js": "^6.3.12",
+    "aws-amplify": "^5.3.18",
```
jnsdls, line 172: all of these *have* to be marked as optional below in `"peerDependenciesMeta"` otherwise package managers will report them as missing peer dependencies in projects where people don't also have the adapter installed (which we do not want)

packages/thirdweb/package.json
```diff
@@ -166,8 +166,18 @@
...
     "ethers": "^5 || ^6",
     "react": ">=18",
+    "react-native-aes-gcm-crypto": "^0.2.2",
+    "react-native-inappbrowser-reborn": "^3.7.0",
+    "react-native-mmkv": ">=2.11.0",
+    "react-native-quick-base64": "^2.1.0",
+    "react-native-quick-crypto": "^0.6.1",
+    "shamir-secret-sharing": "^0.0.3",
```
jnsdls, line 180: ditto

packages/react-native-adapter/package.json
```diff
@@ -25,7 +25,17 @@
...
+    "aws-amplify": "^5.3.18",
+    "amazon-cognito-identity-js": "^6.3.12",
+    "react-native-inappbrowser-reborn": "^3.7.0",
+    "react-native-mmkv": ">=2.11.0",
+    "react-native-aes-gcm-crypto": "^0.2.2",
+    "react-native-quick-base64": "^2.1.0",
+    "react-native-quick-crypto": "^0.6.1",
+    "@aws-sdk/client-lambda": "3.549.0",
+    "@aws-sdk/credential-providers": "3.549.0",
+    "shamir-secret-sharing": "^0.0.3"
```
jnsdls, line 38: we should probably pin all of these and not allow `^` (we'll explicitly decide what to upgrade and when)

packages/thirdweb/package.json
```diff
@@ -214,6 +224,7 @@
     "@types/react": "18.2.17",
     "@vitejs/plugin-react": "^4.2.1",
     "@vitest/ui": "1.5.0",
+    "amazon-cognito-identity-js": "^6.3.12",
```
jnsdls, line 227: how come only this one is in devDeps?

packages/thirdweb/src/utils/random.ts
```diff
@@ -3,8 +3,13 @@ import { uint8ArrayToHex } from "./encoding/hex.js";
 /**
  * @internal
  */
-export async function randomBytes(length = 32) {
-  return uint8ArrayToHex(
-    globalThis.crypto.getRandomValues(new Uint8Array(length)),
-  );
+export function randomBytes(length = 32) {
```
jnsdls, line 6: probably rename this
```suggestion
export function randomBytesHex(length = 32) {
```

packages/thirdweb/src/utils/random.ts
```diff
@@ -3,8 +3,13 @@ import { uint8ArrayToHex } from "./encoding/hex.js";
...
+  return uint8ArrayToHex(randomBytesBuffer(length));
+}
+
+/**
+ * @internal
+ */
+export function randomBytesBuffer(length = 32) {
```
jnsdls, line 13: and then this is
```suggestion
export function randomBytes(length = 32) {
```

packages/thirdweb/src/wallets/in-app/core/authentication/index.ts
```diff
@@ -1,32 +1,41 @@
...
+  let ewSDK: InAppConnector;
+  if (isBrowser()) {
+    const { InAppWebConnector } = await import(
+      "../../web/lib/web-connector.js"
+    );
+    ewSDK = new InAppWebConnector({
+      client: client,
+    });
+  } else {
+    const { InAppNativeConnector } = await import(
```
jnsdls, line 31: this means if someone (for some reason) calls this in node we end up in the native path, likely we want to explicitly handle both web and native and then throw an error in node (for now) - we actually *can* make this work in node though right?

packages/thirdweb/src/wallets/in-app/native/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+  ROUTE_HEADLESS_OAUTH_LOGIN,
+  THIRDWEB_SESSION_NONCE_HEADER,
+} from "./helpers/constants.js";
+import { createErrorMessage } from "./helpers/errors.js";
+import { isDeviceSharePresentForUser } from "./helpers/storage/local.js";
+import { getCognitoUser, setCognitoUser } from "./helpers/storage/state.js";
+import type { VerifiedTokenResponse } from "./helpers/types.js";
+
+const HEADERS = {
+  [THIRDWEB_SESSION_NONCE_HEADER]: ANALYTICS.nonce,
```
jnsdls, line 42: this... just looks odd? what is this?

packages/thirdweb/src/wallets/in-app/native/helpers/analytics.ts
```diff
@@ -0,0 +1,5 @@
+import { randomBytes } from "../../../../utils/random.js";
+
+export const ANALYTICS = {
+  nonce: randomBytes(16),
```
jnsdls, line 4: wat

packages/thirdweb/src/wallets/in-app/web/index.ts
```diff
@@ -0,0 +1,36 @@
+// export {
```
jnsdls, line 1: empty file -> delete

jnsdls (commented):
initial review

packages/thirdweb/src/wallets/in-app/native/helpers/analytics.ts
joaquim-verges: Yeah it's really weird. Though I think it's actually used and important. @elasticbottle can you confirm?

packages/react-native-adapter/package.json
joaquim-verges: pinned the JS libs, but actually for the native ones, there's no point adding them here right? since the consumer has to add them themselves anyways

packages/thirdweb/package.json
joaquim-verges: removed. should i add all the peer deps in here? i think its working right now because of the legacy package

packages/thirdweb/src/wallets/in-app/core/authentication/index.ts
joaquim-verges: node not supported right now, will throw

packages/thirdweb/src/wallets/in-app/native/helpers/analytics.ts
joaquim-verges: this is actually used, but for tracking purposes only. all requests seems to be logged with this nonce header

not sure if we actually use to track stuff, maybe to get a view of an entire login session like latency?

packages/thirdweb/src/wallets/in-app/web/index.ts
joaquim-verges: actually need to re-export those

joaquim-verges: deleting for now, we can re-add later

packages/react-native-adapter/package.json
jnsdls: yeah so:
- for things consumer *has* to install additionally (native deps for linking) -> peer dep + dev dep, both with `^` version
- for things that are pure JS deps, pin them and make them direct dependencies

packages/thirdweb/package.json
jnsdls: you should add as peer + dev dependencies all dependencies that you import from in the code (and they should all be optional peer deps since they are only needed if react native is being used)

packages/thirdweb/src/wallets/in-app/native/helpers/analytics.ts
jnsdls: I guess that's fine but I would've expected us to move it to where we do any fetches (and just call a function then that returns the "cached" value if its already been called) -> right now this will generate random bytes the moment this file gets included in the bundle which seems like the wrong time to do it

packages/thirdweb/src/wallets/in-app/native/auth.ts
jnsdls: similar to the other comment this should just import a function that generates the nonce the first time it is used and then re-uses it afterwards, and it should call it *when* the fetch is being made

packages/react-native-adapter/package.json
joaquim-verges: yep

packages/thirdweb/src/wallets/in-app/native/helpers/analytics.ts
joaquim-verges: yeah ill change that

packages/react-native-adapter/package.json
joaquim-verges: so its actually I think this works

## Adapter
- hard = js deps with pinned version
- peer = mandatory native deps with `*` version (or maybe `^` if we need a particular min version)

## main package
- dev = js + native deps (with pinned version)

idea is that i only need dev deps in the main package to compile, since the idea is that to use RN you need to install the adapter AND all the adapter peers

jnsdls: that's fine but there's no reason to not also list the peers as optional in the main package, feels cleaner to include them there

re `*` versions we should definitely not do that, very scary - `^` is fine for peers, we want to control "at least x version" and `^` does that (within highest set version range) so `^0.1.0` allows `0.1.x`, but not `0.2.x` or `1.x.x` etc - or with major version `^1.0.0` allows `1.x.x` but not `2.x.x` (same for patch, too technically but there any higher version will be not be allowed (since it then only acts on the patch version)

jnsdls: tl;dr it should be:

### adapter
- hard = js deps with pinned version
- peer = mandatory native deps with `^`

### main
- peer = adapter (workspace version, will get auto pinned on release) + mandatory native deps with `^` (in sync with adapter) -- all of these optional
- dev = same thing as peer

joaquim-verges: i think id rather only set the RN peer deps in the react-native-adapter - dont want bundlers to eagerly install these unless you're explicitely installing the RN adapter.

for the peer version, I basically want "anything above this version is fine" - I dont want to block ppl upgrading a minor/major if the API that we use internally is non breaking

jnsdls: package managers will install dev dependencies regardless of what you do

jnsdls: you overestimate the ecosystem ;)

jnsdls: if that's what you want then use `>=` for the version, but never *

jnsdls (approved):
LGTM!

PR #3082, opened 2024-05-20
Add "Switch Account" button in ConnectButton for injected metamask provider

MananTank:
## PR-Codex overview
This PR adds a switch account button for MetaMask in the `ConnectButton` details section.

### Detailed summary
- Added `SwitchMetamaskAccount` component to allow switching MetaMask accounts
- Imported `ShuffleIcon` and `TextAlignJustifyIcon`
- Imported `injectedProvider` from `mipdStore.js`
- Implemented logic to switch MetaMask account and request permissions


packages/thirdweb/src/react/web/ui/ConnectWallet/Details.tsx
```diff
@@ -983,3 +987,31 @@ function SwitchNetworkButton(props: {
     </Button>
   );
 }
+
+function SwitchMetamaskAccount(props: {
+  closeModal: () => void;
+}) {
+  const wallet = useActiveWallet();
+  const connectLocale = useConnectUI().connectLocale;
+  const injectedMetamaskProvider = injectedProvider("io.metamask");
```
jnsdls, line 996: probably better to only run this once we know the wallet id, and then do another level of check for this -> that way this doesn't run if the activeWallet isn't metamask first

jnsdls (changes requested):
small change

packages/thirdweb/src/react/web/ui/ConnectWallet/Details.tsx
MananTank: Done

jnsdls (approved):

PR #3139, opened 2024-05-23
`signTypedData`: throw error for incorrectly typed chainId

gregfromstl:
## Problem solved

The v4 SDK uses a hex for chain ID when signing typed data, but v5 expects a number. This can lead to difficult-to-debug errors for JS users. This adds an explicit error message for this case and a helper function to add additional type-checking of typed data, which has fairly complex and confusing type definitions.

---

## PR-Codex overview
This PR focuses on converting `chainId` from hex to number for proper `signTypedData` type. 

### Detailed summary
- Converted `chainId` hex to number in `signTypedData` functions across multiple files
- Added `parseTypedData` function to handle the conversion
- Implemented the conversion for proper signTypedData type


packages/thirdweb/src/utils/signatures/helpers/parseTypedData.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export function parseTypedData<
+  typedData extends TypedData | Record<string, unknown> = TypedData,
+  primaryType extends keyof typedData | "EIP712Domain" = keyof typedData,
+>(
+  typedData: TypedDataDefinition<typedData, primaryType>,
+): TypedDataDefinition<typedData, primaryType> {
+  const domain = typedData.domain as unknown & { chainId?: unknown }; // TODO: create our own typed data types so this is cleaner
+  if (domain?.chainId !== undefined && isHex(domain.chainId)) {
+    throw new Error("signTypedData: chainId must be of type number");
```
joaquim-verges, line 16: can we not transform it ourselves to a number?

gregfromstl: We can't because this type is _very_ particular and it's a combination of viem types and abitype so there's no way to get around it until we create our own from the ground up (which we should do just afraid to try it now)

joaquim-verges: ok

joaquim-verges (approved):

PR #3140, opened 2024-05-24
Allow exporting in-app wallets

gregfromstl:
Show in-app wallet users an option to export their private key

### Usage

```tsx
import { inAppWallet } from "thirdweb/wallets";
const wallet = inAppWallet({
  showPrivateKeyExport: true,
});
```

https://github.com/thirdweb-dev/js/assets/17715009/9b34027e-b619-4c54-be10-e6d9243b1e71

---

## PR-Codex overview
This PR introduces the ability to hide private key export in the in-app wallet, with UI changes and new components added.

### Detailed summary
- Added `hidePrivateKeyExport` option in in-app wallet
- Added `KeyIcon` component for private key export
- Added `PrivateKey` screen for exporting private key
- Updated UI for private key export option in wallet details menu


.changeset/shaggy-mice-jump.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+Show in-app wallet users an option to export their private key
+
+### Usage
+
+```tsx
+import { inAppWallet } from "thirdweb/wallets";
+const wallet = inAppWallet({
+  showPrivateKeyExport: true,
```
joaquim-verges, line 12: It's a bit odd to out this option here since it only applies to the UI components.

Using in app wallet with your own UI would make this option do nothing at all.

I know we have the auth options in there too which also don't really make sense there. Feels like we should move those options to the connect button/embed itself. Would only apply is inAppWallet is in the wallet list, but def feels more appropriate there. Wdyt?

packages/thirdweb/src/react/web/ui/ConnectWallet/Details.tsx
```diff
@@ -458,7 +462,6 @@ export const ConnectedWalletDetails: React.FC<{
       </Container>
       <Spacer y="md" />
       <Container px="md">
-        {/* Network Switcher */}
```
joaquim-verges, line 461: These are useful, would leave them

packages/thirdweb/src/wallets/in-app/core/wallet/types.ts
```diff
@@ -41,6 +41,7 @@ export type InAppWalletCreationOptions =
         };
       };
       smartAccount?: SmartWalletOptions;
+      showPrivateKeyExport?: boolean;
```
joaquim-verges, line 44: Wonder if we should make this true by default.

Also can you post a screenshot or better enable it in the playground?

gregfromstl: Agree. Changed it to `hidePrivateKeyExport` and added a demo video to the PR description

.changeset/shaggy-mice-jump.md
gregfromstl: I don't love the idea of adding it to the connect button or embed props given there are already so many of those and they're very hard to discover/understand. At the same time, I agree it's a bit odd here for someone not using the UI.

Maybe we should enable it in our UI and not provide the option to disable it? If someone wants to change our defaults, they should build their own UI (and we should provide more low-level components to make that easier).

joaquim-verges (commented):
Shows dark in light mode too for me

gregfromstl:
> Shows dark in light mode too for me
> 
> 

This should be fixed now @joaquim-verges

joaquim-verges (approved):
Looks great!!

PR #3141, opened 2024-05-24
[Ecosystem Wallets] Adds the ability to connect to an ecosystem wallet with an integrator ID

gregfromstl:
```ts
import { inAppWallet } from "thirdweb/wallets";
const wallet = inAppWallet({
  integratorId: "...",
});
```


joaquim-verges (approved):
remember to pass it to the native connector too, will likely need to pass through some api's in there

PR #3190, opened 2024-05-31
Fix AutoConnect on MIPD wallets

gregfromstl:
## Problem solved

AutoConnect was only using user-specified wallets, meaning installed but not specified wallets would not reconnect on refresh

---

## PR-Codex overview
The focus of this PR is to update the `useAutoConnect` hook to use available wallets from installed providers for auto connection.

### Detailed summary
- Added import for `createWallet` and `getInstalledWalletProviders`
- Updated `useAutoConnect` hook to use available wallets from installed providers
- Modified logic to find active wallet using the updated list of available wallets


joaquim-verges (approved):
good catch

PR #3198, opened 2024-06-01
Export `toSerializableTransaction`, `getGasPrice`, and `estimateGasCost`

gregfromstl:
## PR-Codex overview
The focus of this PR is to enhance the `thirdweb` package by adding functions to estimate gas cost, get gas price, and convert transactions to serializable format.

### Detailed summary
- Added `estimateGasCost` function to estimate gas cost in ether and wei
- Added `getGasPrice` function to get gas price for a chain
- Updated `toSerializableTransactionOptions` to `ToSerializableTransactionOptions`
- Improved documentation and examples for better understanding


packages/thirdweb/src/transaction/actions/to-serializable-transaction.ts
```diff
@@ -6,17 +6,44 @@ import type { PreparedTransaction } from "../prepare-transaction.js";
...
+ *   transaction,
+ * });
+ *
+ * account.sendTransaction(serializableTransaction);
+ * ```
+ * @transaction
  */
 export async function toSerializableTransaction(
-  options: ToSerialiableTransactionOptions,
+  options: ToSerializableTransactionOptions,
```
joaquim-verges, line 46: wonder if we should rename this to `populateTransaction` ? kind ethers-y but it kinda does make sense since that's what its doing, filling in the blanks. wdyt?

gregfromstl: Works for me 👍 
Going to keep the internal name `toSerializable` since it matches the viem flow we're using and will export as `populateTransaction` for anyone looking to get the final tx params

joaquim-verges: nah we should keep the same name, its gonna look messed up in the docs otherwise I think. lets just keep toSerializableTransaction - i'll make the change for you

joaquim-verges (approved):

PR #3230, opened 2024-06-04
[WIP] Allow specifying cost of transaction in ERC20 tokens

MananTank:
## PR-Codex overview
This PR adds support for specifying transaction cost in ERC20 tokens, enabling users to pay with ERC20 tokens when lacking specified ones.

### Detailed summary
- Added `valueERC20` option in `PreparedTransaction` object
- Updated `BuyForTx` type to include token details
- Modified functions to handle ERC20 token cost
- Updated UI components to display ERC20 token information

### DIFF
--- a/.changeset/wild-snakes-deliver.md
+++ b/.changeset/wild-snakes-deliver.md
@@ -2,27 +2,23 @@
 "thirdweb": minor
 ---
 
-Add `overrideTxCost` option in `payModal` to override the calculated transaction cost to add support for prompting the user to buy ERC-20 tokens via Pay Modal.
+Add `valueERC20` option in `PreparedTransaction` object to specify the transaction cost in ERC20 tokens to allow opening the Pay Modal for ERC20 tokens if user does not have specified ERC20 tokens.
 
 ```tsx
-const sendTx = useSendTransaction({
-  payModal: {
-    // set the transaction cost as 10 USDC
-    overrideTxCost: {
-      value: sendUSDPolygonAmount,
-      token: {
-        address: "0x3c499c542cEF5E3811e1192ce70d8cC03d5c3359",
-        name: "USD Coin",
-        symbol: "USDC",
-      },
-    },
-  },
-});
+const sendTx = useSendTransaction();
 
 return (
   <button
     onClick={async () => {
-      await sendTx.mutateAsync({ ... });
+      const someTx = { ... };
+
+      // declare that transaction cost is 10 USDC
+      someTx.valueERC20 = {
+        amount: "10",
+        tokenAddress: "0x3c499c542cEF5E3811e1192ce70d8cC03d5c3359",
+      };
+
+      await sendTx.mutateAsync(someTx);
     }}
   >
     send tx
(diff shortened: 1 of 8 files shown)


.changeset/wild-snakes-deliver.md
```diff
@@ -0,0 +1,45 @@
...
+```tsx
+const sendTx = useSendTransaction();
+
+return (
+  <button
+    onClick={async () => {
+      const someTx = { ... };
+
+      // declare that transaction cost is 10 USDC
+      someTx.valueERC20 = {
```
joaquim-verges, line 16: can you actually mutate the transaction like that? i thought fields were read only

packages/thirdweb/src/react/core/hooks/contract/useSendTransaction.ts
```diff
@@ -20,7 +21,7 @@ type ShowModalData = {
   tx: PreparedTransaction;
   sendTx: () => void;
   rejectTx: () => void;
-  totalCostWei: bigint;
+  cost: string;
```
joaquim-verges, line 24: what's the unit here? also do we still need this?

packages/thirdweb/src/react/core/hooks/contract/useSendTransaction.ts
```diff
@@ -97,34 +98,66 @@ export function useSendTransactionCore(
...
-            const walletBalanceWei = walletBalance.value;
-
-            // if enough balance, send tx
-            if (totalCostWei < walletBalanceWei) {
-              sendTx();
-              return;
+                tokenAddress: tx.valueERC20.tokenAddress,
+              });
+
+              // if enough balance, send tx
```
joaquim-verges, line 141: we might need to check approval too maybe?

what's cool with this API is that we can automatically prompt for approval if needed, for any transaction

packages/thirdweb/src/transaction/prepare-transaction.ts
```diff
@@ -74,6 +74,11 @@ export type PreparedTransaction<
 > = Readonly<options> & {
   __preparedMethod?: () => Promise<PreparedMethod<abiFn>>;
   __contract?: ThirdwebContract<abi>;
+  // additional
+  valueERC20?: {
+    amount: string; // units of tokens
+    tokenAddress: string;
+  };
```
joaquim-verges, line 81: shouldn't this be in `PrepareTransactionOptions` ?

joaquim-verges (commented):
Can you add an example in the pay playground of this? I can add a token in the example drop contract to require usdc for example

packages/thirdweb/src/react/core/hooks/contract/useSendTransaction.ts
MananTank: It feels weird to send an approval transaction automatically before the given transaction. UX would be confusing for mobile users because they would be prompted to do 2 transactions one after another which would suddenly send them to wallet app. I think we should leave that up to developer so they can design the 2-step UI ( similar to what we have in Pay UI )

packages/thirdweb/src/react/core/hooks/contract/useSendTransaction.ts
MananTank: unit is number of tokens. This object is only used internally - this is not public API

.changeset/wild-snakes-deliver.md
MananTank: Yes, that's why it's added on `PreparedTransaction` and not on `PrepareTransactionOptions`

packages/thirdweb/src/transaction/prepare-transaction.ts
MananTank: Added on `PreparedTransaction` to allow mutating them to set them after getting tx object from extension

joaquim-verges: this pattern doesnt feel right. 

we should just add it as optional inputs to prepareTransaction / prepareContractCall and the overrides for extensions. Same way as you pass native value currently.

MananTank:
We'll be going with another method for this - closing it

PR #3237, opened 2024-06-06
fix: resolves viem#2306

jxom:
Fixes issue related to: https://github.com/wevm/viem/issues/2306

---

## PR-Codex overview
This PR updates various files in `thirdweb` package related to Ethereum adapters, wallet functions, package dependencies, and test cases.

### Detailed summary
- Updated `ethers6.ts` and `ethers5.ts` to include `accessList` property
- Updated `package.json` with new versions for dependencies
- Modified test cases to run based on a secret key
- Refactored wallet functions to use `serializeTypedData`
- Updated dependencies and resolved peer dependency issues

> The following files were skipped due to too many changes: `packages/thirdweb/src/extensions/marketplace/english-auctions/english-auctions.test.ts`, `packages/thirdweb/src/extensions/marketplace/direct-listings/direct-listings.test.ts`

### DIFF
--- a/packages/thirdweb/package.json
+++ b/packages/thirdweb/package.json
@@ -197,7 +197,7 @@
     "mipd": "0.0.7",
     "node-libs-browser": "2.2.1",
     "uqr": "0.1.2",
-    "viem": "2.10.9"
+    "viem": "2.13.7"
   },
   "peerDependencies": {
     "@aws-sdk/client-lambda": "^3",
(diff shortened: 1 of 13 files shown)


joaquim-verges:
thanks for this @jxom ! Not lowercasing strings types makes sense, but I'm actually surprised we need this lowercasing of address types in the first place, would have expected checksum instead.

Can you point me to some resources to understand why addresses need to be lowercased before sending with `eth_signTypedData`? Reason I'm asking is because we might have the same bug in Unity/dotnet, want to make sure we do the right thing there.

jxom:
Not all Wallets behave correctly when they are passed a checksum address. Can’t remember off the top of my head which Wallets didn’t behave correctly. This behavior exists in Ethers as well.

PR #3248, opened 2024-06-08
Add cross-platform support for crypto in randomBytesBuffer function

sunsetlover36:
## Problem solved

This function is used in the `generatePayload` method when generating the payload in the login route to implement SIWE authorization on the server side. Currently, the function only works in the browser, but this fix will allow it to execute on the server as well, enabling it to generate the login payload without errors.

### How to test

Just call `thirdwebAuth.generatePayload` method and you'll get an error without the fix:
```js
TypeError: Cannot read properties of undefined (reading 'getRandomValues')
```

If the fix is applied, there will be no error.

Ref: https://portal.thirdweb.com/connect/auth/frameworks/react-express

## Changes made

## How to test

## Contributor NFT

Paste in your wallet address below and we will airdrop you a special NFT when your pull request is merged. 

```Address: 0x6047ec2435C5906241AeAcC13b09D3c0eb09Cb45```

---

## PR-Codex overview
The focus of this PR is to convert synchronous operations to asynchronous in the authentication and random generation processes.

### Detailed summary
- Converted `randomBytesHex` and `randomBytesBuffer` functions to be asynchronous.
- Updated functions in `cognitoAuth.ts` and `fetchers.ts` to use `await` for `randomBytesHex`.
- Modified functions in various files to await `getSessionHeaders` calls for nonce generation.

### DIFF
--- a/packages/thirdweb/src/auth/core/generate-login-payload.ts
+++ b/packages/thirdweb/src/auth/core/generate-login-payload.ts
@@ -35,10 +35,9 @@ export function generateLoginPayload(options: AuthOptions) {
       // use the user passed nonce generator or fall back to generationg uuid
       nonce: await (
         options.login?.nonce?.generate ||
-        (() =>
-          import("../../utils/random.js").then(({ randomBytesHex }) =>
-            randomBytesHex(),
-          ))
+        (() => import("../../utils/random.js").then(({ randomBytesHex }) =>
+          randomBytesHex()
+        ))
       )(),
       statement: options.login?.statement || DEFAULT_LOGIN_STATEMENT,
       version: options.login?.version || DEFAULT_LOGIN_VERSION,
--- a/packages/thirdweb/src/utils/random.ts
+++ b/packages/thirdweb/src/utils/random.ts
@@ -11,5 +11,7 @@ export function randomBytesHex(length = 32) {
  * @internal
  */
 export function randomBytesBuffer(length = 32) {
-  return globalThis.crypto.getRandomValues(new Uint8Array(length));
+  return (typeof globalThis.crypto !== 'undefined'
+    ? globalThis.crypto
+    : (await import('crypto')).webcrypto).getRandomValues(new Uint8Array(length));
 }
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by sunsetlover36** [comment]
Also, there is a workaround to fix it without this PR, but it needs to be specified in the documentation, I think. It's not type-safe though.

```js
import { webcrypto } from 'crypto';

globalThis.crypto = webcrypto;
```

joaquim-verges:
thank you for the contribution! just need to double check this doesnt break anything with the metro bundler in react native. will report back

jnsdls:
one interesting note here, starting in `node20` this should no longer be required: https://nodejs.org/docs/latest-v20.x/api/webcrypto.html

jnsdls:
> one interesting note here, starting in `node20` this should no longer be required: https://nodejs.org/docs/latest-v20.x/api/webcrypto.html

@sunsetlover36 I assume you're using node18 in a scenario where this breaks?

jnsdls:
given [the nodejs release cycle](https://nodejs.org/en/about/previous-releases) I wonder if there is a way to do this in a more polyfill-for-node-18 way, we're basically just _slightly_ ahead of when we could ignore this problem and tell people "node 20+ only please"

sunsetlover36:
> > one interesting note here, starting in `node20` this should no longer be required: https://nodejs.org/docs/latest-v20.x/api/webcrypto.html
> 
> @sunsetlover36 I assume you're using node18 in a scenario where this breaks?

Yes, I was using version 18.19.0. I tried with a newer version, and it actually works. I wasn't aware of that. With Node 20+ the issue should be resolved.

jnsdls:
thanks for confirming! - I think it this case it might be good enough to advise folks on node 18 to polyfill `globalThis.crypto = require("crypto").webCrypto` in this case. I'm a little worried about the consequences to bundlers etc across the stack (in frontend environments mostly) that might up this path and break otherwise.

joaquim-verges:
Going to close this for now and add the note about the node 18 polyfill to the docs

PR #3264, opened 2024-06-10
Wallet connection manager fixes

MananTank:
## PR-Codex overview
This PR removes the feature that automatically sets another connected wallet as active when disconnecting the current active wallet. It also stops saving the personal wallet separately in the connected wallets list.

### Detailed summary
- Removed feature to set another connected wallet as active upon disconnecting current active wallet
- Personal wallet no longer saved separately in connected wallets list


packages/thirdweb/src/wallets/manager/index.ts
```diff
@@ -121,9 +114,6 @@ export function createConnectionManager(storage: AsyncStorage) {
 
     handleSetActiveWallet(activeWallet);
 
-    // add personal wallet to connected wallets list
-    addConnectedWallet(personalWallet);
```
joaquim-verges, line 125: I think we should still add the EOA as a connected wallet. What's the reasoning for removing this?

MananTank: reverted this change

gregfromstl (approved):

Issue #3320, opened 2024-06-14
 Fixing Popup Handling Issue in Coinbase Smart Wallet Integration


bablu573:
When users attempt to process transactions, they get stuck on the "waiting for confirmation" screen due to popup handling issues. This primarily occurs when popups are blocked or not managed properly.

This was not an issue when using RainbowKit. Here are two websites to test:
- https://thirdweb-demo-liard.vercel.app/
- https://my-rainbowkit-app-xi.vercel.app/
<img width="360" alt="Screenshot 2024-06-15 at 12 55 27 AM" src="https://github.com/thirdweb-dev/js/assets/100342157/dc868332-12a6-4a41-b0c2-ef7d36b86ad9">

Please check using Safari with popups disabled.

ThirdWeb SDK version: 5.29.1

joaquim-verges:
@bablu573 thanks for reporting this! I think I know what's going on, will fix asap

bablu573:
Hey @joaquim-verges  when to expect the release

joaquim-verges:
@bablu573 PR #3330 fixes the issue.

You can test it right now by installing the dev version from the PR

```
npm i thirdweb@0.0.0-dev-906c9785e323ac0b10f42d8888e28188f8e7353b-20240615094612
```

worked well for me, but let me know if it works for you and we'll get it released to stable asap.

bablu573:
Its not working, can you try it on safari mobile and after connecting wallet getting error here are images and git repo 
https://github.com/AvinashNayak27/thirdweb-demo
https://thirdweb-demo-liard.vercel.app/

<img width="360" alt="Screenshot 2024-06-15 at 3 35 05 PM" src="https://github.com/thirdweb-dev/js/assets/100342157/c5462ac3-9994-498a-8c5d-2c463e7e132d">
<img width="360" alt="Screenshot 2024-06-15 at 3 35 23 PM" src="https://github.com/thirdweb-dev/js/assets/100342157/a7cf09ee-f56a-4236-90bb-cced07c89ea2">

jnsdls:
yep, confirming I see the same.

we're actively tackling this, however it might take a day or two, we'll keep updating this issue as we have builds to try

jnsdls:
https://github.com/thirdweb-dev/js/pull/3334 will solve the failed connection screen issue

[...]

PR #3330, opened 2024-06-15
feat: preloading wallet providers for improved performance

joaquim-verges:
## PR-Codex overview
This PR improves the performance and fixes issues with the Coinbase wallet popup not opening on Safari. 

### Detailed summary
- Added a function to preload wallet providers for performance
- Enhanced the ConnectModalContent component to handle immediate connection for Coinbase wallet

### DIFF
--- a/packages/thirdweb/src/react/web/ui/ConnectWallet/Modal/ConnectModalContent.tsx
+++ b/packages/thirdweb/src/react/web/ui/ConnectWallet/Modal/ConnectModalContent.tsx
@@ -1,5 +1,6 @@
 "use client";
 import { Suspense, lazy, useCallback } from "react";
+import { COINBASE } from "../../../../../wallets/constants.js";
 import type { Wallet } from "../../../../../wallets/interfaces/wallet.js";
 import { useSiweAuth } from "../../../../core/hooks/auth/useSiweAuth.js";
 import { useConnectUI } from "../../../../core/hooks/others/useWalletConnectionCtx.js";
@@ -41,6 +42,7 @@ export const ConnectModalContent = (props: {
     connectModal,
     connectLocale,
     client,
+    chain,
   } = useConnectUI();
   const setActiveWallet = useSetActiveWallet();
   const setSelectionData = useSetSelectionData();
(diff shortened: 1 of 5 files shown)


joaquim-verges:
* **#3330** <a href="https://app.graphite.dev/github/pr/thirdweb-dev/js/3330?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @joaquim-verges and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

joaquim-verges:
fixes #3320

packages/thirdweb/src/react/web/utils/usePreloadWalletProviders.ts
```diff
@@ -0,0 +1,48 @@
+import { useEffect } from "react";
+import type { ThirdwebClient } from "../../../client/client.js";
+import { COINBASE } from "../../../wallets/constants.js";
+import type { Wallet } from "../../../wallets/interfaces/wallet.js";
+
+export const usePreloadWalletProviders = ({
+  client,
+  wallets,
+}: { client: ThirdwebClient; wallets: Wallet[] }) => {
+  useEffect(() => {
```
jnsdls, line 10: this will potentially fire multiple times, need to use `useQuery` here - I'll submit a pr on top

packages/thirdweb/src/react/web/ui/ConnectWallet/Modal/ConnectModalContent.tsx
```diff
@@ -94,7 +96,26 @@ export const ConnectModalContent = (props: {
...
+        // this fixes an issue on safari where the CB popup window would get blocked
+        // because of the delay between this click and the actual connection
+        // that happens within 'CoinbaseSDKConnection' component 2 layers down
+        // when the cb extension is installed -> pops the extension
+        // when the extension is not installed -> pops a new window
+        // then in 'CoinbaseSDKConnection' the connection is requested again
+        // which just "links" to the extension / popup window a few ms later
+        // and handle the completion / error handling there
+        // this connection flow structure needs to be refactored to properly handle this
+        if (newWallet.id === COINBASE) {
```
jnsdls, line 112: does this properly set up all the listeners etc?

packages/thirdweb/src/react/web/utils/usePreloadWalletProviders.ts
jnsdls: addressed in #3331

packages/thirdweb/src/react/web/ui/ConnectWallet/Modal/ConnectModalContent.tsx
jnsdls: this comes with more problems, here's what I've found so far:

1. coinbase wallet will now call connect twice (here, and in the useEffect below)
2. this causes that "second" connection to *always* fail -> we always end up in an error screen

to resolve this we really have 2 options:
1. refactor this behavior so *all* wallets connect on-click (if they have distinct connect behaviors that can be known at click-time) -> then pass the wallet instance that is connecting down to the screen, which can subscribe to the `onConnect` listener (we also likely need to expose an `onConnectionError` listener to handle that case) -> that screen becomes purely the "retry" and "visualize the state" screen instead of ever triggering a connection
2. add an explicit button to that screen to trigger the connection, means every user has to make an extra click but is likely an easier refactor - again we'd want to do this for every wallet not just coinbase

jnsdls (changes requested):
read existing comments, in the current state this will cause more problems than it solves

packages/thirdweb/src/react/web/ui/ConnectWallet/Modal/ConnectModalContent.tsx
jnsdls: solved upstack

jnsdls (approved):
approved based on changes higher up in stack

PR #3338, opened 2024-06-15
chore: Update chain typo types.ts

wellitongervickas:
## Problem solved

Fixed type "true" instead of "boolean"

---

## PR-Codex overview
This PR updates the `testnet` property in the `types.ts` file from a hardcoded `true` value to a boolean type.

### Detailed summary
- Updated `testnet` property in `types.ts` to use a boolean type instead of hardcoded `true`.


jnsdls:
> @jnsdls you are correct. I'm trying to reuse a few things that is on the codebase like viem chains, I need to force a type "true" to handle these cases when try to reuse same code. Thx for asking

Could you share an example that currently doesn't work and will work with this change?

wellitongervickas:
@jnsdls  I think would be very good if community make some pattern since there's almost same data from each chain. 
```
import { baseSepoliaChain, baseChain, Chain } from './chains'; /// wagmi chains  +  additional dapp prop

export const mainnets = [baseChain];
export const testnets = [baseSepoliaChain];

export const allowedChains = (
  process.env.NEXT_PUBLIC_TESTNET_MODE === `1` ? testnets : mainnets
) as [Chain, ...Chain[]];

export const thirdwebChains = allowedChains.map((chain) => ({
  name: chain.name,
  id: chain.id,
  rpc: chain.rpcUrls.default.http[0],
  nativeCurrency: chain.nativeCurrency,
  testnet: chain.testnet as true, /// here weird approach
  blockExplorers: chain.blockExplorers?.default
    ? [
        {
          name: chain.blockExplorers.default.name!,
          url: chain.blockExplorers.default.url!,
          apiUrl: chain.blockExplorers.default.apiUrl!,
        },
      ]
    : [],
}));
```

wellitongervickas:
@jnsdls Thirdweb says its work "side by side" with [viem](https://github.com/wevm/viem/tree/main/src/chains). So why not enjoy their code as well? I would love to hear from you about that

jnsdls:
makes sense, this _should_ today if you pass the chains from `viem` or `wagmi` into `defineChain(<chain here>)` from `thirdweb`

so I'd expect something like this to work, please let me know if it doesn't:

```ts
import { defineChain } from "thirdweb";
import { someChain } from "viem/chains"; // or wagmi, I believe the definitions are shared(?)

// transform `viem` chain definition into tw chain
const twChain = defineChain(someChain);

// then use the chain with thirdweb below
```

that said I agree with your overall sentiment of there needing to be more of a standard, in general one pain point I feel is that chain definitions are kind of all over the place, now obviously we're somewhat guilty of that here as well...

[chain definitions in a nutshell](https://xkcd.com/927/)

I don't yet know what to do about it. In the meantime please let me know if the above approach works for you!

PR #3347, opened 2024-06-17
Add ERC721 extension: updateMetadata & updateBatchBaseURI

kien-ngo:
## PR-Codex overview
This PR adds an ERC721 extension `updateMetadata` for updating NFT metadata in a batch. 

### Detailed summary
- Added `updateMetadata` function to update NFT metadata in ERC721 extension
- Added `getUpdateMetadataParams` function to prepare update metadata parameters
- Added tests for `updateMetadata` function with different scenarios
- Added `FreezeBatchBaseURIParams` and related functions for freezing batch base URI

> The following files were skipped due to too many changes: `packages/thirdweb/src/extensions/erc721/__generated__/DropERC721/write/freezeBatchBaseURI.ts`, `packages/thirdweb/src/extensions/erc721/__generated__/DropERC721/write/setMaxTotalSupply.ts`, `packages/thirdweb/src/extensions/erc721/__generated__/DropERC721/write/updateBatchBaseURI.ts`

### DIFF
--- a/packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.test.ts
+++ b/packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.test.ts
@@ -1,0 +1,127 @@
+import { beforeEach, describe, expect, it } from "vitest";
+import { ANVIL_CHAIN } from "~test/chains.js";
+import { TEST_CLIENT } from "~test/test-clients.js";
+import { TEST_ACCOUNT_A } from "~test/test-wallets.js";
+
+import {
+  type ThirdwebContract,
+  getContract,
+} from "../../../../contract/contract.js";
+import { deployERC721Contract } from "../../../../extensions/prebuilts/deploy-erc721.js";
+import { sendAndConfirmTransaction } from "../../../../transaction/actions/send-and-confirm-transaction.js";
+import type { NFTInput } from "../../../../utils/nft/parseNft.js";
+import { getNFT } from "../../read/getNFT.js";
+import { getNFTs } from "../../read/getNFTs.js";
+import { lazyMint } from "../../write/lazyMint.js";
+import { getUpdateMetadataParams, updateMetadata } from "./updateMetadata.js";
+
+let contract: ThirdwebContract;
+const account = TEST_ACCOUNT_A;
+const client = TEST_CLIENT;
+const chain = ANVIL_CHAIN;
+
+describe("erc721: updateMetadata", () => {
+  beforeEach(async () => {
+    const address = await deployERC721Contract({
+      client,
+      chain,
+      account,
+      type: "DropERC721",
+      params: {
+        name: "NFTDrop",
+      },
+    });
+    contract = getContract({
+      address,
+      client,
+      chain,
+    });
+  });
+
+  it("should update the metadata for the selected token | case 1", async () => {
+    const metadatas: NFTInput[] = [
+      { name: "token 0" },
+      { name: "token 1" },
+      { name: "token 2" },
+    ];
+
+    const lazyMintTx = lazyMint({ contract, nfts: metadatas });
+    await sendAndConfirmTransaction({ transaction: lazyMintTx, account });
+
+    const transaction = updateMetadata({
+      contract,
+      targetTokenId: 0n,
+      newMetadata: { name: "token 0 - updated" },
+      client,
+    });
+    await sendAndConfirmTransaction({ transaction, account });
+    const nfts = await getNFTs({ contract });
+    expect(nfts[0]?.metadata.name).toBe("token 0 - updated");
+
+    // The other tokens should not change
+    expect(nfts[1]?.metadata.name).toBe("token 1");
+    expect(nfts[2]?.metadata.name).toBe("token 2");
+  });
+
+  it("should update the metadata for the selected token | case 2", async () => {
+    const nftBatch1: NFTInput[] = [
+      { name: "token 0" },
+      { name: "token 1" },
+      { name: "token 2" },
+    ];
+
+    const batch1 = lazyMint({ contract, nfts: nftBatch1 });
+    await sendAndConfirmTransaction({ transaction: batch1, account });
+
+    const nftBatch2: NFTInput[] = [
+      { name: "token 3" },
+      { name: "token 4" },
+      { name: "token 5" },
+    ];
+    const batch2 = lazyMint({ contract, nfts: nftBatch2 });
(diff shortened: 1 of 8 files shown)


packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      { name: "token 2" },
+    ];
+
+    const lazyMintTx = lazyMint({ contract, nfts: metadatas });
+    await sendAndConfirmTransaction({ transaction: lazyMintTx, account });
+
+    const transaction = await updateMetadata({
+      contract,
+      targetTokenId: 0n,
+      newMetadata: { name: "token 0 - updated" },
```
joaquim-verges, line 53: nice! can you also add a test that does 2 batches of lazy mint, and updates a token in one of them, then another in the other batch?

packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.ts
```diff
@@ -0,0 +1,93 @@
...
+import type { NFT, NFTInput } from "../../../../utils/nft/parseNft.js";
+import { getBaseURICount } from "../../__generated__/IBatchMintMetadata/read/getBaseURICount.js";
+
+export type UpdateMetadataParams = {
+  targetTokenId: bigint;
+  newMetadata: NFTInput;
+  client: ThirdwebClient;
+};
+
+export async function updateMetadata(
```
joaquim-verges, line 12: this function should only 'prepare' the transaction so it should not be `async`

instead, all the work should happen inside the `params` async function (yes it can be a function!)

the idea is that you defer the heavy execution to when the user actually tries to send the transaction, making 'preparing' this transaction very cheap

packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  // This can be replaced with `updateBatchBaseURI()` once we add that extension
+  const { prepareContractCall } = await import(
+    "../../../../transaction/prepare-contract-call.js"
+  );
+
+  return prepareContractCall({
+    contract,
+    method: "function updateBatchBaseURI(uint256 _index, string calldata _uri)",
+    params: [batchIndex, _uri],
```
joaquim-verges, line 91: make this a async function and move all the logic down here

packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.test.ts
kien-ngo: Oh yeah for sure!

packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  return [batchIndex, _uri];
+}
+
+/**
+ * Update the metadata of the single token in an NFT Drop (DropERC721) collection
+ * For NFT Collection, please use `setTokenURI`
+ * @param options
+ * @returns the prepared transaction
+ * @extension ERC721
```
jnsdls, line 95: find the ABI for it, throw it in the relevant file in `scripts/generate/abis` and run `pnpm build:generate`

packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.ts
```diff
@@ -0,0 +1,123 @@
...
+  const [{ getBatchIdAtIndex }, { getNFT }] = await Promise.all([
+    import("../../__generated__/IBatchMintMetadata/read/getBatchIdAtIndex.js"),
+    import("../../read/getNFT.js"),
+  ]);
+  let startTokenId = 0n;
+  let endTokenId = 0n;
+  let batchIndex = 0n;
+  for (let i = 0n; i < batchCount; i += 1n) {
+    batchIndex = i;
+    endTokenId = await getBatchIdAtIndex({ contract, index: batchIndex });
```
jnsdls, line 37: this will block on every iteration, any way to do this in parallel?

packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.ts
```diff
@@ -0,0 +1,123 @@
...
+    startTokenId = endTokenId;
+  }
+
+  const range = Array.from(
+    { length: Number(endTokenId - startTokenId) },
+    (_, k) => BigInt(k) + startTokenId,
+  );
+
+  const currentMetadatas = await Promise.all(
+    range.map((id) => getNFT({ contract, tokenId: id, includeOwner: false })),
```
jnsdls, line 50: I'm assuming the answer is "yes" but do we need to fetch the full NFT for each token here or can we instead simply fetch the `tokenURI`?

kien-ngo: Unfortunately we have to download all the content from IPFS, replace the items that we want to update, and then re-upload everything to IPFS again.

This is by-design from the smart contract. Need to replace the whole batch (even if all you want to update is 1 item)

packages/thirdweb/src/extensions/erc721/drops/write/updateMetadata.ts
kien-ngo: Yeah we can. I was thinking, what if there are 10,000 batches and the token ID we want to update is like, `0n`, then we are wasting resources.

But the chances of a drop collection having too many batches are low, so this "loop" way might provide not-too-bad performance, while saving on RPC requests.

wdyt @jnsdls

kien-ngo: So like, if I upload 10,000 NFTs at once, that's just 1 batch

kien-ngo: acutally i dont think i make sense anymore

jnsdls: I think I follow, that somewhat makes sense this seems to just be pretty un-ergonomic on the contract side

kien-ngo: Ty. (yeah i have brought up this contract design a few times now)

jnsdls (approved):
LGTM

PR #3451, opened 2024-06-22
[Feature] Add session transaction storage

gregfromstl:
https://linear.app/thirdweb/issue/CNCT-1398/ts-sdk-store-in-session-transactions-in-wallet

This update introduces the ability to track transactions during a "session" within the thirdweb SDK. The key additions and changes are:

1. Exporting `getSessionTransactions` from `transaction-store.js`.
2. Adding calls to `addTransactionToSession` in `send-gasless-transaction.ts` and `send-transaction.ts`.
3. Creating a new test file `transaction-store.test.ts` to cover the transaction session functionalities.
4. Including new constants in `settings.ts` specific to session storage keys.
5. Extending `LocalStorage` class functionalities to handle transaction session data.

These changes will act as the foundation of the v1 "transactions" page in the react SDK

---

---

## PR-Codex overview
This PR adds a `getTransactionStore` function to retrieve transaction stores by address and enhances transaction handling with a store.

### Detailed summary
- Added `getTransactionStore` function to retrieve transaction stores by address
- Implemented `addTransactionToStore` to store transactions for an address
- Updated `sendTransaction` to store transactions after sending
- Added tests for transaction store functionality

### DIFF
--- a/packages/thirdweb/src/exports/transaction.ts
+++ b/packages/thirdweb/src/exports/transaction.ts
@@ -55,6 +55,7 @@ export {
   toSerializableTransaction,
   type ToSerializableTransactionOptions,
 } from "../transaction/actions/to-serializable-transaction.js";
+export { getTransactionStore } from "../transaction/transaction-store.js";
 
 //types & utils
 export {
--- a/packages/thirdweb/src/transaction/transaction-store.ts
+++ b/packages/thirdweb/src/transaction/transaction-store.ts
@@ -1,0 +1,46 @@
+import { type Store, createStore } from "../reactive/store.js";
+import type { Hex } from "../utils/encoding/hex.js";
+
+type StoredTransaction = {
+  transactionHash: Hex;
+  chainId: number;
+};
+
+const transactionsByAddress = new Map<string, Store<StoredTransaction[]>>();
+
+/**
+ * Retrieve the transaction store for a given address.
+ * @param address - The address to retrieve the transaction store for.
+ * @returns A store of transactions for the given account to subscribe to.
+ */
+export function getTransactionStore(
+  address: string,
+): Store<StoredTransaction[]> {
+  const existingStore = transactionsByAddress.get(address);
+  if (existingStore) {
+    return existingStore;
+  }
+
+  const newStore = createStore<StoredTransaction[]>([]);
+  transactionsByAddress.set(address, newStore);
+  return newStore;
+}
+
+/**
+ * @internal
+ */
+export function addTransactionToStore(options: {
+  address: string;
+  transactionHash: Hex;
+  chainId: number;
+}) {
+  const { address, transactionHash, chainId } = options;
+  const tranasctionStore = getTransactionStore(address);
+
+  tranasctionStore.setValue([
+    ...tranasctionStore.getValue(),
+    { transactionHash, chainId },
+  ]);
+
+  transactionsByAddress.set(address, tranasctionStore);
+}
--- a/.changeset/cuddly-dots-fry.md
+++ b/.changeset/cuddly-dots-fry.md
@@ -1,0 +1,5 @@
+---
+"thirdweb": minor
+---
+
+Adds `getTransactionStore` to retrieve a transaction store for a given address
--- a/packages/thirdweb/src/transaction/actions/gasless/send-gasless-transaction.ts
+++ b/packages/thirdweb/src/transaction/actions/gasless/send-gasless-transaction.ts
@@ -1,6 +1,7 @@
 import type { TransactionSerializable } from "viem";
 import type { Account } from "../../../wallets/interfaces/wallet.js";
 import type { PreparedTransaction } from "../../prepare-transaction.js";
+import { addTransactionToStore } from "../../transaction-store.js";
 import type { WaitForReceiptOptions } from "../wait-for-tx-receipt.js";
 import type { GaslessOptions } from "./types.js";
 
@@ -32,12 +33,14 @@ export async function sendGaslessTransaction({
 
   // TODO: multiply gas by 2 for some reason(?) - we do in v4, *should* we?
 
+  let result: WaitForReceiptOptions | undefined;
+
   // biconomy
   if (gasless.provider === "biconomy") {
     const { relayBiconomyTransaction } = await import(
       "./providers/biconomy.js"
     );
-    return relayBiconomyTransaction({
+    result = await relayBiconomyTransaction({
       account,
       transaction,
       serializableTransaction,
@@ -50,7 +53,7 @@ export async function sendGaslessTransaction({
     const { relayOpenZeppelinTransaction } = await import(
       "./providers/openzeppelin.js"
     );
-    return relayOpenZeppelinTransaction({
+    result = await relayOpenZeppelinTransaction({
       account,
       transaction,
       serializableTransaction,
@@ -60,7 +63,7 @@ export async function sendGaslessTransaction({
 
   if (gasless.provider === "engine") {
     const { relayEngineTransaction } = await import("./providers/engine.js");
-    return relayEngineTransaction({
+    result = await relayEngineTransaction({
       account,
       transaction,
       serializableTransaction,
@@ -68,5 +71,13 @@ export async function sendGaslessTransaction({
     });
   }
 
-  throw new Error("Unsupported gasless provider");
+  if (!result) {
+    throw new Error("Unsupported gasless provider");
+  }
+  addTransactionToStore({
+    address: account.address,
+    transactionHash: result.transactionHash,
+    chainId: transaction.chain.id,
+  });
+  return result;
 }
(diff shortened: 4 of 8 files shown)


gregfromstl:
* **#3451** <a href="https://app.graphite.dev/github/pr/thirdweb-dev/js/3451?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @gregfromstl and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/thirdweb/src/transaction/transaction-store.ts
```diff
@@ -0,0 +1,106 @@
...
+import { waitForReceipt } from "./actions/wait-for-tx-receipt.js";
+import type { TransactionReceipt } from "./types.js";
+
+type StoredTransaction = {
+  transactionHash: Hex;
+  status: "PENDING" | "CONFIRMED";
+  receipt?: TransactionReceipt;
+};
+
+let storage: LocalStorage | undefined;
```
joaquim-verges, line 16: This should be asyncStorage ideally. Should work for both web and RN

joaquim-verges: we prob need to pass the storage in each of the functions below so it can be injected from the top

jnsdls: yeah this is actually a blocker.

packages/thirdweb/src/transaction/transaction-store.ts
```diff
@@ -0,0 +1,106 @@
...
+    ...(transactionsForAccount ?? []),
+    { transactionHash: transactionHash, status: "PENDING" },
+  ];
+  await setTransactionsInStorage(_storage, transactionStore);
+
+  waitForReceipt({
+    transactionHash,
+    chain,
+    client,
+  }).then(async (receipt) => {
```
jnsdls, line 65: this can throw, meaning you *have* to handle that case

packages/thirdweb/src/transaction/transaction-store.ts
```diff
@@ -0,0 +1,106 @@
...
+      ...(transactionsForAccount ?? []),
+      newTransaction,
+    ];
+
+    setTransactionsInStorage(_storage, transactionStore);
+  });
+}
+
+async function getTransactionsFromStorage(
+  storage: LocalStorage,
```
jnsdls, line 87: should be asyncstorage to be reusable with react native (same for all other instances of `LocalStorage`

jnsdls: also hm, we're calling it `session storage` but then we're saving to local storage? why? if its meant to be for the session then that should suffice?

worry here is that if we only ever add to it then at some point someone's local storage is going to run into the size limit -> we'll break the page

packages/thirdweb/src/transaction/transaction-store.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const transactionStore = await getTransactionsFromStorage(_storage);
+  const transactionsForAccount = transactionStore[account.address];
+  transactionStore[account.address] = [
+    ...(transactionsForAccount ?? []),
+    { transactionHash: transactionHash, status: "PENDING" },
+  ];
+  await setTransactionsInStorage(_storage, transactionStore);
+
+  waitForReceipt({
```
jnsdls, line 61: idea:
1. instead of doing this here simply _only_ store the tx hash
2. in the _list_ for each tx hash (from this list) retrieve the tx information (can probably use waitForReceipt, it _should?_ instantly resolve for tx that are already mined and ready?)

this way you only do this extra work when the user actually cares (they are looking at the tx) at the cost of having to show some loading state while u resolve (seems OK)

jnsdls: this also means you can simply "infer" the `state` of a tx, if u have a receipt it's mined, if not it's pending

jnsdls (changes requested):
some blockers and some ideas on how to improve it

packages/thirdweb/src/transaction/transaction-store.ts
gregfromstl: @joaquim-verges
Had a talk with Jonas about this, we're just gonna store in memory for now

packages/thirdweb/src/transaction/transaction-store.ts
```diff
@@ -0,0 +1,37 @@
+import type { Address } from "../utils/address.js";
+import type { Hex } from "../utils/encoding/hex.js";
+
+type StoredTransaction = {
+  transactionHash: Hex;
+  chainId: number;
+};
+
+const transactionStore = new Map<Address, StoredTransaction[]>();
```
joaquim-verges, line 9: Makes sense as a session, but UX wise isn't weird if you refresh and your tx is gone?

We can start with this but wonder if we should still store locally?

Maybe it's not a session to list but a 'local' tx list?

joaquim-verges: What if the API lets you pass a optional storage, and if not passed in stores in memory?

jnsdls: needs indexer for this

jnsdls, line 9: can you set this up to work as a "external store" out of the box:
- see: https://react.dev/reference/react/useSyncExternalStore
- other things we already have using this (example): `pendingSwapTx.ts` -> `createStore()`

it should work just fine in bare typescript, and then when we want to add this to react we can simply use `useSyncExternalStore` to hook into it (instead of having to figure out our own way to hook into it)

jnsdls: essentially you'll have to have a way to:
- get the current value of the store
- set the current value of the store (update)
- register listeners that get called when something updates the value

it should be relatively straight forward to make this work that way here

jnsdls (changes requested):

packages/thirdweb/src/transaction/transaction-store.ts
gregfromstl: Yeah it's definitely not perfect UX @joaquim-verges but I think the conclusion we landed on was to keep it really light right now and extend with optional storage if needed. Want to avoid splitting sendTransaction implementation between web and native so definitely don't want local storage to be baked into it

packages/thirdweb/src/transaction/transaction-store.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+const transactionsByAddress = new Map<Address, Store<StoredTransaction[]>>();
+
+/**
+ * Retrieve transactions for the given account for the current session. Sessions are stored in memory and are not persisted.
+ * @param account {Account} - The account to retrieve transactions for.
+ * @returns {StoredTransaction[]} An array of stored pending and confirmed transactions.
+ */
+export async function getSessionTransactions(
```
jnsdls, line 17: this doesn't have to be async I believe?

packages/thirdweb/src/transaction/transaction-store.ts
```diff
@@ -0,0 +1,57 @@
...
+  address: Address,
+): Promise<StoredTransaction[]> {
+  const tranasctionStore = transactionsByAddress.get(address);
```
jnsdls, line 20: probably instead simply use `getTransactionStore()`, there is no harm creating an empty store in this case

jnsdls: .. and that way you have 1 clear path to how to interact with the map

packages/thirdweb/src/exports/transaction.ts
```diff
@@ -55,6 +55,7 @@ export {
   toSerializableTransaction,
   type ToSerializableTransactionOptions,
 } from "../transaction/actions/to-serializable-transaction.js";
+export { getSessionTransactions } from "../transaction/transaction-store.js";
```
jnsdls, line 58: maybe it makes sense to always let people access the store instead here, that way they can still `getValue()` to get the data, but also they can subscribe to updates etc which seems useful?

jnsdls: and you could remove the implementation of `getSessionTransactions` entirely (since its essentially a proxy of `getTransactionStore().getValues()` anyways?

packages/thirdweb/src/transaction/transaction-store.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export function getTransactionStore(
+  address: string,
+): Store<StoredTransaction[]> {
+  const existingStore = transactionsByAddress.get(address);
+  if (existingStore) {
+    return existingStore;
+  }
+
+  return createStore<StoredTransaction[]>([]);
```
jnsdls, line 24: you're never actually then adding this to the map

jnsdls (changes requested):

jnsdls (approved):

PR #3676, opened 2024-07-12
Allow to add extra call data to prepareContractCall

kien-ngo:
## PR-Codex overview
This PR enhances the `prepareContractCall` function in the `thirdweb` package to support adding extra call data to transactions.

### Detailed summary
- Added support for `extraCallData` in `prepareContractCall`
- Implemented `getDataFromTx` to extract transaction data
- Implemented `getExtraCallDataFromTx` to extract extra call data

### DIFF
--- a/packages/thirdweb/src/transaction/prepare-contract-call.test.ts
+++ b/packages/thirdweb/src/transaction/prepare-contract-call.test.ts
@@ -1,13 +1,11 @@
-import { describe, expect, test } from "vitest";
-
+import { describe, expect, it } from "vitest";
 import { TEST_WALLET_B } from "../../test/src/addresses.js";
-
 import { USDT_CONTRACT_WITH_ABI } from "../../test/src/test-contracts.js";
 import { toWei } from "../utils/units.js";
 import { prepareContractCall } from "./prepare-contract-call.js";
 
-describe("prepareContractCall", () => {
-  test("should prepare a contract call with ABI", () => {
+describe.runIf(process.env.TW_SECRET_KEY)("prepareContractCall", () => {
+  it("should prepare a contract call with ABI", () => {
     const preparedTx = prepareContractCall({
       contract: USDT_CONTRACT_WITH_ABI,
       method: "transfer",
--- a/packages/thirdweb/src/transaction/prepare-contract-call.ts
+++ b/packages/thirdweb/src/transaction/prepare-contract-call.ts
@@ -121,6 +121,23 @@ export type PrepareContractCallOptions<
  *  params: [to, value],
  * });
  * ```
+ *
+ * @example
+ * Passing extra call data to the transaction
+ * ```ts
+ * import { getContract, prepareContractCall } from "thirdweb";
+ * const contract = getContract({
+ *   ..., // chain, address, client
+ * });
+ *
+ * const transaction = prepareContractCall({
+ *   contract,
+ *   method: "function transfer(address to, uint256 value)",
+ *   params: [...],
+ *   // The extra call data MUST be encoded to hex before passing
+ *   extraCallData: "0x......."
+ * });
+ * ```
  */
 export function prepareContractCall<
   const TAbi extends Abi,
--- a/.changeset/strange-yaks-move.md
+++ b/.changeset/strange-yaks-move.md
@@ -1,0 +1,5 @@
+---
+"thirdweb": patch
+---
+
+Allow to add extra call data to `prepareContractCall`
--- a/packages/thirdweb/src/transaction/actions/encode.test.ts
+++ b/packages/thirdweb/src/transaction/actions/encode.test.ts
@@ -1,13 +1,25 @@
 import { describe, expect, it } from "vitest";
-import { TEST_WALLET_A } from "~test/addresses.js";
+import { TEST_WALLET_A, TEST_WALLET_B } from "~test/addresses.js";
+import { FORKED_ETHEREUM_CHAIN } from "~test/chains.js";
+import { TEST_CLIENT } from "~test/test-clients.js";
 import { USDT_CONTRACT, USDT_CONTRACT_WITH_ABI } from "~test/test-contracts.js";
-import { encode } from "./encode.js";
-
+import { concatHex } from "../../utils/encoding/helpers/concat-hex.js";
+import { toHex } from "../../utils/encoding/hex.js";
+import { toWei } from "../../utils/units.js";
 import { prepareContractCall } from "../prepare-contract-call.js";
+import { prepareTransaction } from "../prepare-transaction.js";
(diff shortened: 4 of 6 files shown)


iuwqyir (approved):
Also would be good to add a test for a function call with parameters

kien-ngo:
actually the test is for a function call with parameters. We need one without a parameter

kien-ngo:
Added!

packages/thirdweb/src/transaction/prepare-contract-call.ts
```diff
@@ -52,6 +53,7 @@ export type PrepareContractCallOptions<
   > & {
     contract: ThirdwebContract<TAbi>;
     method: TMethod | TPreparedMethod;
+    extraCallData?: ToHexInput;
```
jnsdls, line 56: make people pass Hex here instead of doing it under the hood

packages/thirdweb/src/transaction/prepare-contract-call.test.ts
```diff
@@ -15,4 +14,36 @@ describe("prepareContractCall", () => {
     });
     expect(preparedTx.to).toBe(USDT_CONTRACT_WITH_ABI.address);
   });
+
+  test("should be able to prepare a contract call with parameters & extra call data", async () => {
```
jnsdls, line 18: add a test that actually also executes the TX -> make sure that didn't break

packages/thirdweb/src/utils/encoding/hex.ts
```diff
@@ -474,6 +474,8 @@ export type ToHexParameters = {
   size?: number;
 };
 
+export type ToHexInput = string | number | bigint | boolean | Uint8Array;
```
jnsdls, line 477: won't be needed if you always make people pass hex

packages/thirdweb/src/transaction/prepare-contract-call.ts
```diff
@@ -193,11 +195,39 @@ export function prepareContractCall<
           preparedM = await preparedMethodPromise();
         }
 
+        let extraData: Hex | null = null;
+        if (extraCallData) {
+          const { toHex } = await import("../utils/encoding/hex.js");
+          extraData = toHex(extraCallData);
+        }
```
jnsdls, line 202: instead make people pass hex

packages/thirdweb/src/transaction/prepare-contract-call.ts
```diff
@@ -193,11 +195,39 @@ export function prepareContractCall<
...
+
+        // just return the fn sig directly -> no params
         if (preparedM[1].length === 0) {
-          // just return the fn sig directly -> no params
+          if (extraData) {
+            const { concatHex } = await import(
```
jnsdls, line 207: just always import it and also use it for the other concatenation we do already

jnsdls (changes requested):

packages/thirdweb/src/transaction/prepare-contract-call.ts
```diff
@@ -193,11 +213,27 @@ export function prepareContractCall<
           preparedM = await preparedMethodPromise();
         }
 
+        // just return the fn sig directly -> no params
         if (preparedM[1].length === 0) {
```
joaquim-verges, line 217: I wouldn't append the extra data here. Instead i would append it inside `encode(transaction)`

packages/thirdweb/src/transaction/prepare-contract-call.ts
```diff
@@ -52,6 +54,7 @@ export type PrepareContractCallOptions<
   > & {
     contract: ThirdwebContract<TAbi>;
     method: TMethod | TPreparedMethod;
+    extraCallData?: Hex;
```
joaquim-verges, line 57: can you add this in `StaticPrepareTransactionOptions` instead? that way its easy to make it work with extension functions too (which you need to handle as well)

joaquim-verges (changes requested):
couple of changes needed

packages/thirdweb/src/transaction/prepare-contract-call.ts
kien-ngo: image

@joaquim-verges joa
do you mean we should do the concatHex for the extra calldata in here?

packages/thirdweb/src/transaction/prepare-contract-call.ts
kien-ngo: Done

packages/thirdweb/src/transaction/prepare-contract-call.test.ts
```diff
@@ -1,18 +1,64 @@
...
+      method: "approve",
+      params: [TEST_WALLET_B, 100n],
+      extraCallData,
+    });
+    const callData =
+      typeof transaction.data === "string"
+        ? transaction.data
+        : (await transaction.data?.()) || "0x";
+    const decodedData = fromHex(callData, "string");
+    expect(decodedData).toContain(extraString);
```
joaquim-verges, line 36: this should just be a call to `encode(transaction)` no?

packages/thirdweb/src/transaction/prepare-contract-call.test.ts
```diff
@@ -1,18 +1,64 @@
...
+      contract: USDT_CONTRACT_WITH_ABI,
+      method: "decimals",
+      params: [],
+      extraCallData,
+    });
+    const callData =
+      typeof transaction.data === "string"
+        ? transaction.data
+        : (await transaction.data?.()) || "0x";
+    const decodedData = fromHex(callData, "string");
```
joaquim-verges, line 56: same here

packages/thirdweb/src/transaction/prepare-contract-call.test.ts
kien-ngo: oh this is old test case for back then when i added the extraCallData to prepareContractCall only. I will remove these changes and prepare new set of tests

packages/thirdweb/src/transaction/prepare-contract-call.ts
```diff
@@ -121,6 +121,23 @@ export type PrepareContractCallOptions<
...
+ * Passing extra call data to the transaction
+ * ```ts
+ * import { getContract, prepareContractCall } from "thirdweb";
+ * const contract = getContract({
+ *   ..., // chain, address, client
+ * });
+ *
+ * const transaction = prepareContractCall({
+ *   contract,
+ *   method: "transfer",
```
joaquim-verges, line 135:
```suggestion
 *   method: "function transfer(address to, uint256 value)",
```

joaquim-verges (approved):

jnsdls (approved):

PR #3712, opened 2024-07-16
updates: UserOp example + receipt retrieval function

farhanW3:
## PR-Codex overview
This PR updates functions in `smart.ts`, `userop.ts`, `paymaster.ts`, and `bundler.ts` to include a `decodeRevertReason` flag and adds `getPaymasterAndData` function for Paymaster details.

### Detailed summary
- Updated `getUserOpReceipt` and `waitForUserOpReceipt` functions
- Added `getUserOpReceiptRaw` function
- Exposed `getPaymasterAndData` function for Paymaster details
- Modified parameters in `createUnsignedUserOp` and `signUserOp`
- Added `formatUserOperationReceipt` function for user operation receipt formatting


packages/thirdweb/src/wallets/smart/lib/bundler.ts
```diff
@@ -144,7 +145,7 @@ export async function getUserOpReceipt(
   if (!res) {
     return undefined;
   }
-  if (res.success === false) {
+  if (res.success === false && args.decodeRevertReason) {
```
jnsdls, line 148: this changes default behavior, not sure if we can do this

instead what could work is adding a `skipDecodeRevertReason` (inverse logic) so the default behavior stays the same but you can explicitly skip this step

farhanW3: makes sense. will do

farhanW3: @jnsdls made the updates

packages/thirdweb/src/wallets/smart/lib/userop.ts
```diff
@@ -54,6 +54,7 @@ export async function waitForUserOpReceipt(
     userOpHash: Hex;
     timeoutMs?: number;
     intervalMs?: number;
+    skipDecodeRevertReason?: boolean;
```
gregfromstl, line 57: Is this used at all here?

gregfromstl (changes requested):

packages/thirdweb/src/wallets/smart/lib/userop.ts
farhanW3: actually, now since we are not using createUnsignedOp() & signUserOp() we might not need this.

joaquim-verges (approved):

gregfromstl (approved):

PR #3730, opened 2024-07-17
Increased Transaction cost buffer for Pay Modal

MananTank:
## PR-Codex overview
The focus of this PR is to update the buffer calculation of transactions when gas estimation fails for the Pay Modal.

### Detailed summary
- Update buffer calculation to use the transaction value + 2M * gas price
- Introduce gasPrice calculation using getGasPrice function
- Adjust the fallback logic to include the new buffer calculation

### DIFF
--- a/.changeset/blue-steaks-join.md
+++ b/.changeset/blue-steaks-join.md
@@ -1,0 +1,5 @@
+---
+"thirdweb": patch
+---
+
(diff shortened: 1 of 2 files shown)


joaquim-verges (approved):
@MananTank lets do the change we talked about before merging

joaquim-verges (approved):

PR #3781, opened 2024-07-22
Dashboard color fixes

MananTank:
## PR-Codex overview
This PR updates text colors and styles across components for better consistency and readability. 

### Detailed summary
- Updated text color classes to use `text-success-text` and `text-destructive-text` for better clarity
- Modified background colors to match the new text color changes
- Improved styling for deprecated and audited components

### DIFF
--- a/apps/dashboard/src/@/components/ui/DatePickerWithRange.tsx
+++ b/apps/dashboard/src/@/components/ui/DatePickerWithRange.tsx
@@ -69,7 +69,7 @@ export function DatePickerWithRange(props: {
         <DynamicHeight>
           <div>
             {!isValid && (
-              <p className="text-sm pt-2 justify-center text-destructive-foreground text-center flex items-center gap-2">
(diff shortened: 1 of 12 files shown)


jnsdls (commented):
are these based on the design system?

Also lets please keep the css vars we have to the *absolute* minimum we can - every new css var we "invent" at this point on top of what shadcn comes with adds maintenance burden

MananTank:
> are these based on the design system?

no, I'm just fixing up the issues I noticed - missing colors, text and background having the same color, etc 

We need to add 2 new colors for the showing green and red texts - shadcn doesn't have CSS variables for that - it only has a CSS variable for "destructive" action - and that is for creating a destructive background and a text that looks good on the destructive background 

shadcn doesn't have a "success" color - we added it, but it was not following the shadcn pattern where `"success"` is background color and `"success-foreground"` is for text that looks good on success background. 

That's why we need to add two new colors for showing green-ish and red-ish texts that looks good on any neutral bg

joaquim-verges (approved):

PR #3847, opened 2024-07-27
Prebuilt UI: BuyDirectListingButton for Marketplace v3

kien-ngo:
## PR-Codex overview
This PR adds a new prebuilt component `BuyDirectListingButton` for Marketplace v3 in Thirdweb.

### Detailed summary
- Added `BuyDirectListingButton` component for buying NFTs from a listing
- Includes props for contract address, chain, client, listing ID, and quantity
- Utilizes `TransactionButton` for transaction handling

### DIFF
--- a/.changeset/honest-gorillas-count.md
+++ b/.changeset/honest-gorillas-count.md
@@ -1,0 +1,18 @@
+---
+"thirdweb": minor
+---
+
+Add prebuilt component: BuyDirectListingButton for Marketplace v3
+```tsx
+import { BuyDirectListingButton } from "thirdweb/react";
+
+<BuyDirectListingButton
+  contractAddress="0x..." // contract address of the marketplace v3
+  chain={...} // the chain which the marketplace contract is deployed on
+  client={...} // thirdweb client
+  listingId={100n} // the listingId or the item you want to buy
+  quantity={1n} // optional - see the docs to learn more
+>
+  Buy NFT
+</BuyDirectListingButton>
+```
--- a/packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyDirectListingButton/index.tsx
+++ b/packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyDirectListingButton/index.tsx
@@ -1,0 +1,200 @@
+"use client";
+
+import { useCallback } from "react";
+import type { Chain } from "../../../../../../chains/types.js";
+import type { ThirdwebClient } from "../../../../../../client/client.js";
+import { getContract } from "../../../../../../contract/contract.js";
+import { getListing } from "../../../../../../extensions/marketplace/direct-listings/read/getListing.js";
+import type { BaseTransactionOptions } from "../../../../../../transaction/types.js";
+import { useReadContract } from "../../../../../core/hooks/contract/useReadContract.js";
+import type { TransactionButtonProps } from "../../../../../core/hooks/transaction/transaction-button-utils.js";
+import { useSendAndConfirmTransaction } from "../../../../../core/hooks/transaction/useSendAndConfirmTransaction.js";
+import { useActiveAccount } from "../../../../../core/hooks/wallets/useActiveAccount.js";
+import { TransactionButton } from "../../../TransactionButton/index.js";
+
+export type BuyDirectListingButtonProps = Omit<
+  TransactionButtonProps,
+  "transaction"
+> & {
+  /**
+   * The contract address of the Marketplace v3 contract.
+   */
+  contractAddress: string;
+
+  /**
+   * The chain which the Drop contract is deployed on
+   */
+  chain: Chain;
+
+  /**
+   * thirdweb Client
+   */
+  client: ThirdwebClient;
+
+  /**
+   * ID of the marketplace's DirectListing
+   */
+  listingId: bigint;
+
+  /**
+   * Qty to buy (optional)
+   *
+   * - For ERC721 listing: the `quantity` is always hard-coded to 1n - passing this props doesn't do anything
+   *
+   * - For ERC1155 listing: the `quantity` defaults to the quantity of the listing if not specified.
+   *
+   * The component will also throw an error if  you pass a `quantity` and it's greater than the listing's quantity
+   */
+  quantity?: bigint;
+};
+
+/**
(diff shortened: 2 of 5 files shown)


.changeset/honest-gorillas-count.md
```diff
@@ -0,0 +1,5 @@
+---
+"thirdweb": minor
+---
+
+Add prebuilt component: BuyButton for Marketplace v3
```
jnsdls, line 5: add a usage example? :)

packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyButton/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
+   * thirdweb Client
+   */
+  client: ThirdwebClient;
+
+  /**
+   * ID of the marketplace's DirectListing
+   */
+  listingId: bigint;
```
jnsdls, line 33: if this is for direct listings only then maybe we can name this `BuyDirectListingButton` or something? We'll likely also want a component later for auctions etc and I wonder if it would get confusing otherwise? (my name suggestion might also not be good but more explicit naming is probably beneficial here?)

kien-ngo: No I agree as that thought has been in my head. I don't want any vague name but I was also scared of naming things too "verbosely".

Now that you have mentioned, I'm gonna rename it.

packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyDirectListingButton/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+        import(
+          "../../../../../../extensions/marketplace/direct-listings/write/buyFromListing.js"
+        ),
+      ]);
+    if (!listing) {
+      throw new Error(`Could not retrieve listing with ID: ${listingId}`);
+    }
+
+    // For ERC721 listing, always hard-code to 1n
+    const _quantity =
```
gregfromstl, line 128: Instead of implicitly overriding the inputted quantity here we should throw an error if they try specifying a quantity > 1 for ERC721. Or adjust the types to not allow it. Or both, ideally both to cover both JS and TS

kien-ngo: wow that's good thinking ser. I will add "cover both js and ts" to my list!

kien-ngo: although I don't think we can adjust the type? Because we won't know if it's ERC721 until we fetch the listing. So maybe just throw error?

packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyDirectListingButton/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+      quantity: _quantity,
+      recipient: account?.address || "",
+    });
+
+    const approveTx = await getApprovalForTransaction({
+      transaction: buyTx,
+      account,
+    });
+
+    if (approveTx) {
```
gregfromstl, line 154: Kind of weird we execute a transaction inside the prepare but I guess that's the limitation of TransactionButton

kien-ngo: Yeah.. I talked to Joaquim about making a change to the TB, so that we can "chain" multiple transactions.

But for now we think this should do

packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyDirectListingButton/index.tsx
kien-ngo: @gregfromstl actually do you think it would be better UX + DX to handle the ERC721 case for user? Because in the code we only know if it's ERC721 once we have fetched the listing, and if users intend to buy, it's always gonna be at least "1" anyway

packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyDirectListingButton/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+ *   contractAddress="0x..." // contract address of the marketplace v3
+ *   chain={...} // the chain which the marketplace contract is deployed on
+ *   client={...} // thirdweb client
+ *   listingId={100n} // the listingId or the item you want to buy
+ *   quantity={1n} // optional - see the docs to learn more
+ * >
+ *   Buy NFT
+ * </BuyDirectListingButton>
+ * ```
+ */
```
joaquim-verges, line 77: remember to tag with @component

packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyDirectListingButton/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+    contract,
+    listingId,
+    queryOptions: {
+      enabled: !defaultPayModalMetadata,
+    },
+  });
+
+  const { mutateAsync } = useSendAndConfirmTransaction();
+
+  async function prepareBuyTransaction() {
```
joaquim-verges, line 106: could use a `useCallback` for this so it doesnt recreate on every render, but only when props change

packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyDirectListingButton/index.tsx
kien-ngo: oops, got it

packages/thirdweb/src/react/web/ui/prebuilt/thirdweb/BuyDirectListingButton/index.tsx
kien-ngo: Done!

joaquim-verges (approved):

PR #3910, opened 2024-08-02
<Crusade> Claim NFT drop to v5

kien-ngo:
## Problem solved

---

## PR-Codex overview
The focus of this PR is to refactor NFT claim functionality by updating contract handling and adding transaction confirmation.

### Detailed summary
- Refactored `NFTClaimButton` and `NFTClaimForm` to use `contractAddress` and `chainId`
- Added transaction confirmation for NFT claiming
- Updated contract handling in `NFTClaimForm`

### DIFF
--- a/apps/dashboard/src/contract-ui/tabs/nfts/components/claim-button.tsx
+++ b/apps/dashboard/src/contract-ui/tabs/nfts/components/claim-button.tsx
@@ -1,35 +1,19 @@
 import { Icon, useDisclosure } from "@chakra-ui/react";
 import { GiDiamondHard } from "@react-icons/all-files/gi/GiDiamondHard";
-import type { useContract } from "@thirdweb-dev/react";
-import { extensionDetectedState } from "components/buttons/ExtensionDetectButton";
 import { Button, Drawer } from "tw-components";
 import { NFTClaimForm } from "./claim-form";
 
 interface NFTClaimButtonProps {
-  contractQuery: ReturnType<typeof useContract>;
+  contractAddress: string;
+  chainId: number;
 }
 
 export const NFTClaimButton: React.FC<NFTClaimButtonProps> = ({
-  contractQuery,
+  contractAddress,
+  chainId,
   ...restButtonProps
 }) => {
   const { isOpen, onOpen, onClose } = useDisclosure();
-
-  const detectedState = extensionDetectedState({
-    contractQuery,
-    feature: [
-      // erc 721
-      "ERC721ClaimPhasesV1",
-      "ERC721ClaimPhasesV2",
-      "ERC721ClaimConditionsV1",
-      "ERC721ClaimConditionsV2",
-      "ERC721ClaimCustom",
-    ],
-  });
-
-  if (detectedState !== "enabled" || !contractQuery.contract) {
-    return null;
-  }
 
   return (
     <>
@@ -40,7 +24,7 @@ export const NFTClaimButton: React.FC<NFTClaimButtonProps> = ({
         onClose={onClose}
         isOpen={isOpen}
       >
-        <NFTClaimForm contract={contractQuery.contract} />
+        <NFTClaimForm contractAddress={contractAddress} chainId={chainId} />
       </Drawer>
       <Button
         colorScheme="primary"
--- a/apps/dashboard/src/contract-ui/tabs/nfts/components/claim-form.tsx
+++ b/apps/dashboard/src/contract-ui/tabs/nfts/components/claim-form.tsx
@@ -7,13 +7,15 @@ import {
   Stack,
   useModalContext,
 } from "@chakra-ui/react";
-import { type DropContract, useClaimNFT } from "@thirdweb-dev/react";
 import { TransactionButton } from "components/buttons/TransactionButton";
 import { useTrack } from "hooks/analytics/useTrack";
 import { useTxNotifications } from "hooks/useTxNotifications";
+import { thirdwebClient } from "lib/thirdweb-client";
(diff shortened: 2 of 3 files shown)


apps/dashboard/src/contract-ui/tabs/nfts/components/claim-form.tsx
```diff
@@ -79,29 +88,32 @@ export const NFTClaimForm: React.FC<NFTClaimFormProps> = ({ contract }) => {
...
-                    label: "error",
-                    error,
-                  });
-                  onError(error);
-                },
+
+            const transaction = claimTo({
+              contract,
+              to: d.to ?? "",
+              quantity: BigInt(d.amount),
```
MananTank, line 95: V4 was able to take no `to` field. not sure what's changed in v5

MananTank: @jnsdls can probably verify this one

kien-ngo: it's an extension so ig it is what it is (we have to specify the field)

jnsdls: `claimTo` _has_ to have an address, the behavior in v4 for claim without address would claim to the connected address so we have to do the same here, empty string will throw an error during the mutation here I believe

jnsdls: unless we handle this in the extension (we shouldn't!) the consumer is responsible to pass the address here (meaning dashboard has to pass it) `d.to` is no longer optional here

apps/dashboard/src/contract-ui/tabs/nfts/components/claim-form.tsx
```diff
@@ -23,12 +24,20 @@ import {
...
 }
 
-export const NFTClaimForm: React.FC<NFTClaimFormProps> = ({ contract }) => {
+export const NFTClaimForm: React.FC<NFTClaimFormProps> = ({
+  contractAddress,
+  chainId,
+}) => {
+  const contract = getContract({
+    address: contractAddress,
+    chain: defineChain(chainId),
```
jnsdls, line 37: v4tov5definechain util thing here instead?

apps/dashboard/src/contract-ui/tabs/nfts/components/claim-button.tsx
```diff
@@ -40,7 +24,7 @@ export const NFTClaimButton: React.FC<NFTClaimButtonProps> = ({
         onClose={onClose}
         isOpen={isOpen}
       >
-        <NFTClaimForm contract={contractQuery.contract} />
+        <NFTClaimForm contractAddress={contractAddress} chainId={chainId} />
```
jnsdls, line 27: I wonder if this implicitly means we are removing support for some older claim types? does v5 claimTo extension handle all the v4 supported cases? I legitimately do not know cc @joaquim-verges @gregfromstl?

jnsdls: if it does mean that then same as with marketplace that's probably OK we just have to make sure support is informed of this - so we're not surprised of _possible_ issues

apps/dashboard/src/contract-ui/tabs/nfts/page.tsx
```diff
@@ -48,6 +49,18 @@ export const ContractNFTPage: React.FC<NftOverviewPageProps> = ({
     "ERC721ClaimCustom",
   ]);
 
+  const detectedClaimState = extensionDetectedState({
```
jnsdls, line 52: nice, I like moving this up - we also have to figure out how to do this same thing in v5 soon

jnsdls (changes requested):

apps/dashboard/src/contract-ui/tabs/nfts/components/claim-form.tsx
kien-ngo: correct. just read the code for getClaimParams. I believe if we dont pass it we will get an "execution reverted"

apps/dashboard/src/contract-ui/tabs/nfts/components/claim-form.tsx
kien-ngo: yesser

apps/dashboard/src/contract-ui/tabs/nfts/components/claim-form.tsx
jnsdls: yep, so form should be invalid here if no address is filled out meaning we should never need an  empty string for this case

kien-ngo: let me check. i think typescript was screaming at me (irrc it didn't understand the required* field from react-form-hook which should prevent the falsy case)

jnsdls: if that's the case just `if(!d.to) {don't try to do the tx just show the error immediately here}`

jnsdls (approved):
please make sure support is informed that we changed this and to report any uptick in issues with claiming they may see

apps/dashboard/src/contract-ui/tabs/nfts/components/claim-button.tsx
kien-ngo: Will imform the guys about this change!

PR #4032, opened 2024-08-09
feat(react): add `enabled` param to `useWalletBalance` hook

alecananian:
Adds an `enabled` param to the `queryOptions` in the `useWalletBalance` hook so consumers of this hook can control the overall enabling/disabling of the fetch request without having to do a hacky workaround like passing `address: undefined`.

An example use case for this is fetching the balance of a token that may exist on some chains but not others:
```
const result = useWalletBalance({
  client,
  chain,
  address,
  tokenAddress,
}, {
  enabled: !!tokenAddress,
});
```

---

## PR-Codex overview
The focus of this PR is to add an `enabled` parameter to the `useWalletBalance` hook in `useWalletBalance.ts`.

### Detailed summary
- Added `enabled` parameter to `useWalletBalance` hook
- Updated the `UseWalletBalanceQueryOptions` type to include the `enabled` parameter
- Modified the `useWalletBalance` function to utilize the new `enabled` parameter for conditional activation

### DIFF
--- a/.changeset/yellow-numbers-yell.md
+++ b/.changeset/yellow-numbers-yell.md
@@ -1,0 +1,5 @@
+---
+"thirdweb": patch
+---
+
+Added `enabled` param to `useWalletBalance` hook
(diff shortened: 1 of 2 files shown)


jnsdls:
nice, thanks @alecananian this makes sense - will get this in and cut a release :)

jnsdls (approved):

PR #4100, opened 2024-08-14
[Dashboard] ContractMetadata settings to v5

kien-ngo:
## PR-Codex overview
This PR updates the contract metadata settings page by improving the schema, handling social URLs, and refactoring contract-related functions.

### Detailed summary
- Updated `CommonContractSchema` with new fields
- Refactored `SettingsMetadata` component to handle social URLs
- Updated contract-related functions and imports

### DIFF
--- a/apps/dashboard/src/constants/schemas.ts
+++ b/apps/dashboard/src/constants/schemas.ts
@@ -49,3 +49,19 @@ const AddressSchema = z.custom<string>(
 export const AddressOrEnsSchema = z.union([AddressSchema, EnsSchema], {
   invalid_type_error: "Provided value was not a valid address or ENS name",
 });
+
+const FileSchema = z.instanceof(File) as z.ZodType<InstanceType<typeof File>>;
+
+const FileOrStringSchema = z.union([FileSchema, z.string()]);
+
+export const CommonContractSchema = z
+  .object({
+    name: z.string(),
+    description: z.string().optional(),
+    image: FileOrStringSchema.optional(),
+    external_link: z.string().optional(),
+    app_uri: z.string().optional(),
+    social_urls: z.record(z.string()).optional(),
+    defaultAdmin: AddressOrEnsSchema.optional(),
+  })
+  .catchall(z.unknown());
--- a/apps/dashboard/src/contract-ui/tabs/settings/components/metadata.tsx
+++ b/apps/dashboard/src/contract-ui/tabs/settings/components/metadata.tsx
@@ -9,26 +9,23 @@ import {
   Textarea,
 } from "@chakra-ui/react";
 import { zodResolver } from "@hookform/resolvers/zod";
-import { useContractMetadata } from "@thirdweb-dev/react";
-import {
-  CommonContractSchema,
-  type ValidContractInstance,
-} from "@thirdweb-dev/sdk/evm";
 import type { ExtensionDetectedState } from "components/buttons/ExtensionDetectButton";
 import { TransactionButton } from "components/buttons/TransactionButton";
 import { FileInput } from "components/shared/FileInput";
+import { CommonContractSchema } from "constants/schemas";
 import { useTrack } from "hooks/analytics/useTrack";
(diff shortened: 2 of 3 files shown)


apps/dashboard/src/constants/schemas.ts
```diff
@@ -49,3 +49,31 @@ const AddressSchema = z.custom<string>(
...
+    : // @fixme, this is a hack to make browser happy for now
+      (z.instanceof(Buffer) as z.ZodTypeAny);
+
+const FileOrBufferSchema = z.union([
+  FileOrBufferUnionSchema,
+  z.object({
+    data: z.union([FileOrBufferUnionSchema, z.string()]),
+    name: z.string(),
+  }),
+]);
```
jnsdls, line 65: I do not think you need this in dashboard, all you need is File or string, there are no `Buffer` in the frontend here? Can we simplify it based on that?

apps/dashboard/src/contract-ui/tabs/settings/components/metadata.tsx
```diff
@@ -63,27 +60,26 @@ function extractDomain(url: string) {
...
 
   const transformedQueryData = useMemo(() => {
     return {
       ...metadata.data,
       name: metadata.data?.name || "",
+      image: metadata.data?.image || "",
       dashboard_social_urls: Object.entries(
         metadata.data?.social_urls || { twitter: "", discord: "" },
-      ).map(([key, value]) => ({ key, value })),
+      ).map(([key, value]) => ({ key, value: String(value) })),
```
jnsdls, line 82: why is this required, what else other than string would it be?

apps/dashboard/src/contract-ui/tabs/settings/components/metadata.tsx
```diff
@@ -111,20 +107,10 @@ export const SettingsMetadata = <
...
-          client: thirdwebClient,
-        })
-      : null;
-
   const { onSuccess, onError } = useTxNotifications(
     "Successfully updated metadata",
     "Error updating metadata",
-    contractV5,
+    contract,
   );
```
jnsdls, line 114: not a blocker but I would like to request an update to this hook (and it also no longer has to be a hook in this world):

we now do a lot of `await mutation.mutateAsync()` -> can we please make a new version of this that takes a `mutation.mutateAsync` promise result (from the tx hook) -> passes it to sonner for toasts etc?

jnsdls (changes requested):

apps/dashboard/src/contract-ui/tabs/settings/components/metadata.tsx
kien-ngo: This was to make typescript happy as the type here is unknown. I can try to find a better approach

jnsdls: what's the type of `metadata.data.social_urls` ?

kien-ngo: `any`. Because here's the type of `metadata.data` from v5:

```ts
(property) data: {
    [key: string]: any;
    name: string;
    symbol: string;
} | undefined
```

kien-ngo: Do we update the type in SDK?

jnsdls: no, you'll probably have to add a typeguard here, so a function that takes in `metadata` and validates that the keys have certain shapes etc (you could use zod for this) and then returns it back out

apps/dashboard/src/contract-ui/tabs/settings/components/metadata.tsx
kien-ngo: I made the linear ticket (CNCT-1163)
will work on it

jnsdls (approved):

PR #4109, opened 2024-08-15
[Dashboard] Migrate type: ProfileMetadata + ProfileMetadataInput

kien-ngo:
## Problem solved

---

## PR-Codex overview
This PR updates import paths and types for `ProfileMetadata` in multiple components and schemas.

### Detailed summary
- Updated import paths for `ProfileMetadata` in various components
- Changed type definitions to use `ProfileMetadataInput` and `ProfileMetadata` from `constants/schemas`
- Added `ProfileSchemaInput` and `ProfileSchemaOutput` in `schemas.ts`

### DIFF
--- a/apps/dashboard/src/components/contract-components/hooks.ts
+++ b/apps/dashboard/src/components/contract-components/hooks.ts
@@ -18,7 +18,6 @@ import {
   type ExtraPublishMetadata,
   type FeatureName,
   type FeatureWithEnabled,
-  type ProfileMetadata,
   type PublishedContract,
   type ThirdwebSDK,
   detectFeatures,
@@ -38,6 +37,7 @@ import {
   zkDeployContractFromUri,
 } from "@thirdweb-dev/sdk/evm/zksync";
 import type { SnippetApiResponse } from "components/contract-tabs/code/types";
+import type { ProfileMetadata, ProfileMetadataInput } from "constants/schemas";
 import type { providers } from "ethers";
 import { useSupportedChain } from "hooks/chains/configureChains";
 import { isEnsName, resolveEns } from "lib/ens";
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/components/contract-components/hooks.ts
```diff
@@ -300,7 +300,10 @@ async function fetchPublisherProfile(publisherAddress?: string | null) {
     getDashboardChainRpc(polygon.id, undefined),
   );
   invariant(publisherAddress, "address is not defined");
-  return await sdk.getPublisher().getPublisherProfile(publisherAddress);
+  return (await sdk
+    .getPublisher()
+    // todo: remove type-casting once we have replaced this method
+    .getPublisherProfile(publisherAddress)) as ProfileMetadata;
```
jnsdls, line 306: would love to do that at the same time as this in the PR? what would that take? if it's too much we can ship this first but the real win is removing this

kien-ngo: I will do it in this PR. there "no rush" getting this one out anyway

kien-ngo: hmm I think we still have to type-cast one way or another? Because we are getting the profile metadata from an IPFS uri  (by calling the method `getPublisherProfileUri` on our Publisher contract)

Maybe I should do a separate PR where I replace all the v4 publisher code inside `src/components/contract-components/hooks.ts`

wdyt @jnsdls

jnsdls: @kien-ngo if you want to do that, please go ahead - otherwise I'll do it today and we can merge this first

jnsdls (approved):

PR #4145, opened 2024-08-16
[Dashboard] Migrate useContractSources to v5

kien-ngo:
## PR-Codex overview
The focus of this PR is to update the `compiler-metadata` type and enhance the `ContractSourcesPage` with new functionality.

### Detailed summary
- Updated `metadata` type in `compiler-metadata.ts`
- Added `thirdwebClient` import and `getContract` usage in `page.tsx`
- Refactored `useContractSources` function in `useContractSources.ts` to handle IPFS URIs

### DIFF
--- a/apps/dashboard/src/contract-ui/hooks/useContractSources.ts
+++ b/apps/dashboard/src/contract-ui/hooks/useContractSources.ts
@@ -30,7 +30,9 @@ export function useContractSources(contract?: ThirdwebContract) {
             const source = await download({
               uri: `ipfs://${ipfsHash}`,
               client: thirdwebClient,
-            }).then((r) => r.text());
+            })
+              .then((r) => r.text())
+              .catch(() => "Failed to fetch source from IPFS");
             return {
               filename: path,
               source,
--- a/apps/dashboard/src/contract-ui/hooks/useContractSources.ts
+++ b/apps/dashboard/src/contract-ui/hooks/useContractSources.ts
@@ -6,7 +6,7 @@ import invariant from "tiny-invariant";
 export function useContractSources(contract?: ThirdwebContract) {
   return useQuery(
     [
-      "dashboard-contract-sources",
+      "contract-sources",
       contract?.chain.id || "",
       contract?.address || "",
     ],
--- a/apps/dashboard/src/contract-ui/hooks/useContractSources.ts
+++ b/apps/dashboard/src/contract-ui/hooks/useContractSources.ts
@@ -5,6 +5,8 @@ import { getCompilerMetadata } from "thirdweb/contract";
 import { download } from "thirdweb/storage";
 import invariant from "tiny-invariant";
 
+// An example for a contract that has IPFS URIs in its metadata: abstract-testnet/0x8A24a7Df38fA5fCCcFD1259e90Fb6996fDdfcADa
+
 export function useContractSources(contract?: ThirdwebContract) {
   return useQuery({
     queryKey: [
--- a/apps/dashboard/src/contract-ui/hooks/useContractSources.ts
+++ b/apps/dashboard/src/contract-ui/hooks/useContractSources.ts
@@ -1,6 +1,8 @@
+import { thirdwebClient } from "@/constants/client";
 import { useQuery } from "@tanstack/react-query";
 import type { ThirdwebContract } from "thirdweb";
 import { getCompilerMetadata } from "thirdweb/contract";
+import { download } from "thirdweb/storage";
 import invariant from "tiny-invariant";
 
 export function useContractSources(contract?: ThirdwebContract) {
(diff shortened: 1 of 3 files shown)


apps/dashboard/src/contract-ui/hooks/useContractSources.ts
```diff
@@ -1,17 +1,21 @@
...
+      invariant(contract, "contract is required");
+      const data = await getCompilerMetadata(contract);
+      const sources = data.metadata.sources || {};
+      const arr = Object.keys(sources).map((key) => ({
+        filename: key,
+        source: sources[key]?.content || "",
+      }));
+      return arr;
     },
+    { enabled: !!contract },
```
jnsdls, line 19: since ur updating this, also move it to `useQuery({...})` notation please

jnsdls (changes requested):

apps/dashboard/src/contract-ui/hooks/useContractSources.ts
```diff
@@ -1,17 +1,25 @@
...
+      contract?.chain.id || "",
+      contract?.address || "",
+    ],
+    queryFn: async (): Promise<Array<{ filename: string; source: string }>> => {
+      invariant(contract, "contract is required");
+      const data = await getCompilerMetadata(contract);
+      const sources = data.metadata.sources || {};
+      const arr = Object.keys(sources).map((key) => ({
+        filename: key,
+        source: sources[key]?.content || "",
```
joaquim-verges, line 19: this is not enough to get the sources. if you look at the implementation in v4, we actually check if there's an url for the attached source (ipfs hash) and download the sources in there

joaquim-verges (changes requested):
this wont work for a lot of contracts that have sources a ipfs links rather than flat content in the compiler metadata

apps/dashboard/src/contract-ui/hooks/useContractSources.ts
kien-ngo: ah I see, ty ser!

joaquim-verges (approved):

apps/dashboard/src/contract-ui/hooks/useContractSources.ts
```diff
@@ -1,17 +1,48 @@
...
+      contract?.address || "",
+    ],
+    queryFn: async (): Promise<Array<{ filename: string; source: string }>> => {
+      invariant(contract, "contract is required");
+      const data = await getCompilerMetadata(contract);
+      const sources = data.metadata.sources || {};
+      return await Promise.all(
+        // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+        Object.entries(sources).map(async ([path, info]: [string, any]) => {
+          const urls = info.urls as string[];
```
jnsdls, line 24: too many `as <x>` here, please type this properly, it seems like `getCompilerMetadata` should be typed correctly -> that should solve things?

apps/dashboard/src/contract-ui/hooks/useContractSources.ts
```diff
@@ -1,17 +1,48 @@
...
+          const ipfsLink = urls
+            ? urls.find((url) => url.includes("ipfs"))
+            : undefined;
+          if (ipfsLink) {
+            const ipfsHash = ipfsLink.split("ipfs/")[1];
+            const source = await download({
+              uri: `ipfs://${ipfsHash}`,
+              client: thirdwebClient,
+            }).then((r) => r.text());
```
jnsdls, line 33: what should happen if a single source fails?

jnsdls (changes requested):

apps/dashboard/src/contract-ui/hooks/useContractSources.ts
kien-ngo: We didn't handle it in v4. I can do `.catch(() => "Failed to fetch source")`

apps/dashboard/src/contract-ui/hooks/useContractSources.ts
kien-ngo: It's because we have `any` in the return type from the SDK:
```
export type PublishedMetadata = {
  name: string;
  abi: Abi;
  metadata: Record<string, any>; // <-------------------
  info: {
    title?: string;
    author?: string;
    details?: string;
    notice?: string;
  };
  licenses: string[];
  isPartialAbi?: boolean;
};
```

How/where can I get the type for this?

jnsdls: probably need to type it correctly inside SDK then

packages/thirdweb/src/contract/actions/compiler-metadata.ts
```diff
@@ -7,7 +7,9 @@ export type PublishedMetadata = {
   name: string;
   abi: Abi;
   // biome-ignore lint/suspicious/noExplicitAny: TODO: fix later
-  metadata: Record<string, any>;
+  metadata: Record<string, any> & {
```
jnsdls, line 10: I believe we should type this based on [the spec](https://docs.soliditylang.org/en/latest/metadata.html) ?

jnsdls (changes requested):

packages/thirdweb/src/contract/actions/compiler-metadata.ts
jnsdls: @joaquim-verges what do u think? I think we should be able to rely on this right?

joaquim-verges: yes we should

joaquim-verges: wouldnt block on it though - can you leave a TODO and link to the spec @kien-ngo

kien-ngo: Yep got it

PR #4150, opened 2024-08-17
[SDK] resolveAddress for Lens handles

kien-ngo:
## PR-Codex overview
The focus of this PR is to enhance the `Lens` extension in the `thirdweb` package by adding functionality to resolve addresses based on lens handles.

### Detailed summary
- Added `resolveAddress` function to resolve wallet addresses for lens handles
- Updated contract address overrides for various lens-related contracts
- Improved testing for address resolution functionality

### DIFF
--- a/packages/thirdweb/src/extensions/lens/read/resolveAddress.test.ts
+++ b/packages/thirdweb/src/extensions/lens/read/resolveAddress.test.ts
@@ -1,0 +1,23 @@
+import { describe, expect, it } from "vitest";
+import { TEST_CLIENT } from "~test/test-clients.js";
+import { isAddress } from "../../../utils/address.js";
+import { resolveAddress } from "./resolveAddress.js";
+
+describe("resolve lens address", () => {
+  it("should resolve to correct address", async () => {
+    const address = await resolveAddress({
+      client: TEST_CLIENT,
+      name: "captain_jack",
+    });
+    // "captain_jack" is a valid localname so it should definitely resolve to a valid address
+    expect(isAddress(address)).toBe(true);
+  });
+
+  it("should throw if passed an invalid lens handle", async () => {
+    await expect(() =>
+      resolveAddress({ client: TEST_CLIENT, name: "vitalik.eth" }),
+    ).rejects.toThrowError(
+      "Could not fetch the wallet address for lens handle: vitalik.eth",
+    );
+  });
+});
--- a/packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
+++ b/packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
@@ -1,0 +1,121 @@
+import type { Chain } from "../../../chains/types.js";
+import type { ThirdwebClient } from "../../../client/client.js";
+import { ZERO_ADDRESS } from "../../../constants/addresses.js";
+import { getAddress, isAddress } from "../../../utils/address.js";
+import { withCache } from "../../../utils/promise/withCache.js";
+
+/**
+ * @extension LENS
+ */
+export type ResolveLensAddressParams = {
+  /**
+   * Either a full lens handle, or a lens localName
+   * Examples:
+   * 1. Full handle: "lens/vitalik"
+   * 2. Just local name: "vitalik"
+   */
+  name: string;
+  /**
+   * a [`ThirdwebClient`](https://portal.thirdweb.com/references/typescript/v5/ThirdwebClient)
+   */
+  client: ThirdwebClient;
+  /**
+   * Override parameters for Lens contract, defaults to LensHandle contract on Polygon
+   */
+  overrides?: {
+    /**
+     * The contract address of the [`LensHandle contract`](https://www.lens.xyz/docs/resources/smart-contracts)
+     */
+    lensHandleContractAddress?: string;
+    /**
+     * The chain which `lensHandleContractAddress` is deployed on
+     */
+    chain?: Chain;
+  };
+};
+
+/**
+ * Take in a Lens handle or local-name and return the wallet address behind that handle/local-name.
+ * For example, "lens/vitalik" is a handle, with "lens" being the namespace and "vitalik" being the local name
+ * @extension LENS
+ * @example
+ * ```ts
+ * import { resolveAddress } from "thirdweb/extensions/lens";
+ *
+ * const walletAddress = await resolveAddress({
+ *   name: "vitalik",
+ *   client,
+ * });
+ * ```
+ */
+export async function resolveAddress(
+  options: ResolveLensAddressParams,
+): Promise<string> {
+  const { name, overrides, client } = options;
+  if (isAddress(name)) {
+    return getAddress(name);
+  }
+  return withCache(
+    async (): Promise<string> => {
+      const [
+        { getContract },
+        { getTokenId },
+        { ownerOf },
+        { polygon },
+        { LENS_HANDLE_ADDRESS },
+      ] = await Promise.all([
+        import("../../../contract/contract.js"),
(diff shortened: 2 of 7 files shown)


packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
```diff
@@ -0,0 +1,80 @@
...
+import { getAddress, isAddress } from "../../../utils/address.js";
+
+/**
+ * @extension LENS
+ */
+export type ResolveLensAddressParams = {
+  handleOrLocalName: string;
+  client: ThirdwebClient;
+  overrides?: {
+    lensHandleAddress?: string;
```
joaquim-verges, line 12: lensHandleContractAddress i assume? otherwise its confusing.

can you also add ts doc to each prop?

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Take in a Lens handle or local-name and return the wallet address behind that handle/local-name.
+ * For example, "lens/vitalik" is a handle, with "lens" being the namespace and "vitalik" being the local name
+ * @extension LENS
+ * @example
+ * ```ts
+ * import { resolveAddress } from "thirdweb/extensions/lens";
+ *
+ * const walletAddress = await resolveAddress("vitalik");
```
joaquim-verges, line 25: example is wrong right? should be an object

kien-ngo: oops I'll fix

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
```diff
@@ -0,0 +1,80 @@
+import type { Chain } from "../../../chains/types.js";
+import type { ThirdwebClient } from "../../../client/client.js";
+import { getAddress, isAddress } from "../../../utils/address.js";
+
+/**
+ * @extension LENS
+ */
+export type ResolveLensAddressParams = {
+  handleOrLocalName: string;
```
joaquim-verges, line 9: i think i would just call this name, to stay inline with the ENS one.

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
```diff
@@ -0,0 +1,80 @@
...
+    throw new Error(`Could not retrieve tokenId from localName: ${localName}`);
+  }
+
+  /**
+   * Another thing to note is that even if you enter an invalid localName,
+   * `getTokenId` still returns you a tokenId - so never rely on the result alone.
+   * Check if the tokenId truly exists using `exists` or in this case, `ownerOf`
+   */
+
+  return await ownerOf({ contract, tokenId });
```
joaquim-verges, line 79: hmm what does this return if there's no owner? zero address? throw?

kien-ngo: I want it to throw. Maybe I should edit the code here to make the error message meaningful

kien-ngo: updated!

packages/thirdweb/src/extensions/lens/read/resolveAddress.test.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, expect, it } from "vitest";
+import { TEST_CLIENT } from "~test/test-clients.js";
+import { isAddress } from "../../../utils/address.js";
+import { resolveAddress } from "./resolveAddress.js";
+
+describe("resolve lens address", () => {
+  it("should resolve to correct address", async () => {
+    const address = await resolveAddress({
+      client: TEST_CLIENT,
+      handleOrLocalName: "captain_jack",
```
joaquim-verges, line 10: need to update to name in the test too

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
```diff
@@ -0,0 +1,106 @@
...
+    throw new Error(`Could not retrieve tokenId from localName: ${localName}`);
+  }
+
+  /**
+   * Another thing to note is that even if you enter an invalid localName,
+   * `getTokenId` still returns you a tokenId - so never rely on the result alone.
+   * Check if the tokenId truly exists using `exists` or in this case, `ownerOf`
+   */
+  const address = await ownerOf({ contract, tokenId }).catch(() => "");
+  if (!address) {
```
joaquim-verges, line 100: i believe this can return zero address too?

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
```diff
@@ -0,0 +1,106 @@
...
+    throw new Error(
+      `Could not fetch the owner address for tokenId: ${tokenId}`,
```
joaquim-verges, line 102: maybe make it a lens specific error msg, like 'could not find user ${name}`

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
```diff
@@ -0,0 +1,106 @@
...
+   */
+  name: string;
+  /**
+   * a [`ThirdwebClient`](https://portal.thirdweb.com/references/typescript/v5/ThirdwebClient)
+   */
+  client: ThirdwebClient;
+  /**
+   * Override parameters for Lens contract, defaults to LensHandle contract on Polygon
+   */
+  overrides?: {
```
joaquim-verges, line 23: do we really need overrides nesting here? just add them inline imo since they're optional

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
```diff
@@ -0,0 +1,106 @@
...
+ * ```
+ */
+export async function resolveAddress(
+  options: ResolveLensAddressParams,
+): Promise<string> {
+  const { name, overrides, client } = options;
+  if (isAddress(name)) {
+    return getAddress(name);
+  }
+  const [
```
joaquim-verges, line 56: you should add a cache to this since it doesnt change often. look how we do it in the ENS resolveAddress

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
kien-ngo: I feel like it should either be something or nothing. But I want to know more about your thought process (why should we have zero address here)

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
kien-ngo: So in the other Lens methods that have more than one override-able contracts, I group them all into an object to signify that "all these contracts have to be on the same chain".

For this one I keep the same pattern for consistency - but I'm okay with either way

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
joaquim-verges: I'm saying you should throw here if ownerOf is zero address since I had the behavior you're looking for no?

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
joaquim-verges: Ok that makes sense

packages/thirdweb/src/extensions/lens/read/resolveAddress.ts
kien-ngo: ah okay :) updating the code

joaquim-verges (approved):

PR #4177, opened 2024-08-19
[Portal] Update some docs

kien-ngo:
## Problem solved

---

## PR-Codex overview
The focus of this PR is to introduce gasless transaction configurations and transaction annotations across different modules.

### Detailed summary
- Added gasless transaction configurations for Engine, OpenZeppelin, and Biconomy providers
- Updated transaction annotations in various files
- Exposed GaslessOptions type for gasless transactions configuration

### DIFF
--- a/.changeset/empty-swans-smoke.md
+++ b/.changeset/empty-swans-smoke.md
@@ -1,0 +1,12 @@
+---
+"thirdweb": patch
+---
+
+Expsose GaslessConfigs type
+```ts
+const gaslessOptions: GaslessOptions = {
+  provider: "engine",
+  relayerUrl: "https://thirdweb.engine-***.thirdweb.com/relayer/***",
+  relayerForwarderAddress: "0x...",
+}
+```
--- a/packages/thirdweb/src/transaction/read-contract.ts
+++ b/packages/thirdweb/src/transaction/read-contract.ts
@@ -70,7 +70,7 @@ export type ReadContractOptions<
 >;
 
 /**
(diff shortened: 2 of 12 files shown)


.changeset/empty-swans-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+"thirdweb": patch
+---
+
+Expsose GaslessConfigs type
```
gregfromstl, line 5: Could you include a code snippet or type examples here? When we expose previously internal functions of types we have to treat it like its a net new feature (because it is from the public API POV)

packages/thirdweb/src/transaction/read-contract.ts
```diff
@@ -111,7 +111,26 @@ export type ReadContractOptions<
...
+ *
+ * const contract = getContract({
+ *   client,
+ *   address: "0x...",
+ *   chain: sepolia,
+ * });
+ *
+ * const { data, isLoading } = useReadContract({
+ *   contract,
+ *   method: resolveMethod("tokenURI"),
```
gregfromstl, line 133: Does this work by itself without an ABI on the contract?

gregfromstl (approved):
LGTM just a few small comments/questons

packages/thirdweb/src/transaction/read-contract.ts
kien-ngo: yes! but we lose the type-safe feature. The result will be of type "unknown"

PR #4236, opened 2024-08-21
[Docs] Feature: Automatically unroll function inputs and outputs

gregfromstl:
## Problem solved

Previously we made no attempt to unroll function inputs and outputs in the docs. This made things difficult to quickly scan since every function has a typed input and output object you would have to navigate to.

This PR is a first attempt at basic unrolling functionality of these typed inputs and outputs.

Before:
Screenshot 2024-08-21 at 3 20 24 PM

After:
Screenshot 2024-08-21 at 3 20 49 PM

As an additional readability bonus, the input and output sections now start as an open accordion.

Note: There are still improvements to be made especially on functions with more complex types. These functions' interfaces should be simplified in future versions.

## Next Steps
Future features should add nested input descriptions (supported by typedoc but not `typedoc-better-json`). Future SDK versions should use simpler, explicitly typed inputs and outputs to simplify the unrolling process (and easier LSP support for users).

## To Test
Open the preview URL, click around to various functions and inspect the parameter and return types

---

## PR-Codex overview
This PR enhances TypeScript documentation fetching, refactors transaction options to an interface, and adds detailed type information to function components.

### Detailed summary
- Added `fetchTypeScriptDoc` and `getSlugToDocMap` functions for TypeScript documentation fetching
- Refactored `SendTransactionOptions` to an interface with detailed comments
- Improved type information rendering in function components with `ReturnsTDoc` and `ParameterTDoc` functions

### DIFF
--- a/apps/portal/src/app/references/components/TDoc/Function.tsx
+++ b/apps/portal/src/app/references/components/TDoc/Function.tsx
@@ -3,6 +3,9 @@ import {
 	FunctionParameter,
 	FunctionSignature,
 	getFunctionSignature,
+    InterfaceDoc,
+    Summary,
+    TypeInfo,
 } from "typedoc-better-json";
 import { CodeBlock, InlineCode } from "../../../../components/Document/Code";
 import { TypedocSummary } from "./Summary";
@@ -16,6 +19,7 @@ import invariant from "tiny-invariant";
 import { getTags } from "./utils/getTags";
 import { getTokenLinks } from "./utils/getTokenLinks";
 import { DocLink, Paragraph } from "@/components/Document";
+import fetchDocBySlug from "./fetchDocs/fetchDocBySlug";
 
 export function FunctionTDoc(props: {
 	doc: FunctionDoc;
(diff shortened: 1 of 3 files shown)


jnsdls (commented):
doesn't seem to work reliably?: https://docs-v2-bt6g06ruj.thirdweb-preview.com/references/typescript/v5/BuyDirectListingButton (just one example, click around in the reference its most (?) things)

random concerns:
- this loses the actual type name which may be relevant to import for users?
- ideally we'd do something like https://shikijs.github.io/twoslash/ which gives us best of both worlds here?

gregfromstl:
> doesn't seem to work reliably?: https://docs-v2-bt6g06ruj.thirdweb-preview.com/references/typescript/v5/BuyDirectListingButton (just one example, click around in the reference its most (?) things)
> 
> random concerns:
> 
> * this loses the actual type name which may be relevant to import for users?
> * ideally we'd do something like https://shikijs.github.io/twoslash/ which gives us best of both worlds here?

We keep the type name in the function signature section.

What's wrong with the example you sent? We can't drill all the way down for union types (at least not without modifying the typedoc json converter, which is a bigger lift), but we can't at least get the main arguments so they have some idea of what to put. This essentially shows what they would see if they clicked into the type name, but on the same page.

twoslash looks nice, but it wouldn't solve the underlying problem, which is deeply abstracted types that we don't parse all the way down.

jnsdls:
oh, I assumed it would unfurl it everywhere - that seems OK then

jnsdls (approved):

PR #4270, opened 2024-08-23
[TS SDK] Feature: Accept a string for signature mint UID

gregfromstl:
## Problem solved

Auto-encodes strings passed to `generateSignatureMint` across all token types. This change could cause confusion since the response uid from the contract will still be in hex form.

---

## PR-Codex overview
This PR focuses on enhancing the signature mint functionality by allowing the acceptance of a string for UID generation.

### Detailed summary
- Updated functions to accept string UIDs for minting signatures
- Added logic to automatically encode provided string UIDs
- Improved testing for mint signature generation with custom values

### DIFF
--- a/packages/thirdweb/src/extensions/erc1155/write/sigMint.ts
+++ b/packages/thirdweb/src/extensions/erc1155/write/sigMint.ts
@@ -8,7 +8,7 @@ import type { ThirdwebContract } from "../../../contract/contract.js";
 import type { BaseTransactionOptions } from "../../../transaction/types.js";
 import { toBigInt } from "../../../utils/bigint.js";
 import { dateToSeconds, tenYearsFromNow } from "../../../utils/date.js";
-import type { Hex } from "../../../utils/encoding/hex.js";
+import { type Hex, isHex, stringToHex } from "../../../utils/encoding/hex.js";
 import type { NFTInput } from "../../../utils/nft/parseNft.js";
 import { randomBytesHex } from "../../../utils/random.js";
 import type { Account } from "../../../wallets/interfaces/wallet.js";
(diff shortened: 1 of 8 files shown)


packages/thirdweb/src/extensions/erc1155/write/sigMint.ts
```diff
@@ -135,7 +135,14 @@ export async function generateMintSignature(
       return "";
     })(),
     // uid computation
-    mintRequest.uid || (await randomBytesHex()),
+    ((): Hex => {
+      if (mintRequest.uid) {
+        return isHex(mintRequest.uid)
```
arcoraven, line 140: A uid that is hex but not bytes32 will fail right?

packages/thirdweb/src/extensions/erc1155/write/sigMint.ts
```diff
@@ -135,7 +135,14 @@ export async function generateMintSignature(
...
+          ? mintRequest.uid
+          : stringToHex(mintRequest.uid, { size: 32 });
```
arcoraven, line 142: Just confirming this produces a 32 byte hex, not a 32 char hex

arcoraven (approved):

packages/thirdweb/src/extensions/erc1155/write/sigMint.ts
gregfromstl: Not here but eventually yes. We could try to cast it to bytes then back to hex but that feels like too much

packages/thirdweb/src/extensions/erc1155/write/sigMint.ts
gregfromstl: Correct

joaquim-verges (approved):

PR #4271, opened 2024-08-23
react: add overrides for connected wallet name and avatar in details button and modal

alecananian:
- Adds override params for the connected wallet name and avatar in the details button and modal
- Adds param to hide the connected wallet image over the avatar
- Fixes extraneous call for in-app wallet info when connected to smart wallet

Rationale: at Treasure, we use our own naming system called [TreasureTag](https://market.treasure.lol/treasuretag) and would like to display this tag in the connected wallet details modal as an alternative to ENS

Examples:

Screenshot 2024-08-23 at 4 35 57 PM

Screenshot 2024-08-23 at 4 40 07 PM

### DIFF
--- a/packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
+++ b/packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
@@ -256,6 +256,19 @@ export type ConnectButton_detailsModalOptions = {
   hideDisconnect?: boolean;
 
   /**
+   * Hide the wallet image on the "Switch Wallet" button in the `ConnectButton` Details Modal.
+   *
+   * By default it is `false`
+   * @example
+   * ```tsx
+   * <ConnectButton detailsModal={{
+   *  hideWalletImage: true
+   * }} />
+   * ```
+   */
+  hideWalletImage?: boolean;
+
+  /**
    * Render custom UI at the bottom of the `ConnectButton` Details Modal
    * @param props - props passed to the footer component which includes a function to close the modal
    * @example
@@ -278,6 +291,16 @@ export type ConnectButton_detailsModalOptions = {
    * thirdweb Pay allows users to buy tokens using crypto or fiat currency.
    */
   payOptions?: Extract<PayUIOptions, { mode?: "fund_wallet" }>;
+
+  /**
+   * Render custom UI for the connected wallet name in the `ConnectButton` Details Modal, overriding ENS name or wallet address.
+   */
+  connectedWalletName?: React.ReactNode;
+
+  /**
+   * Use custom avatar URL for the connected wallet image in the `ConnectButton` Details Modal, overriding ENS avatar or Blobbie icon.
+   */
+  connectedWalletAvatarUrl?: string;
 };
 
 /**
@@ -325,6 +348,16 @@ export type ConnectButton_detailsButtonOptions = {
    * ```
    */
   displayBalanceToken?: Record<number, string>;
+
+  /**
+   * Render custom UI for the connected wallet name in the `ConnectButton` details button, overriding ENS name or wallet address.
+   */
+  connectedWalletName?: React.ReactNode;
+
+  /**
+   * Use custom avatar URL for the connected wallet image in the `ConnectButton` details button, overriding ENS avatar or Blobbie icon.
+   */
+  connectedWalletAvatarUrl?: string;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
```diff
@@ -278,6 +291,16 @@ export type ConnectButton_detailsModalOptions = {
    * thirdweb Pay allows users to buy tokens using crypto or fiat currency.
    */
   payOptions?: Extract<PayUIOptions, { mode?: "fund_wallet" }>;
+
+  /**
+   * Render custom UI for the connected wallet name in the `ConnectButton` Details Modal, overriding ENS name or wallet address.
+   */
+  connectedWalletName?: React.ReactNode;
```
joaquim-verges, line 298: wondering if it would make more sene to let you pass a component for the entire row, so you can do your own avatar + name + balance style. would just be a single prop and gives you max flexibility. wdyt?

alecananian: @joaquim-verges that sounds great as well!

alecananian: I'd just want to be able to preserve the "switch wallet" on-click functionality at times

joaquim-verges: hmmm in that case maybe we just expose both name and avatar as you have it, but allow for `string | React.ReactNode` for both. Let us discuss it tomorrow and we'll try to get this in asap

alecananian: Sounds good! Also [`React.ReactNode` covers `string`](https://reactnative.dev/docs/react-node), so that's why I haven't included it explicitly here.

packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
```diff
@@ -255,6 +255,19 @@ export type ConnectButton_detailsModalOptions = {
...
+   *
+   * By default it is `false`
+   * @example
+   * ```tsx
+   * <ConnectButton detailsModal={{
+   *  hideWalletImage: true
+   * }} />
+   * ```
+   */
+  hideWalletImage?: boolean;
```
joaquim-verges, line 269: do you actually need this? worried its gonna look the layout look very weird

packages/thirdweb/src/react/web/ui/ConnectWallet/Details.tsx
```diff
@@ -226,18 +229,14 @@ export const ConnectedWalletDetails: React.FC<{
...
-          <Text
-            size="xs"
-            color="primaryText"
-            weight={500}
-            className={`${TW_CONNECTED_WALLET}__address`}
-          >
-            {addressOrENS}
-          </Text>
-        ) : (
-          <Skeleton height={fontSize.xs} width="80px" />
```
joaquim-verges, line 239: any reason you removed this?

packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
joaquim-verges: i think we'll go with this, but might rename to `connectedAccountName` and `connectedAccountAvatarUrl`

packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
alecananian: Nice to have - it just covers part of the avatar with its current positioning. I'm not sure if there's some other use of this that I'm missing that would mess up the layout.

Screenshot 2024-08-26 at 4 04 15 PM

packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
alecananian: Sure thing, the param naming is not important to me 🙂

packages/thirdweb/src/react/web/ui/ConnectWallet/Details.tsx
alecananian: It seems like `addressOrENS` should always have a value, so I removed the condition + skeleton component when adding the `connectedWalletName` into the mix

packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
joaquim-verges: i'll combine this with `hideSwitchWallet` instead, if you turn off the button behavior, then it hides the icon too

joaquim-verges (approved):

PR #4289, opened 2024-08-27
feat: AWS KMS Signer

d4mr:
## PR-Codex overview
This PR adds AWS KMS integration for secure key management, including configuration setup, account creation, and signing operations.

### Detailed summary
- Added AWS KMS signer library for secure key management
- Updated package dependencies for AWS SDK and vitest
- Implemented AWS KMS configuration setup and account creation
- Added tests for AWS KMS integration with sign, verify, and transaction functions

> The following files were skipped due to too many changes: `pnpm-lock.yaml`


packages/thirdweb/src/wallets/aws-kms.ts
```diff
@@ -0,0 +1,129 @@
...
+type AwsKmsAccountOptions = {
+  keyId: string;
+  config?: KMSClientConfig;
+  client: ThirdwebClient;
+};
+
+export async function getAwsKmsAccount(
+  options: AwsKmsAccountOptions,
+): Promise<Account> {
+  if (typeof Buffer === "undefined") {
```
d4mr, line 39: because we don't polyfill Buffer, and we don't expect folks to use AWS KMS on browsers, I have explicitly safeguarded against browser usage

packages/thirdweb/src/wallets/aws-kms.ts
```diff
@@ -0,0 +1,129 @@
...
+    throw new Error("AwsKmsAccount only works in Node.js environment");
+  }
+
+  const { keyId, config, client } = options;
+  const signer = new KmsSigner(keyId, config);
+
+  // Populate address immediately
+  const addressUnprefixed = await signer.getAddress();
+  const address = `0x${addressUnprefixed}` as Address;
```
gregfromstl, line 48: Use `getAddress` for type validation here

packages/thirdweb/src/wallets/aws-kms.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  async function signTransaction(tx: TransactionSerializable): Promise<Hex> {
+    const serializedTx = serializeTransaction({ transaction: tx });
+    const txHash = keccak256(serializedTx);
+
+    // we don't polyfill buffer, but signer.sign explicitly requires a buffer
+    // what do we do here?
```
gregfromstl, line 55: Was this resolved with your check above?

packages/thirdweb/src/wallets/aws-kms.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  async function signMessage({
+    message,
+  }: {
+    message: SignableMessage;
+  }): Promise<Hex> {
+    let messageHash: Hex;
+    if (typeof message === "string") {
+      const prefixedMessage = `\x19Ethereum Signed Message:\n${message.length}${message}`;
```
gregfromstl, line 83: What's this prefix?

packages/thirdweb/src/wallets/aws-kms.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+let account: Awaited<ReturnType<typeof getAwsKmsAccount>>;
+
+const anvilTestClient = createTestClient({
+  mode: "anvil",
+  transport: http(ANVIL_CHAIN.rpc),
+});
+
+beforeAll(async () => {
+  account = await getAwsKmsAccount({
```
gregfromstl, line 28: So this will always hit a live KMS account? And is there any way to keep that address constant?

gregfromstl (changes requested):

packages/thirdweb/src/wallets/aws-kms.ts
```diff
@@ -0,0 +1,129 @@
+import type { KMSClientConfig } from "@aws-sdk/client-kms";
+import { KmsSigner } from "aws-kms-signer";
+import type {
```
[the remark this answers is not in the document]
d4mr: just used `private-key.ts` as reference, is there a different import I should do?

packages/thirdweb/src/wallets/aws-kms.test.ts
d4mr: Same KMS keyId would get you the same address. One way to enforce this on a test would be to also expose `TEST_AWS_KMS_EXPECTED_ADDRESS` and verify it matches.

packages/thirdweb/src/wallets/aws-kms.ts
d4mr: This is the required prefix for `eth_sign` & `personal_sign` according to spec (EIP191): https://github.com/ethereum/go-ethereum/issues/3731#issuecomment-283620075

Although this raises another question:
`SignableMessage` can be `string` or `{ raw: Hex | ByteArray }`. If a raw `SignableMessage` is passed along, I am not adding a prefix, because the expectation is that its already processed.

This is based on a large assumption I made around the interface. I am not certain this is the expectation from the interface. What do you think?

(to be clear, signing a message requires this prefix. If you sign without this prefix and try to `verifyMessage` it will fail)

packages/thirdweb/src/wallets/aws-kms.ts
d4mr: yes! removing comments

gregfromstl:
Stale, closing

PR #4295, opened 2024-08-27
[React] Feature: Adds TOS confirmation checkbox

gregfromstl:
## Problem solved

A developer needs the ability to require all users to accept their TOS before creating an in-app wallet so they have the ability to use their emails. This PR adds a flag `requireApproval` on the modal and embed options. When enabled, all login methods will be disabled until the checkbox is selected.
Screenshot 2024-08-27 at 1 41 07 PM

## To Test
- Open playground Sign-In page
- Under "Modal Options", set "Require TOS approval" to true
- Attempt to login with and without checking the approval confirmation

---

## PR-Codex overview
This PR adds a feature to require Terms of Service or Privacy Policy approval before connecting to an in-app wallet.

### Detailed summary
- Added `requireApproval` prop to enforce TOS or Privacy Policy acceptance
- Updated components to handle the new prop for wallet connection approval

> The following files were skipped due to too many changes: `packages/thirdweb/src/react/web/ui/ConnectWallet/WalletSelector.tsx`

### DIFF
--- a/apps/playground-web/src/app/connect/sign-in/button/LeftSection.tsx
+++ b/apps/playground-web/src/app/connect/sign-in/button/LeftSection.tsx
@@ -146,22 +146,41 @@ export function LeftSection(props: {
             </div>
           </div>
 
-          {/* Thirdweb Branding */}
-          <section className="flex items-center gap-2">
-            <Checkbox
-              id="show-thirdweb-branding"
-              checked={connectOptions.ShowThirdwebBranding}
-              onCheckedChange={(checkState) => {
-                setConnectOptions((v) => ({
-                  ...v,
-                  ShowThirdwebBranding: checkState === true,
-                }));
-              }}
-            />
-            <Label htmlFor="show-thirdweb-branding">
-              Show thirdweb branding
-            </Label>
-          </section>
+          <div className="grid grid-cols-1 md:grid-cols-2 gap-6 md:gap-4">
+            {/* Thirdweb Branding */}
+            <section className="flex items-center gap-2">
+              <Checkbox
+                id="show-thirdweb-branding"
+                checked={connectOptions.ShowThirdwebBranding}
+                onCheckedChange={(checkState) => {
+                  setConnectOptions((v) => ({
+                    ...v,
+                    ShowThirdwebBranding: checkState === true,
+                  }));
+                }}
+              />
+              <Label htmlFor="show-thirdweb-branding">
+                Show thirdweb branding
+              </Label>
+            </section>
+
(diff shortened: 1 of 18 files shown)


packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
```diff
@@ -387,6 +387,19 @@ export type ConnectButton_connectModalOptions = {
...
+   *
+   * By default it's `false`
+   * @example
+   * ```tsx
+   * <ConnectButton connectModal={{
+   *  requireApproval: true
+   * }} />
+   * ```
+   */
+  requireApproval?: boolean;
```
MananTank, line 401: If this only affects in-app wallet - shouldn't this option be in `inAppWallet()` ?

joaquim-verges: seems to me more related to the TOS / policy links you can add no? dont feel like its inapp specific

joaquim-verges (approved):
checkbox feels a little hidden / not clear i have to check it to continue. But we can iterate on the UI - option to require a user action makes sense

packages/thirdweb/src/react/core/hooks/connection/ConnectButtonProps.ts
gregfromstl: Yeah we should minimize UI stuff on the inAppWallet constructor

MananTank: In that case, we should disable connecting all the wallets if TOS is not checked right - and not just inApp wallet?

MananTank: Also - noticed that the passkey option is not disabled

apps/playground-web/src/app/connect/sign-in/button/LeftSection.tsx
```diff
@@ -146,22 +146,41 @@ export function LeftSection(props: {
...
+              />
+              <Label htmlFor="show-thirdweb-branding">
+                Show thirdweb branding
+              </Label>
+            </section>
+
+            {/* Require Approval */}
+            <section className="flex items-center gap-2">
+              <Checkbox
+                id="show-thirdweb-branding"
```
MananTank, line 170: Need to change the Checkbox Id here - clicking on "Require TOS approval" text is toggling the state of "Show thirdweb branding" checkbox in the playground right now

PR #4324, opened 2024-08-29
Kien/explorer fn overload

kien-ngo:
## Problem solved

### DIFF
--- a/apps/dashboard/src/components/contract-functions/contract-function.tsx
+++ b/apps/dashboard/src/components/contract-functions/contract-function.tsx
@@ -456,13 +456,20 @@ const FunctionsOrEventsListItem: React.FC<FunctionsOrEventsListItemProps> = ({
         }
         onClick={() => {
           setSelectedFunction(fn);
-          const { name } = fn;
           const path = router.asPath.split("?")[0];
           // Only apply to the Explorer page
           if (path.endsWith("/explorer")) {
-            router.push({ pathname: path, query: { name } }, undefined, {
-              shallow: true,
-            });
+            const { name } = fn;
+            /**
+             * Using function name is incorrect because we have function overloads in solidity
+             * using router shallow pushing in this context will actually re-render the page
+             * and cause the functions with same name to render incorrectly.
+             *
+             * todo: fully migrate this page to v5 and switch to use function's signature instead of name
+             */
+            const url = new URL(window.location.href);
+            url.searchParams.set("name", name);
+            window.history.pushState(null, "", url.toString());
           }
         }}
         color="heading"
--- a/apps/dashboard/src/contract-ui/tabs/listings/components/list-form.tsx
+++ b/apps/dashboard/src/contract-ui/tabs/listings/components/list-form.tsx
@@ -215,6 +215,7 @@ export const CreateListingsForm: React.FC<CreateListingsFormProps> = ({
             try {
               await sendAndConfirmTx.mutateAsync(approveTx);
             } catch {
+              setIsFormLoading(false);
               return toast.error("Failed to approve NFT for marketplace");
             }
           }


apps/dashboard/src/components/contract-functions/contract-function.tsx
```diff
@@ -456,13 +456,20 @@ const FunctionsOrEventsListItem: React.FC<FunctionsOrEventsListItemProps> = ({
...
+            /**
+             * Using function name is incorrect because we have function overloads in solidity
+             * using router shallow pushing in this context will actually re-render the page
+             * and cause the functions with same name to render incorrectly.
+             *
+             * todo: fully migrate this page to v5 and switch to use function's signature instead of name
+             */
+            const url = new URL(window.location.href);
+            url.searchParams.set("name", name);
+            window.history.pushState(null, "", url.toString());
```
joaquim-verges, line 472: hmmm doesnt the fn name come with the overloaded params already? if not can we not add it?

apps/dashboard/src/components/contract-functions/contract-function.tsx
```diff
@@ -456,13 +467,24 @@ const FunctionsOrEventsListItem: React.FC<FunctionsOrEventsListItemProps> = ({
...
+            // In case there are more than 1 fn with the same name, distinguish them by their inputs
+            if (isOverloadFn) {
+              url.searchParams.set(
+                "inputs",
+                fn.inputs.map((i) => i.name).join(","),
+              );
+            } else {
+              url.searchParams.delete("inputs");
+            }
+            window.history.pushState(null, "", url.toString());
```
jnsdls, line 487: have to use nextjs router here

apps/dashboard/src/components/contract-functions/contract-function.tsx
```diff
@@ -456,13 +467,24 @@ const FunctionsOrEventsListItem: React.FC<FunctionsOrEventsListItemProps> = ({
...
+            // Check if there are more than 1 function with the same name
+            const isOverloadFn =
+              (allFunctions || []).filter((o) => o.name === fn.name).length > 1;
+            const url = new URL(window.location.href);
+            url.searchParams.set("name", fn.name);
+            // In case there are more than 1 fn with the same name, distinguish them by their inputs
+            if (isOverloadFn) {
+              url.searchParams.set(
+                "inputs",
+                fn.inputs.map((i) => i.name).join(","),
```
jnsdls, line 482: `name` is not guaranteed to exist, probably need to use the type or something here, maybe just the human readable abi or the fn selector?

jnsdls (changes requested):

apps/dashboard/src/components/contract-functions/contract-function.tsx
kien-ngo: yes switching to fn signature - I can see that you have migrated this part to v5!

apps/dashboard/src/components/contract-functions/contract-function.tsx
```diff
@@ -20,7 +20,12 @@ import {
   Thead,
   Tr,
 } from "@chakra-ui/react";
-import type { Abi, AbiEvent, AbiFunction } from "abitype";
+import {
+  type Abi,
+  type AbiEvent,
+  type AbiFunction,
+  formatAbiItem,
```
jnsdls, line 27: instead use `toFunctionSelector` to generate the deterministic function selector -> then you can map back from that into the selected function inside the explorer

jnsdls: viem also exports `toFunctionSelector` and you can pass it any abiFunction

jnsdls (changes requested):

jnsdls (approved):
very nice

PR #4541, opened 2024-09-11
[SDK] Feature: adds `findUsers` function

gregfromstl:
This replaces https://portal.thirdweb.com/connect/in-app-wallet/how-to/get-in-app-wallet-details-on-server

---

## PR-Codex overview
This PR adds a `getUser` function to query users on the server and updates related files.

### Detailed summary
- Added `getUser` function to query users on the server
- Updated `wallets.ts` to export `getUser` and `GetUserResult`
- Added `getUser.ts` with implementation of `getUser` function

### DIFF
--- a/.changeset/tasty-nails-fry.md
+++ b/.changeset/tasty-nails-fry.md
@@ -1,0 +1,14 @@
+---
+"thirdweb": minor
+---
+
+Adds getUsers function to query users on the server
+
+```ts
+import { getUser } from "thirdweb/wallets";
+
+const user = await getUser({
+  client,
+  walletAddress: "0x123...",
+});
+```
--- a/packages/thirdweb/src/exports/thirdweb.ts
+++ b/packages/thirdweb/src/exports/thirdweb.ts
@@ -69,6 +69,14 @@ export {
 } from "../contract/contract.js";
 
 /**
+ * WALLETS
+ */
+export {
+  getUser,
+  type GetUserResult,
+} from "../wallets/in-app/core/users/getUser.js";
+
+/**
  * TRANSACTIONS
  */
 
--- a/packages/thirdweb/src/exports/wallets.ts
+++ b/packages/thirdweb/src/exports/wallets.ts
@@ -102,6 +102,11 @@ export {
   linkProfile,
 } from "../wallets/in-app/core/wallet/profiles.js";
 
+export {
+  getUser,
+  type GetUserResult,
+} from "../wallets/in-app/core/users/getUser.js";
+
 export type {
   CoinbaseWalletCreationOptions,
   CoinbaseSDKWalletConnectionOptions,
--- a/packages/thirdweb/src/react/core/hooks/others/useProfiles.ts
+++ b/packages/thirdweb/src/react/core/hooks/others/useProfiles.ts
@@ -18,7 +18,7 @@ import { useActiveWallet } from "../wallets/useActiveWallet.js";
  * const { data: profiles } = useProfiles();
  *
  * console.log("Type:", profiles[0].type); // "discord"
- * console.log("Email:", profiles[0].email); // "john.doe@example.com"
+ * console.log("Email:", profiles[0].details.email); // "john.doe@example.com"
  * ```
  *
  * @wallet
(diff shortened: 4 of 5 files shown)


packages/thirdweb/src/wallets/in-app/core/users/findUsers.ts
```diff
@@ -0,0 +1,110 @@
...
+  } else if (phone) {
+    url.searchParams.set("queryBy", "phone");
+    url.searchParams.set("phone", phone);
+  } else if (id) {
+    url.searchParams.set("queryBy", "id");
+    url.searchParams.set("id", id);
+  } else {
+    throw new Error(
+      "Please provide a walletAddress, email, phone, or id to query by. Note: You should never use your secret key on the client.",
+    );
```
jnsdls, line 82: could check if `typeof window !== "undefined"` -> throw this error at the top of the function (regarding no secret key on client)

packages/thirdweb/src/wallets/in-app/core/users/findUsers.ts
```diff
@@ -0,0 +1,110 @@
...
+    email?: string;
+    phone?: string;
+    id?: string;
+  }>
+>): Promise<FindUserResult[]> {
+  if (!client.secretKey) {
+    throw new Error(
+      "A secret key is required to query for users. If you're making this request from the server, please add a secret key to your client.",
+    );
+  }
```
jnsdls, line 61: below this I'd check if ur in browser -> throw

packages/thirdweb/src/wallets/in-app/core/users/findUsers.ts
```diff
@@ -0,0 +1,116 @@
...
+import type { OneOf, Prettify } from "../../../../utils/type-utils.js";
+import type { Profile } from "../authentication/types.js";
+
+export type FindUserResult = {
+  userId: string;
+  walletAddress: string;
+  email?: string;
+  phone?: string;
+  createdAt: string;
+  profiles: Profile[];
```
joaquim-verges, line 13: are you planning for this to be social profiles too? other i would just call these linkedAccounts i think to not confuse

.changeset/tasty-nails-fry.md
```diff
@@ -0,0 +1,14 @@
+---
+"thirdweb": minor
+---
+
+Adds findUsers function to query users on the server
+
+```ts
+import { findUsers } from "thirdweb/wallets";
+
+const users = await findUsers({
```
joaquim-verges, line 10: got confused by the name of the function.

how about `getUserDetails()` or `getUser()` especially since you can only get 1 user result

gregfromstl: The return type is an array (I'm not sure if that's just a quirk of the backend or you can actually get multiple)

packages/thirdweb/src/wallets/in-app/core/users/findUsers.ts
gregfromstl: We call them profiles already (and in all the copy around that feature) so I figured linkedAccounts would be more confusing. Eventually we should consolidate these and social profiles

.changeset/tasty-nails-fry.md
gregfromstl: Yeah you can get multiple because I could have two accounts with the same email (e.g. discord and google)

joaquim-verges: are you sure? im pretty sure if its the same email, it resolves to the same 'user' (same wallet address)

.changeset/tasty-nails-fry.md
```diff
@@ -0,0 +1,14 @@
+---
+"thirdweb": minor
+---
+
+Adds getUsers function to query users on the server
+
+```ts
+import { getUsers } from "thirdweb/wallets";
+
+const users = await getUsers({
```
joaquim-verges, line 10: still think its confusing to get an array here... i don't think i would expect that as a user and now i have to figure out which one in the array i care about

.changeset/tasty-nails-fry.md
gregfromstl: You're right, it autolinks if the emails are the same (even if I used a different auth method). Updated the function name

.changeset/tasty-nails-fry.md
gregfromstl: Oh good catch I just forgot to update the changelog. It doesn't return an array anymore

joaquim-verges (approved):

PR #4561, opened 2024-09-12
Fix no redirect to dashboard if user is already logged in and lands on login page

MananTank:
Did not go the middleware route here because even if we properly check the logged in status in middleware and redirect - it's possible for user to get stuck in that case if for some reason the wallet never gets into "connecting" status and so the useLoggedInUser query never runs and never redirects the user to login page ( and even if it did - they would immediately get redirected to other page because they have a "valid" auth cookie ) 

so safest solution here is to let the user land on login page and once then redirect from there after wallet connection + sign in is done

---

## PR-Codex overview
The focus of this PR is to refactor the login functionality in the dashboard app. 

### Detailed summary
- Removed unused imports and code related to `next/navigation`.
- Refactored `doLogin` function to only take payload as a parameter.
- Added `useDashboardRouter` hook for navigation handling.
- Implemented `onLoginSuccessful` function for successful login redirection.
- Updated `CustomConnectEmmbed` component to handle login and redirect logic.
- Added `isValidRedirectPath` function for validating redirect paths.

### DIFF
--- a/apps/dashboard/src/app/login/page.tsx
+++ b/apps/dashboard/src/app/login/page.tsx
@@ -2,6 +2,7 @@
 
 import { ColorModeToggle } from "@/components/color-mode-toggle";
 import { thirdwebClient } from "@/constants/client";
+import { useDashboardRouter } from "@/lib/DashboardRouter";
 import { useTheme } from "next-themes";
 import { useSearchParams } from "next/navigation";
 import { Suspense, useEffect, useState } from "react";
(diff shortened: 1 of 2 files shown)


MananTank:
* **#4561** <a href="https://app.graphite.dev/github/pr/thirdweb-dev/js/4561?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @MananTank and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

apps/dashboard/src/app/login/page.tsx
```diff
@@ -38,14 +39,39 @@ export default function LoginPage() {
...
-        isLoggedIn,
+        isLoggedIn: async (x) => {
+          const isLoggedInResult = await isLoggedIn(x);
+          if (isLoggedInResult) {
+            onLoginSuccessful();
+          }
+          return isLoggedInResult;
+        },
       }}
       client={thirdwebClient}
```
MananTank, line 50: We used to have a similar logic on the middleware - but that was simply checking for the existence of auth cookie - which is not enough to determine if the user is actually logged in 

Other alternative is to do the proper isLoggedIn check on the middleware if the user lands on /login page

jnsdls (approved):

PR #4609, opened 2024-09-16
Joe chainlist redirects

Joe-Thirdweb:
## Problem solved

Correcting chainlist icon redirects. 

Adding redirects:
      source: "/:network/connect-sdk",
      destination: "https://portal.thirdweb.com/connect",

      source: "/:network/rpc-edge",
      destination:
        "https://portal.thirdweb.com/infrastructure/rpc-edge/overview",

      source: "/:network/engine",
      destination: "https://portal.thirdweb.com/engine",

      source: "/:network/account-abstraction",
      destination:
        "https://portal.thirdweb.com/connect/account-abstraction/overview",

      source: "/:network/pay",
      destination: "https://portal.thirdweb.com/connect/pay/overview",

---

## PR-Codex overview
This PR adds multiple redirect rules in the `redirects.js` file for different network paths to redirect to specific URLs on a third-party portal.

### Detailed summary
- Added redirect rules for different network paths to third-party portal URLs
- Added permanent flag to ensure permanent redirections


apps/dashboard/redirects.js
```diff
@@ -1,6 +1,33 @@
 /** @type {import('next').NextConfig['redirects']} */
 function redirects() {
   return [
+    {
+      source: "/:network/connect-sdk",
+      destination: "https://portal.thirdweb.com/connect",
+      permanent: true,
```
jnsdls, line 7: we never use permanent redirects because if we change our mind users will be stuck with these

jnsdls (changes requested):
why are these necessary?

apps/dashboard/redirects.js
Joe-Thirdweb: These changes are for when a user clicks on the icon of a product on the chainlist page. Such as Connect, Engine, Pay, etc. Currently they will 404. 

Screenshot 2024-09-16 at 10 08 08 AM



For permanent: 
Won't permanent tell crawlers and search engines to use the new destination link? If we ever have to update these again with permanent = true, these would be overridden as well. 

Happy to change if not the case.

jnsdls: if that's what they are for lets instead simply change the links on those, redirects are for when we can't solve it otherwise

jnsdls (changes requested):

jnsdls:
fixed in: https://github.com/thirdweb-dev/js/pull/4638

PR #4644, opened 2024-09-17
Handle custom id for published contract

kumaryash90:
## Problem solved

---

## PR-Codex overview
This PR focuses on handling custom IDs for published contracts. 

### Detailed summary
- Added handling for custom contract IDs
- Updated `deploy-metadata.ts` to include custom contract name

### DIFF
--- a/packages/thirdweb/src/utils/any-evm/deploy-metadata.ts
+++ b/packages/thirdweb/src/utils/any-evm/deploy-metadata.ts
@@ -50,6 +50,7 @@ export async function fetchDeployMetadata(
     ...parsedMeta,
     version: rawMeta.version,
     bytecode: deployBytecode,
+    name: rawMeta.name,
   };
 }
 
--- a/.changeset/clever-moles-do.md
+++ b/.changeset/clever-moles-do.md
@@ -1,0 +1,5 @@
+---
+"thirdweb": patch
+---
+
+Handle custom id for published contract


packages/thirdweb/src/utils/any-evm/deploy-metadata.ts
```diff
@@ -50,6 +50,7 @@ export async function fetchDeployMetadata(
     ...parsedMeta,
     version: rawMeta.version,
     bytecode: deployBytecode,
+    name: rawMeta.name,
```
joaquim-verges, line 53: can you explain this fix ?

kumaryash90: Both `rawMeta` and `parsedMeta` have the field `name`.
But rawMeta.name is actually the contract-id used for publishing, whereas parsedMeta.name is the target contract name.

For handling custom id, we need rawMeta.name, but currently it gets replaced with parsedMeta.name due to the order we deconstruct these objects.

So, either we reverse the order, or reassign name as `rawMeta.name`

Ref from legacy package: https://github.com/thirdweb-dev/js/pull/3424/files#diff-a8673381af2cf9fa6622981b06dd878c1e9e76b6f8d7af5017b2eec29ac4933f

joaquim-verges: makes sense

joaquim-verges (approved):

PR #4849, opened 2024-09-30
[SDK] Update docs for claimTo extensions

kien-ngo:
## Problem solved

---

## PR-Codex overview
This PR enhances the documentation for the `claimTo` function across three extensions: `ERC1155`, `ERC20`, and `ERC721`. It adds examples for basic usage and includes specific guidance for handling drops with allowlists.

### Detailed summary
- Added a "Basic usage" section with example code for `claimTo` in `ERC1155`, `ERC20`, and `ERC721`.
- Included a section for drops with allowlists, explaining the need for the `from` parameter.
- Provided example code for the `claimTo` function with the `from` parameter in all three extensions.

### DIFF
--- a/packages/thirdweb/src/extensions/erc20/drops/write/claimTo.ts
+++ b/packages/thirdweb/src/extensions/erc20/drops/write/claimTo.ts
@@ -20,6 +20,8 @@ export type ClaimToParams = {
  * @param options - The options for the transaction
  * @extension ERC20
  * @example
+ *
+ * ### Basic usage
  * ```ts
  * import { claimTo } from "thirdweb/extensions/erc20";
  * import { sendTransaction } from "thirdweb";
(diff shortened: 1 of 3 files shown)


packages/thirdweb/src/extensions/erc20/drops/write/claimTo.ts
```diff
@@ -34,6 +34,7 @@ export type ClaimToParams = {
  * ```
  * @throws If no claim condition is set
  * @returns A promise that resolves with the submitted transaction hash.
+ * @important If the claiming address is not the same as the receiving address, you need to specify the claimer address as the `from` param to avoid any issue with the allowlist (if one exists)
```
gregfromstl, line 37: I think we parse @note for the docs but not @important

gregfromstl (changes requested):

packages/thirdweb/src/extensions/erc20/drops/write/claimTo.ts
kien-ngo: oh. thanks im gonna check the build for potal when I wake up

kien-ngo: We dont also render `@note`

kien-ngo: updated

jnsdls (approved):

PR #4869, opened 2024-09-30
Check zksync stack

kumaryash90:
## Problem solved

---

## PR-Codex overview
This PR focuses on updating the `isZkSyncChain` function to be asynchronous and modifying several parts of the codebase to await its result. This change enhances the handling of ZkSync chains during contract deployments and transaction requirements.

### Detailed summary
- Changed `isZkSyncChain` to an `async` function.
- Updated calls to `isZkSyncChain` to use `await`.
- Modified contract deployment logic to handle ZkSync chains asynchronously.
- Enhanced error handling in the new `getChainStack` function for fetching chain stack data.

### DIFF
--- a/packages/thirdweb/src/extensions/prebuilts/get-required-transactions.ts
+++ b/packages/thirdweb/src/extensions/prebuilts/get-required-transactions.ts
@@ -43,6 +43,8 @@ export async function getRequiredTransactions(
     modules = [],
   } = options;
 
+  const isZkSync = await isZkSyncChain(chain);
+
   if (deployMetadata?.deployType === "autoFactory") {
     const results: (
       | DeployTransactionResult
@@ -53,7 +55,7 @@ export async function getRequiredTransactions(
         chain,
         client,
       }).then((c) =>
-        c || isZkSyncChain(chain)
+        c || isZkSync
           ? null
           : ({ type: "infra", contractId: "Create2Factory" } as const),
       ),
@@ -62,7 +64,7 @@ export async function getRequiredTransactions(
         client,
         contractId: "Forwarder",
       }).then((c) =>
-        c || isZkSyncChain(chain)
+        c || isZkSync
           ? null
           : ({ type: "infra", contractId: "Forwarder" } as const),
       ),
@@ -78,7 +80,7 @@ export async function getRequiredTransactions(
           }),
         },
       }).then((c) =>
-        c || isZkSyncChain(chain)
+        c || isZkSync
           ? null
           : ({ type: "infra", contractId: "TWCloneFactory" } as const),
       ),
--- a/packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
+++ b/packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
@@ -1,12 +1,44 @@
 import type { Chain } from "../../../chains/types.js";
+import { withCache } from "../../promise/withCache.js";
 
-export function isZkSyncChain(chain: Chain) {
-  return (
-    chain.id === 324 ||
-    chain.id === 300 ||
-    chain.id === 302 ||
-    chain.id === 11124 ||
-    chain.id === 282 || // cronos zkevm testnet
-    chain.id === 388 // cronos zkevm mainnet
+export async function isZkSyncChain(chain: Chain) {
+  if (chain.id === 1337 || chain.id === 31337) {
+    return false;
+  }
+
+  const stack = await getChainStack(chain.id).catch(() => {
+    // fall back to checking against these zksync chain-ids
+    if (
+      chain.id === 324 ||
+      chain.id === 300 ||
+      chain.id === 302 ||
+      chain.id === 11124 ||
+      chain.id === 282 || // cronos zkevm testnet
+      chain.id === 388 // cronos zkevm mainnet
+    ) {
+      return "zksync-stack";
+    }
+
+    return "";
+  });
+
+  return stack === "zksync-stack";
+}
+
+async function getChainStack(chainId: number): Promise<string> {
+  return withCache(
+    async () => {
+      const res = await fetch(`https://${chainId}.rpc.thirdweb.com/stack`);
+
+      if (!res.ok) {
+        res.body?.cancel();
+        throw new Error(`Error fetching stack for ${chainId}`);
+      }
+
+      const data = await res.json();
+
+      return data.stack;
+    },
+    { cacheKey: `stack:${chainId}`, cacheTime: 24 * 60 * 60 * 1000 },
   );
 }
--- a/apps/dashboard/src/components/contract-components/contract-deploy-form/custom-contract.tsx
+++ b/apps/dashboard/src/components/contract-components/contract-deploy-form/custom-contract.tsx
@@ -386,7 +386,9 @@ export const CustomContractForm: React.FC<CustomContractFormProps> = ({
         throw new Error("no chain");
       }
 
-      const compilerType = isZkSyncChain(walletChain) ? "zksolc" : "solc";
+      const compilerType = (await isZkSyncChain(walletChain))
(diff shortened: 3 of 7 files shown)


packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
```diff
@@ -1,12 +1,27 @@
...
+  const data = await res.json();
+
+  if (res.ok) {
+    const stack = data.stack;
+    chainStackCache.set(chainId, { stack });
+
+    return stack;
+  } else {
+    throw new Error(`Error: ${res.status} - ${data}`);
+  }
```
jnsdls, line 26: we have a way to cache a promise: `withCache` use it for this

jnsdls (changes requested):

packages/thirdweb/src/extensions/prebuilts/get-required-transactions.ts
```diff
@@ -52,17 +52,17 @@ export async function getRequiredTransactions(
       getDeployedCreate2Factory({
         chain,
         client,
-      }).then((c) =>
-        c || isZkSyncChain(chain)
+      }).then(async (c) =>
+        c || (await isZkSyncChain(chain))
```
joaquim-verges, line 56: you should do this only once outside of the promise all

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
```diff
@@ -1,12 +1,27 @@
...
+
+export async function isZkSyncChain(chain: Chain) {
+  const stack = await getChainStack(chain.id);
+
+  return stack === "zksync-stack";
+}
+
+export async function getChainStack(chainId: number): Promise<string> {
+  if (chainStackCache.has(chainId)) {
+    return chainStackCache.get(chainId);
```
joaquim-verges, line 13: we have a utility for this that handles caching much better (including caching promises)

`return withCache(async () => { ... })`

check out how we do it in `fetchPublishedContractMetadata()`

you probably want a 1 day cache on it, just in case

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
kumaryash90: updated

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
```diff
@@ -1,12 +1,24 @@
...
-    chain.id === 11124 ||
-    chain.id === 282 || // cronos zkevm testnet
-    chain.id === 388 // cronos zkevm mainnet
+export async function isZkSyncChain(chain: Chain) {
+  const stack = await getChainStack(chain.id);
+
+  return stack === "zksync-stack";
+}
+
+export async function getChainStack(chainId: number): Promise<string> {
```
joaquim-verges, line 10: This probably doesn't need to be exported, don't think you're using it anywhere

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
```diff
@@ -1,12 +1,24 @@
...
+  return withCache(
+    async () => {
+      const res = await fetch(`https://${chainId}.rpc.thirdweb.com/stack`);
+      const data = await res.json();
+
+      if (res.ok) {
+        return data.stack;
+      } else {
+        return null;
```
joaquim-verges, line 19: I can't remember the behavior of with cache, cam you double check that it won't cache null as a result?

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
jnsdls: I can do u one better, lint is not allowing you to export it SINCE you are not using it anywhere

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
jnsdls: if you don't want cache simply throw on `!res.ok`

joaquim-verges: yeah that's what i thought

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
kumaryash90: fixed

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
```diff
@@ -1,12 +1,24 @@
...
+  return withCache(
+    async () => {
+      const res = await fetch(`https://${chainId}.rpc.thirdweb.com/stack`);
+      const data = await res.json();
+
+      if (res.ok) {
+        return data.stack;
+      }
+
+      return null;
```
joaquim-verges, line 20: Never return null, just return the .json() and let it throw here

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
```diff
@@ -1,12 +1,24 @@
...
-export function isZkSyncChain(chain: Chain) {
-  return (
-    chain.id === 324 ||
-    chain.id === 300 ||
-    chain.id === 302 ||
-    chain.id === 11124 ||
-    chain.id === 282 || // cronos zkevm testnet
-    chain.id === 388 // cronos zkevm mainnet
+export async function isZkSyncChain(chain: Chain) {
+  const stack = await getChainStack(chain.id);
```
joaquim-verges, line 5: In here, add a .catch() that defaults to checking the old hardcoded list behavior, that way we have a fallback if things go wrong

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
```diff
@@ -1,12 +1,38 @@
...
+  return stack === "zksync-stack";
+}
+
+async function getChainStack(chainId: number): Promise<string> {
+  return withCache(
+    async () => {
+      const res = await fetch(`https://${chainId}.rpc.thirdweb.com/stack`);
+      const data = await res.json();
+
+      if (!res.ok) {
```
joaquim-verges, line 30: Would even remove this. Just let it throw without any checks

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
```diff
@@ -1,12 +1,38 @@
...
-export function isZkSyncChain(chain: Chain) {
-  return (
-    chain.id === 324 ||
-    chain.id === 300 ||
-    chain.id === 302 ||
-    chain.id === 11124 ||
-    chain.id === 282 || // cronos zkevm testnet
-    chain.id === 388 // cronos zkevm mainnet
+export async function isZkSyncChain(chain: Chain) {
+  const stack = await getChainStack(chain.id).catch(() => {
```
joaquim-verges, line 5: Add localhost chain Id check above this

packages/thirdweb/src/utils/any-evm/zksync/isZkSyncChain.ts
jnsdls: eh, even errors CAN parse valid JSON meaning it would trust the error

it's probably good to check this BUT:
1. it should happen before we do `res.json()`
2. it should also call `res.body?.cancel()` so that it does not leak memory

jnsdls (changes requested):

jnsdls (approved):

PR #4956, opened 2024-10-08
feat: Support multiple wallets in Engine dashboard

arcoraven:
- Adds support for configuring creds for AWS + GCP independently
- Moves all touched UI to tailwind
- Import / Create wallet buttons allow selecting the wallet type
- Unrelated: Moved version upgrade button to the new enpdoint. Will deprecate the old one.

---

## PR-Codex overview
This PR focuses on enhancing the handling of backend wallet configurations in the `engine` module. It introduces new components and updates existing ones to improve the integration of local, AWS KMS, and GCP KMS wallets, along with better feature support checks.

### Detailed summary
- Removed `instanceUrl` prop in favor of `instance` for several components.
- Introduced `LocalConfig` component for local wallet setup.
- Updated `EngineWalletConfig` to manage wallet types more efficiently.
- Added support for `EngineBackendWalletType` and related options.
- Enhanced form handling for wallet configurations using React Hook Form.
- Improved validation and user feedback for wallet import and creation.
- Simplified the usage of wallet configuration hooks across components.


apps/dashboard/src/@3rdweb-sdk/react/hooks/useEngine.ts
```diff
@@ -1639,6 +1639,9 @@ export function useEngineCreateNotificationChannel(engineId: string) {
         `${THIRDWEB_API_HOST}/v1/engine/${engineId}/notification-channels`,
         {
           method: "POST",
+          headers: {
```
arcoraven, line 1642: This was missing but it was working for some reason?

apps/dashboard/src/@3rdweb-sdk/react/hooks/useEngine.ts
```diff
@@ -138,6 +145,15 @@ export function useEngineSystemHealth(
   });
 }
 
+// Helper function to check if a feature is supported.
+export function useHasEngineFeature(
+  instanceUrl: string,
+  feature: EngineFeature,
+) {
+  const { data } = useEngineSystemHealth(instanceUrl);
+  return !!data?.features?.includes(feature);
```
MananTank, line 154: We should also return a `isPending` state here so that we can use it to show loading indicator in UI

jnsdls: we should generally expose the entire useQuery return value, we should not prescribe how a consumer consumes the custom hook, that said in this case we want a specific behavior so I think it's OK to override the "data" to the boolean flag we want here, but as @MananTank mentioned we should also expose the state of the query at the least

apps/dashboard/src/components/engine/overview/import-backend-wallet-button.tsx
```diff
@@ -64,7 +64,20 @@ export const ImportBackendWalletButton: React.FC<
   );
   const form = useForm<ImportBackendWalletInput>();
 
-  const onSubmit = (data: ImportBackendWalletInput) => {
+  const onSubmit = (raw: ImportBackendWalletInput) => {
+    // Submit only relevant fields.
+    let data: ImportBackendWalletInput = {};
```
MananTank, line 69: Missing the `label` key

apps/dashboard/src/@3rdweb-sdk/react/hooks/useEngine.ts
jnsdls: if the API u call is happy to respond with JSON even when not explicitly requested then it'll work

apps/dashboard/src/@3rdweb-sdk/react/hooks/useEngine.ts
```diff
@@ -401,29 +415,22 @@ export function useEngineTransactions(instance: string, autoUpdate: boolean) {
...
-    }
-  | {
-      type: "gcp-kms";
-      gcpApplicationProjectId: string;
-      gcpKmsLocationId: string;
-      gcpKmsKeyRingId: string;
-      gcpApplicationCredentialEmail: string;
-      gcpApplicationCredentialPrivateKey: string;
-    };
+export interface WalletConfigResponse {
```
MananTank, line 418: Can we bring back union type here so it's clear which combinations are possible/not possible

apps/dashboard/src/components/engine/configuration/engine-wallet-config.tsx
```diff
@@ -1,102 +1,80 @@
...
+          name,
+          isActive: activeTab === key,
+          isEnabled: true,
+          onClick: () => setActiveTab(key),
+          icon:
+            (key === "aws-kms" && !isAwsKmsConfigured) ||
+            (key === "gcp-kms" && !isGcpKmsConfigured)
+              ? ({ className }) => (
+                  <ToolTipLabel label="Not configured">
+                    <CircleAlertIcon className={className} />
```
MananTank, line 69: This icon is too large for alert icon, set `size-4` instead
 
```ts
<CircleAlertIcon className={cn(className, 'size-4')} />
```

apps/dashboard/src/components/engine/configuration/kms-aws-config.tsx
```diff
@@ -1,126 +1,173 @@
...
+        </div>
+
+        <div className="flex items-center justify-end gap-4">
+          {!supportsMultipleWalletTypes && (
+            <p className="text-destructive-text text-sm">
+              This will clear other credentials.
+            </p>
+          )}
+          <Button disabled={!form.formState.isDirty} type="submit">
+            {form.formState.isSubmitting ? "Saving..." : "Save"}
```
MananTank, line 167: Render a `<Spinner className='size-4' />` on left side of Saving. ( You'll also have to add `gap-2` className on `Button`

apps/dashboard/src/components/engine/configuration/kms-gcp-config.tsx
```diff
@@ -1,165 +1,220 @@
...
-            w={{ base: "full", md: "49%" }}
-            placeholder="Your Google App Private Key"
+          </FormFieldSetup>
+        </div>
+
+        <FormFieldSetup
+          label="Private Key"
+          errorMessage={
+            form.getFieldState(
+              "gcpApplicationCredentialPrivateKey",
```
MananTank, line 186: none of these form fields are required in this form?

apps/dashboard/src/components/engine/configuration/kms-gcp-config.tsx
```diff
@@ -1,165 +1,220 @@
...
+        </FormFieldSetup>
+
+        <div className="flex items-center justify-end gap-4">
+          {!supportsMultipleWalletTypes && (
+            <p className="text-destructive-text text-sm">
+              This will clear other credentials.
+            </p>
+          )}
+          <Button disabled={!form.formState.isDirty} type="submit">
+            {form.formState.isSubmitting ? "Saving..." : "Save"}
```
MananTank, line 214: same here - setGcpKmsConfig is not awaited in onSubmit - you won't get correct `isSubmitting` state - prefer using `mutation.isPending` directly or add await

apps/dashboard/src/components/engine/overview/create-backend-wallet-button.tsx
```diff
@@ -60,67 +86,166 @@ export function CreateBackendWalletButton(props: {
...
-              <div className="flex flex-col gap-4">
-                <div>
-                  <p className="mb-1 text-sm">Wallet Type</p>
-                  <Badge className="text-sm" variant="outline">
-                    {walletType}
-                  </Badge>
-                </div>
+                <div className="flex flex-col gap-5">
+                  {/* Wallet type */}
+                  {/* <FormField
```
MananTank, line 128: If not required - let's just delete this

apps/dashboard/src/components/engine/overview/import-backend-wallet-button.tsx
```diff
@@ -10,204 +9,327 @@ import {
...
-                </DialogTitle>
-              </DialogHeader>
+                <div className="flex flex-col gap-5">
+                  {/* Wallet type */}
+                  <FormItem>
+                    <FormLabel>Wallet Type</FormLabel>
+                    <FormControl>
+                      <Select
+                        onValueChange={(value) =>
+                          setWalletType(value as EngineBackendWalletType)
```
MananTank, line 157: reset all the other keys here. 
It would be confusing to keep the data of previous wallet type ( like label )

apps/dashboard/src/components/engine/overview/import-backend-wallet-button.tsx
```diff
@@ -10,204 +9,327 @@ import {
...
+                        <FormFieldSetup
+                          isRequired
+                          label="AWS KMS ARN"
+                          errorMessage={
+                            form.getFieldState("awsKmsArn", form.formState)
+                              .error?.message
+                          }
+                          htmlFor="awsKmsArn"
+                          tooltip={null}
+                        >
```
MananTank, line 252: add `key` to all the `FormFieldSetup` components here to avoid having wrong values in the inputs

MananTank (commented):
Please address the comments

apps/dashboard/src/@3rdweb-sdk/react/hooks/useEngine.ts
arcoraven: Hmm the endpoint now doesn't accept a union type which is why I matched it on the dashboard side.

apps/dashboard/src/@3rdweb-sdk/react/hooks/useEngine.ts
arcoraven: I guess this was meant to be a thin wrapper/helper around an existing query. But I can expose it if that helps.

apps/dashboard/src/components/engine/overview/import-backend-wallet-button.tsx
arcoraven: Good catch, forgot this.

apps/dashboard/src/components/engine/configuration/engine-wallet-config.tsx
arcoraven: I don't think the tab content needs a warning. Once they're in the content they should know they're adding/updating config.

apps/dashboard/src/components/engine/configuration/kms-aws-config.tsx
arcoraven: Makes sense, this was old code from the previous form behavior.

arcoraven:
Thanks @MananTank 

I tested these flows manually:

---- Engine >= v2.0.22

- ✅ add local wallet
- ✅ add aws wallet
- ✅ import local wallet
- ✅ import aws wallet
- ✅ send transfer on local wallet
- ✅ send transfer on aws wallet
- ❌ delete aws creds
  - This is currently blocked by a backend change but IMO it shouldn't block this PR

---- Engine < v2.0.22

- ✅ add local wallet
- ✅ add aws wallet
- ✅ import local wallet
- ✅ import aws wallet
- ✅ send transfer on local wallet
- ✅ send transfer on aws wallet
- ✅ delete aws creds

MananTank (approved):

PR #5086, opened 2024-10-19
[SDK] Feature: Track transactions and SIWE

gregfromstl:
## Problem solved

---

## PR-Codex overview
This PR focuses on enhancing the tracking functionality in the `thirdweb` SDK by refining the analytics tracking methods for wallet connections and transactions, ensuring better data collection and reporting.

### Detailed summary
- Updated imports for `trackConnect` and `trackPayEvent` to specific paths.
- Introduced `trackConnect` and `trackPayEvent` functions with detailed parameters.
- Added tracking calls in wallet connection and transaction methods.
- Implemented tests for `trackConnect`, `track`, `trackPayEvent`, and `trackLogin`.
- Enhanced error handling and response structures in tracking functions.


packages/thirdweb/src/analytics/track/pay.test.ts
```diff
@@ -0,0 +1,98 @@
...
+import { trackPayEvent } from "./pay.js";
+import type { ThirdwebClient } from "../../client/client.js";
+
+const server = setupServer(
+  http.post("https://c.thirdweb.com/event", () => {
+    return HttpResponse.json({});
+  }),
+);
+
+describe("trackPayEvent", () => {
```
joaquim-verges, line 13: 👍

packages/thirdweb/src/analytics/track/siwe.ts
```diff
@@ -0,0 +1,69 @@
...
+    type: "login:error",
+  });
+}
+
+/**
+ * @internal
+ */
+async function trackSiweEvent(
+  event: SiweEvent & {
+    type: "login:success" | "login:error";
```
joaquim-verges, line 54: Nit: would just call this action so it matches the schema. Wasn't sure what it mapped to when reading the call site

joaquim-verges: Also not sure we need two login actions, I think it's simpler and more efficient to query all 'login' action, and then split by errorCode. 

We should put a default errorCode for error events like "unknown error" in case the err msg is empty

packages/thirdweb/src/analytics/track/transaction.ts
```diff
@@ -0,0 +1,78 @@
...
+  error?: undefined;
+};
+
+/**
+ * @internal
+ */
+export async function trackTransaction(args: TransactionSuccessEvent) {
+  return trackTransactionEvent({
+    ...args,
+    event: "transaction:success",
```
joaquim-verges, line 33: Same reasoning here

packages/thirdweb/src/wallets/coinbase/coinbaseWebSDK.ts
```diff
@@ -290,12 +302,43 @@
...
-            from: this.address,
+            from: getAddress(address),
             to: tx.to as Address,
             data: tx.data,
           },
         ],
       })) as Hex;
+
+      // Non-blocking wait for receipt to track transaction
+      waitForReceipt({
```
joaquim-verges, line 313: Oof, I don't think we can do this. Feels very heavy to always poll just for a tracking event.

Can't we just track on submit on chain? We have the tx hash here, why fo we need to wait for confirmation?

packages/thirdweb/src/wallets/in-app/core/wallet/enclave-wallet.ts
```diff
@@ -190,13 +196,44 @@ export class EnclaveWallet implements IWebWallet {
...
         const signedTx = await _signTransaction(tx);
+
         const transactionHash = await eth_sendRawTransaction(
           rpcRequest,
           signedTx,
         );
-        return {
+
+        // Non-blocking wait for receipt to track transaction
+        waitForReceipt({
```
joaquim-verges, line 206: I think I would just track an event like 'transaction_sent' inside sendTransaction, will cover all wallets in one central point.

And we can also track a separate action like 'transaction_confirmed' inside 'waitForReceipt' or 'getTransactionReceipt'.

These two will share the same transaction hash so we can show interesting data for tx sent vs confirmed easily

joaquim-verges (changes requested):
Would like to change the tx tracking approach

packages/thirdweb/src/wallets/coinbase/coinbaseWebSDK.ts
gregfromstl: Yeah that's fair, we just can't determine if it reverted or not then

joaquim-verges: i think that's ok - we can add a separate event on transaction confirmation and correlate based on txhash. if the app cares about confirmations, then they'll wait for receipt.

packages/thirdweb/src/wallets/coinbase/coinbaseWebSDK.ts
```diff
@@ -290,12 +298,22 @@
...
         ],
       })) as Hex;
+
+      trackTransaction({
+        client: client,
+        walletAddress: getAddress(address),
+        walletType: COINBASE,
+        transactionHash,
+        contractAddress: tx.to ?? undefined,
+        gasPrice: tx.gasPrice,
```
joaquim-verges, line 314: wonder if we should track the function selector here, we could potentially resolve it to the function name too

joaquim-verges: actually if we move this tracking to `sendTransaction` we would track for all existing and future wallets, and also have things like the functionName. Though we would loose the `walletType` prop.

I wonder if that's worth it though?

packages/thirdweb/src/wallets/in-app/web/lib/iframe-wallet.ts
```diff
@@ -242,13 +243,22 @@ export class IFrameWallet implements IWebWallet {
...
         const transactionHash = await eth_sendRawTransaction(
           rpcRequest,
           signedTx,
         );
-        return {
+
+        trackTransaction({
+          client,
+          walletAddress: address,
+          walletType: "inApp",
```
joaquim-verges, line 255: this could be ecosystem too no?

packages/thirdweb/src/wallets/in-app/core/wallet/enclave-wallet.ts
```diff
@@ -190,13 +192,23 @@ export class EnclaveWallet implements IWebWallet {
...
           rpcRequest,
           signedTx,
         );
-        return {
+
+        // Non-blocking wait for receipt to track transaction
+        trackTransaction({
+          client,
+          walletAddress: address,
+          walletType: "inApp",
```
joaquim-verges, line 205: this could be ecosystem too?

packages/thirdweb/src/wallets/in-app/web/lib/iframe-wallet.ts
gregfromstl: It can't since ecosystem wallets are all on enclave now

packages/thirdweb/src/wallets/coinbase/coinbaseWebSDK.ts
gregfromstl: Yeah,I was thinking we could use the indexer with the transaction hash to get the function information later

joaquim-verges: Yeah. Ok let's try that for now

packages/thirdweb/src/wallets/in-app/web/lib/iframe-wallet.ts
joaquim-verges: Right

packages/thirdweb/src/analytics/track/siwe.ts
```diff
@@ -0,0 +1,48 @@
...
+    action: "login:attempt",
+  });
+}
+
+/**
+ * @internal
+ */
+async function trackSiweEvent(
+  event: SiweEvent & {
+    action: "login:attempt";
```
joaquim-verges, line 33: i read this as 'the event before verifying any payload' - but really its the login event, and that can be success of failed if theres an error right?

gregfromstl: I thought we wanted a single action label error and we'd tell from the error if it succeeded or not?

packages/thirdweb/src/wallets/wallet-connect/controller.ts
```diff
@@ -330,13 +338,22 @@
...
             to: tx.to as Address,
             data: tx.data,
           },
         ],
       })) as Hex;
 
+      trackTransaction({
+        client: client,
+        walletAddress: getAddress(address),
+        walletType: "walletConnect",
```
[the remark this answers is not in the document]
gregfromstl: I don't think so, `walletConnect` is the wallet ID here and how we're already recording the connection events

packages/thirdweb/src/analytics/track/siwe.ts
joaquim-verges: yeah, was just thrown off by the 'attempt' label.

would have just called it `action: login` or `action: authenticate`

joaquim-verges (approved):

gregfromstl:
### Merge activity

* **Oct 22, 8:26 PM EDT**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=thirdweb-dev&repo=js) once it meets the requirements.

PR #5216, opened 2024-10-29
Check for linear in PR description

edwardysun:
## Problem solved

Fixes https://linear.app/thirdweb/issue/ENG-2/better-enforce-linear-use

---

## PR-Codex overview
This PR introduces a GitHub Actions workflow that checks for specific keywords in the pull request body to ensure linked issues are mentioned.

### Detailed summary
- Added a new workflow file: `.github/workflows/linear.yml`.
- Defined a workflow named `Check for "linear" in pull request body`.
- Triggered on pull request events (opened and edited).
- Created a job `check_linear_in_pr_body` to verify the presence of keywords "CNCT" or "DASH" in the PR body.

### DIFF
--- a/.github/workflows/linear.yml
+++ b/.github/workflows/linear.yml
@@ -1,0 +1,26 @@
+name: Check for "linear" in pull request body
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+
+jobs:
+  check_linear_in_pr_body:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for linked issue in pull request body
+        run: |
+          pr_body=$(curl -s \
+            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.number }} | jq -r .body)
+
+          echo "Pull Request Body: $pr_body"
+
+          if echo "$pr_body" | grep -iE "CNCT|DASH"; then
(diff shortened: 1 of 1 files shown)


.github/workflows/linear.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Check for "linear" in the pull request body
+        run: |
+          pr_body=$(curl -s \
+            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.number }} | jq -r .body)
+
+          echo "Pull Request Body: $pr_body"
+
+          if echo "$pr_body" | grep -iq "linear"; then
```
joaquim-verges, line 21: You should not need to paste the entire link. Linear can find the GH PR just by pasting the ticket id in the body.

There must be a gh action that does this already properly

jnsdls: yeah this is not the way to do it, also we want external contributions and this would fail those?

edwardysun: I can use an existing github action instead

Re: external contributions - I thought about this and yes, it will fail, but this is an optional rule that serves more as a reminder. How much would we want to optimize for it? I could have a list of known github usernames of employees if we'd like. I just want to get something out first

gregfromstl (approved):

jnsdls (approved):

PR #5413, opened 2024-11-13
[Dashboard] Feature: Updated Claimable

GWSzeto:
- tokenId and setToken introduced at the parent level ClaimableModule
- Claim Conditions are fetched at the parent level

https://linear.app/thirdweb/issue/DES-251/claimable-erc721-erc1155-erc20

---

## PR-Codex overview
This PR primarily modifies the `Claimable` module and related components to enhance functionality and improve user experience by adding token ID management and updating claim conditions.

### Detailed summary
- Changed condition for rendering `ClaimableModule` to check for "Claimable" instead of "Claimable-ignore".
- Updated logic in `CurrencySelector` for checking `initialValue`.
- Added `tokenId` and `noClaimConditionSet` state management in `claimable.stories.tsx`.
- Removed unused `getClaimConditionErc1155Stub` function.
- Introduced a checkbox for `noClaimConditionSet` in the UI.
- Adjusted claim condition logic to include `tokenId` validation.
- Enhanced `ClaimableModule` with new state and props for token ID management.
- Updated `ClaimableModuleUI` to handle token ID input and validation.
- Refactored `ClaimConditionSection` to check for claim conditions and manage form state.
- Added warning alert for cases with no claim conditions set.

### DIFF
--- a/apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx
+++ b/apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx
@@ -1,4 +1,5 @@
 "use client";
+
 import { FormFieldSetup } from "@/components/blocks/FormFieldSetup";
 import { DatePickerWithRange } from "@/components/ui/DatePickerWithRange";
 import {
@@ -7,7 +8,7 @@ import {
   AccordionItem,
   AccordionTrigger,
 } from "@/components/ui/accordion";
-import { Alert, AlertTitle } from "@/components/ui/alert";
+import { Alert, AlertDescription, AlertTitle } from "@/components/ui/alert";
 import { Button } from "@/components/ui/button";
 import {
   Form,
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx
```diff
@@ -57,20 +64,10 @@ export type ClaimConditionValue = {
...
-  endTimestamp: number;
-  auxData: string;
-};
-
 function ClaimableModule(props: ModuleInstanceProps) {
   const { contract, ownerAccount } = props;
   const account = useActiveAccount();
+  const [tokenId, setTokenId] = useState<string>("");
 
   const isErc721 = props.contractInfo.name === "ClaimableERC721";
```
Comment (MananTank) [inline apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx:72]: Price per token is either not showing the correct data or it is not getting updated 

this is what I submited: 



this is what I get after page refetch:

apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx
```diff
@@ -57,20 +64,10 @@ export type ClaimConditionValue = {
   auxData: string;
 };
```
Comment (MananTank) [inline apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx:66]: Need to show a message when no claim condition is set for a token id

apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx
```diff
@@ -82,29 +79,25 @@ function ClaimableModule(props: ModuleInstanceProps) {
   );
 
   const claimConditionQuery = useReadContract(
```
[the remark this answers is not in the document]
Comment (GWSzeto) [reply apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx:81]: Updating it so that if no currency is set, then it will default to the native currency.

Reason being is that in `setClaimCondition.ts`, if the currency is undefined, then it defaults to the native token address

apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx
Comment (GWSzeto) [reply apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx:67]: addressed

apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx
Comment (GWSzeto) [reply apps/dashboard/src/app/(dashboard)/(chain)/[chain_id]/[contractAddress]/modules/components/Claimable.tsx:75]: Addressed with the following code:

```TS
      // default case is zero state, so 0 // 10 ** 18 still results in 0
      pricePerToken: Number(
        toTokens(claimCondition?.pricePerUnit, props.tokenDecimals || 18),
      ),
```

This covers the default case (as mentioned in the comments) since the `pricePerUnit` will always be 0, dividing it by anything will result in zero

MananTank:
The messaging in "No Claim Condition Set" alert is a bit confusing - there is no "Set claim condition" button. We should not show the form at all when there is no claim condition set. It should be just Alert + Add claim condition button and we show the form and remove the alert when you click on "Add claim condition" button

image

MananTank (approved):

PR #5673, opened 2024-12-10
fix: should always contain gas price info even if it's 0

ElasticBottle:
https://linear.app/thirdweb/issue/CNCT-2602/issue-signing-deploy-tx-on-saakura

---

## PR-Codex overview
This PR focuses on ensuring that gas price information is always included for enclave wallets, even when it is set to 0. It modifies the condition for checking `maxFeePerGas` to ensure proper handling of gas fees.

### Detailed summary
- Updated the condition for `maxFeePerGas` to check if it is of type `bigint`.
- Ensured that gas price information is always included for enclave wallets, even if it is 0.

### DIFF
--- a/packages/thirdweb/src/wallets/in-app/core/wallet/enclave-wallet.ts
+++ b/packages/thirdweb/src/wallets/in-app/core/wallet/enclave-wallet.ts
@@ -162,7 +162,7 @@ export class EnclaveWallet implements IWebWallet {
         chainId: toHex(tx.chainId),
       };
 
-      if (tx.maxFeePerGas) {
+      if (typeof tx.maxFeePerGas === "bigint") {
         transaction.maxFeePerGas = toHex(tx.maxFeePerGas);
         transaction.maxPriorityFeePerGas =
           typeof tx.maxPriorityFeePerGas === "bigint"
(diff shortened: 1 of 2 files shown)


packages/thirdweb/src/wallets/in-app/core/wallet/enclave-wallet.ts
```diff
@@ -167,11 +167,11 @@ export class EnclaveWallet implements IWebWallet {
         transaction.maxPriorityFeePerGas =
           typeof tx.maxPriorityFeePerGas === "bigint"
             ? toHex(tx.maxPriorityFeePerGas)
-            : undefined;
+            : toHex("0");
         transaction.type = 2;
       } else {
         transaction.gasPrice =
-          typeof tx.gasPrice === "bigint" ? toHex(tx.gasPrice) : undefined;
+          typeof tx.gasPrice === "bigint" ? toHex(tx.gasPrice) : toHex("0");
```
joaquim-verges, line 174: this is dangerous, a lot of tools like at the presence of gasPrice to know if its a legacy tx

we cant risk sending all values, has to be one or the other

joaquim-verges (approved):

PR #5988, opened 2025-01-20
[SDK] Feature: Makes chain optional on smart wallet config

gregfromstl:
## PR-Codex overview
This PR introduces changes to make the `chain` parameter optional for smart wallet functionalities, enhancing flexibility in smart account integration. It also updates related tests and documentation to reflect these changes.

### Detailed summary
- Updated `ConnectButton` to make `chain` optional in `smartWalletOptions`.
- Modified type definitions for `SmartWalletOptions` and `SmartAccountOptions` to allow optional `chain`.
- Adjusted logic in several functions to handle cases where `chain` is undefined.
- Updated tests to validate behavior with default chain settings.

### DIFF
--- a/packages/thirdweb/src/wallets/smart/index.ts
+++ b/packages/thirdweb/src/wallets/smart/index.ts
@@ -53,6 +53,7 @@ import type {
   SmartWalletConnectionOptions,
   SmartWalletOptions,
   TokenPaymasterConfig,
+  UserOpOptions,
   UserOperationV06,
   UserOperationV07,
 } from "./types.js";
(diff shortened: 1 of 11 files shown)


packages/thirdweb/src/wallets/smart/index.ts
```diff
@@ -92,7 +93,8 @@ export async function connectSmartAccount(
   }
 
   const options = creationOptions;
-  const chain = connectChain ?? options.chain;
+  // Fallback to mainnet if no chain is provided (we only need this for pre-deploy signatures since transactions and deployments must define their own chain)
+  const chain = connectChain ?? options.chain ?? getCachedChain(1);
```
joaquim-verges, line 97: One issue I see is that below this we get the factory and account contracts based on this chain variable.

It works with our default factory because it's deterministic (though need to double check it's on mainnet) but might error if you pass another factory address and no chain.

Also would double check we don't use this chain car for other assumptions in this connect flow

gregfromstl:
Closing as this breaks zk chains

PR #6021, opened 2025-01-22
[service-utils] feat: Add usageV2 support

arcoraven:
https://linear.app/thirdweb/issue/CORE-711/usage-v2-service-utils

---

## PR-Codex overview
This PR introduces support for `usageV2`, adding event handling and Kafka integration to the `service-utils` package, along with updates to the `package.json` and TypeScript definitions.

### Detailed summary
- Added `sendUsageV2Events` function to send events to Kafka.
- Introduced `UsageV2Event` interface for event structure.
- Created `UsageV2Producer` class for Kafka event production.
- Updated `package.json` to include `kafkajs` dependency.
- Minor formatting changes in `package.json` and lock files.


packages/service-utils/package.json
```diff
@@ -46,6 +46,7 @@
   "sideEffects": false,
   "dependencies": {
     "aws4fetch": "1.0.20",
+    "kafkajs": "^2.2.4",
```
jnsdls, line 49: pin the version


```suggestion
    "kafkajs": "2.2.4",
```

packages/service-utils/src/cf-worker/usageV2.ts
```diff
@@ -0,0 +1,45 @@
+import type { UsageV2Event } from "src/core/usageV2.js";
```
jnsdls, line 1: use relative imports

packages/service-utils/src/node/usageV2.ts
```diff
@@ -0,0 +1,110 @@
+import { randomUUID } from "node:crypto";
+import { checkServerIdentity } from "node:tls";
+import { Kafka, type Producer } from "kafkajs";
+import type { UsageV2Event } from "src/core/usageV2.js";
```
jnsdls, line 4: relative import pls

packages/service-utils/src/core/usageV2.ts
```diff
@@ -0,0 +1,56 @@
+export interface UsageV2Event {
+  /**
+   * A unique identifier for the event. Defaults to a random UUID.
+   * Useful if your service retries sending events.
+   */
+  id?: `${string}-${string}-${string}-${string}-${string}`;
```
jnsdls, line 6: might be too strict on the type here, just `string` should be fine - unless you want to specifically enforce uuidv4

packages/service-utils/src/core/usageV2.ts
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * The event timestamp. Defaults to now().
+   */
+  created_at?: Date;
+  /**
+   * The source of the event. Example: "storage"
+   */
+  source: string;
```
jnsdls, line 14: COULD consider enforcing the actual allowed service names here

packages/service-utils/src/node/usageV2.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (!this.producer) {
+      throw new Error("Producer not initialized. Call `init()` first.");
+    }
+
+    const parsedEvents = events.map(({ id, created_at, data, ...rest }) => {
+      return {
+        id: id ?? randomUUID(),
+        created_at: created_at ?? new Date(),
+        data: JSON.stringify(data),
+        ...rest,
```
jnsdls, line 88: might be better to just spell out each field here

jnsdls: for `team_id` field we MAY want to de-normalize or normalize it - since it might be passed in as `team_<id>` or just `<id>` based on how the service consumes it

packages/service-utils/src/node/usageV2.ts
```diff
@@ -0,0 +1,110 @@
...
+      };
+    });
+
+    await this.producer.send({
+      topic: TOPIC_USAGE_V2,
+      messages: parsedEvents.map((event) => ({
+        value: JSON.stringify(event),
+      })),
+    });
```
jnsdls, line 97: what should happen if this were to throw?

jnsdls (changes requested):

packages/service-utils/src/core/usageV2.ts
arcoraven: It's UUID in the ClickHouse schema and I'd prefer event identifiers be consistent for all callers.

packages/service-utils/src/core/usageV2.ts
arcoraven: No strong opinion, given how infrequently this changes the product source be fine to enforce a stricter type.

packages/service-utils/src/node/usageV2.ts
arcoraven: The caller needs to catch and handle throws. I don't want the library to silently ignore any exceptions. For example for AA we may want to retry usage emissions.

packages/service-utils/src/core/usageV2.ts
arcoraven: Actually I like not enforcing a type. It decouples things meaning a new service can start sending a different source right away (e.g. rpcV2) and later create a materialized view to extract/aggregate that data.

packages/service-utils/src/node/usageV2.ts
jnsdls: yeah I would strip it out in this case

packages/service-utils/src/node/usageV2.ts
```diff
@@ -0,0 +1,121 @@
...
+        team_id: event.team_id.startsWith("team_")
+          ? event.team_id.slice(5)
+          : event.team_id,
+        client_id: event.client_id,
+        sdk_name: event.sdk_name,
+        sdk_platform: event.sdk_platform,
+        sdk_version: event.sdk_version,
+        sdk_os: event.sdk_os,
+        product_name: event.product_name,
+        data: JSON.stringify(event.data),
```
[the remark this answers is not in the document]
jnsdls: @arcoraven this is technically correct by the AI however I am unsure what the ingest expects, does it expect an object or a string during ingest for this field? as it is it would receive a string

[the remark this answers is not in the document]
arcoraven: This is intentional _right now_ when the raw_events table expects stringified JSON for `data`. However I'm going to try moving the `data` type to a Clickhouse native JSON type today which accepts a raw object (non stringified).

So before I merge this I'll decide if the native JSON type is better.

packages/service-utils/src/node/usageV2.ts
```diff
@@ -0,0 +1,121 @@
...
+      };
+    });
```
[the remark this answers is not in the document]
arcoraven: Good catch, missed this.

PR #6104, opened 2025-01-29
[TOOL-3260] SDK: Fix Pay Modal logging react-query error when opened

MananTank:
## PR-Codex overview
This PR focuses on fixing the logging of a react-query error in the `thirdweb` Pay Modal when it is opened, ensuring that an error is thrown only when both `account` and `wallet` are not available.

### Detailed summary
- Updated the condition in `TransactionModal` to throw an error if `account` or `wallet` is missing.
- Added a `return null` statement after the error throw.
- Modified the `enabled` property to check both `wallet` and `account`.

### DIFF
--- a/packages/thirdweb/src/react/web/ui/TransactionButton/TransactionModal.tsx
+++ b/packages/thirdweb/src/react/web/ui/TransactionButton/TransactionModal.tsx
@@ -37,7 +37,10 @@ export function TransactionModal(props: ModalProps) {
   useQuery({
     queryKey: ["transaction-modal-event"],
     queryFn: () => {
-      if (!account || !wallet) return;
+      // Query data cannot be undefined, return null
+      if (!account || !wallet) {
+        return null;
+      }
       trackPayEvent({
         client: props.client,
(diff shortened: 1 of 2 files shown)


packages/thirdweb/src/react/web/ui/TransactionButton/TransactionModal.tsx
```diff
@@ -37,14 +37,19 @@ export function TransactionModal(props: ModalProps) {
   useQuery({
     queryKey: ["transaction-modal-event"],
     queryFn: () => {
-      if (!account || !wallet) return;
+      // Query data cannot be undefined, return null
+      if (!account || !wallet) {
+        return null;
```
joaquim-verges, line 42: should it just throw instead so it retries later? 

also should we add !!account to enabled prop below?

MananTank: updated

joaquim-verges (approved):

jnsdls (approved):

PR #6116, opened 2025-01-30
[SDK] Fix: skip factory entrypoint lookup for ZKsync chains

alecananian:
When connecting to an ecosystem wallet with smart wallet options configured in the Thirdweb dashboard, there is a lookup made to fetch any missing details for account abstraction, like the entrypoint address. If the default chain on the dashboard is set to a ZKsync chain (e.g., Treasure chain ID 61166), there is an RPC call made to the factory address that always returns `0x` because the factory doesn't exist and is irrelevant for native account abstraction. This change saves the extraneous RPC call during login.

---

## PR-Codex overview
This PR focuses on optimizing the handling of `ZkSync` chains in the `connectSmartAccount` function by adding a direct check for these chains and simplifying the gas management logic.

### Detailed summary
- Added a check for `ZkSync` chains to skip factory entrypoint lookup.
- Introduced `sponsorGas` variable to manage gas settings more effectively.
- Removed redundant code related to `ZkSync` account creation when a factory is passed.

### DIFF
--- a/packages/thirdweb/src/wallets/smart/index.ts
+++ b/packages/thirdweb/src/wallets/smart/index.ts
@@ -95,14 +95,22 @@ export async function connectSmartAccount(
 
   const options = creationOptions;
   const chain = connectChain ?? options.chain;
-  const isZksyncChain = await isZkSyncChain(chain);
+  const sponsorGas =
+    "gasless" in options ? options.gasless : options.sponsorGas;
+  if (await isZkSyncChain(chain)) {
+    return [
+      createZkSyncAccount({
+        creationOptions,
+        connectionOptions,
+        chain,
+        sponsorGas,
+      }),
+      chain,
+    ];
+  }
 
   // if factory is passed, but no entrypoint, try to resolve entrypoint from factory
-  if (
-    !isZksyncChain &&
-    options.factoryAddress &&
-    !options.overrides?.entrypointAddress
-  ) {
+  if (options.factoryAddress && !options.overrides?.entrypointAddress) {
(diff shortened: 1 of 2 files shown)


packages/thirdweb/src/wallets/smart/index.ts
```diff
@@ -122,13 +127,10 @@ export async function connectSmartAccount(
...
   }
 
-  const factoryAddress =
-    options.factoryAddress ??
-    getDefaultAccountFactory(options.overrides?.entrypointAddress);
   const sponsorGas =
     "gasless" in options ? options.gasless : options.sponsorGas;
 
-  if (await isZkSyncChain(chain)) {
+  if (isZksyncChain) {
```
joaquim-verges, line 133: might be cleaner to just move this block at the very top

joaquim-verges (approved):

packages/thirdweb/src/wallets/smart/index.ts
alecananian: My thinking was the paymaster and entrypoint address configs were still valid for ZKsync chains, but if that's not the case then yes the conditions can be moved around

joaquim-verges: no entrypoint is only for non-zk chains, so makes sense to check zk chain first

alecananian: Great! In that case, I've pushed an update that simplifies the ZKsync chain check

joaquim-verges (approved):

PR #6180, opened 2025-02-06
[SDK] Feature: Add support for forcing legacy transactions in chain configuration

d4mr:
solves TOOL-3532

---

## PR-Codex overview
This PR introduces enhancements related to transaction types and fee handling in the `thirdweb` package, adding support for new transaction types and fee structures.

### Detailed summary
- Added `feeType` to `ChainOptions`.
- Introduced `TransactionType` with various types.
- Updated `StaticPrepareTransactionOptions` to include `type`.
- Modified transaction serialization to handle `type`.
- Enhanced gas fee calculation to prioritize transaction fee types.
- Updated `getDefaultGasOverrides` to consider `feeType`.


packages/thirdweb/src/chains/types.ts
```diff
@@ -26,6 +26,12 @@ export type ChainOptions = {
     increaseZeroByteCount?: boolean;
   };
   faucets?: Array<string>;
+  fees?: {
+    /**
+     * Whether to force legacy transactions (pre EIP-1559) for this chain
+     */
+    forceLegacyTransactions?: boolean;
```
gregfromstl, line 33: Don't like the boolean because its inflexible, a `type: "legacy"` might be better (and matches viem/ox APIs)

gregfromstl: Might also want this on the transaction level if the user is providing it

d4mr: agreed on both counts. So something like `feeType: "legacy" | "eip1559"`?

packages/thirdweb/src/chains/types.ts
```diff
@@ -26,6 +28,7 @@ export type ChainOptions = {
     increaseZeroByteCount?: boolean;
   };
   faucets?: Array<string>;
+  feeType?: FeeType;
```
joaquim-verges, line 31: who will fill in this value though?

packages/thirdweb/src/gas/fee-data.ts
```diff
@@ -50,11 +53,13 @@ export async function getGasOverridesForTransaction(
...
-    resolvePromisedValue(transaction.maxFeePerGas),
-    resolvePromisedValue(transaction.maxPriorityFeePerGas),
-    resolvePromisedValue(transaction.gasPrice),
-  ]);
+  const [maxFeePerGas, maxPriorityFeePerGas, gasPrice, feeType] =
+    await Promise.all([
+      resolvePromisedValue(transaction.maxFeePerGas),
+      resolvePromisedValue(transaction.maxPriorityFeePerGas),
+      resolvePromisedValue(transaction.gasPrice),
+      resolvePromisedValue(transaction.feeType),
```
joaquim-verges, line 61: so we have it on the chain, on the tx AND in the function itself? seems a bit much no?

gregfromstl: It should come from the chains DB if possible but the user override is _only_ on the transaction. User override takes precedence

packages/thirdweb/src/chains/types.ts
d4mr: when a user defines chain, it's good to be able to globally specify only to use legacy transaction
(this is the reason this came on my radar to begin with lol, I need a way to do this in engine)

packages/thirdweb/src/gas/fee-data.ts
d4mr: for context, I felt this need on engine where I need to expose configuration that allows users to force legacy transactions on a chain. Especially important when engine sends nonce cancellation transactions which are not user triggered. (I could manage this in engine too, but felt like it could be a more general thing so the SDK would benefit from it)

joaquim-verges: yeah chain level makes sense to me

packages/thirdweb/src/gas/fee-data.ts
```diff
@@ -113,20 +120,27 @@ export async function getGasOverridesForTransaction(
...
-        maxFeePerGas: feeData.maxFeePerGas,
-        maxPriorityFeePerGas: feeData.maxPriorityFeePerGas,
-      };
-    }
+  // give priority to the transaction fee type over the chain fee type
+  const resolvedFeeType = feeType ?? chain.feeType;
+  // if chain is configured to force legacy transactions or is in the legacy chain list
+  if (
+    resolvedFeeType === "legacy" ||
+    FORCE_GAS_PRICE_CHAIN_IDS.includes(chain.id)
```
joaquim-verges, line 130: while you're at it can you add the chain in question to the FORCE_GAS_PRICE_CHAIN_IDS list please?

packages/thirdweb/src/transaction/prepare-transaction.ts
```diff
@@ -17,6 +17,7 @@ export type StaticPrepareTransactionOptions = {
   maxFeePerGas?: bigint | undefined;
   maxPriorityFeePerGas?: bigint | undefined;
   maxFeePerBlobGas?: bigint | undefined;
+  feeType?: FeeType | undefined;
```
joaquim-verges, line 20: actually lets not add this, lets just do the chain one.

this prop is too similar the tx.type one which also indicates if its a legacy or 1559 tx

d4mr: we still would want a way to tell at the transaction level, hey please calculate the gas and send as legacy. If not this interface, we can use a different one, any thoughts?

joaquim-verges: we have `type` in `SerializableTransaction` already should map to that

packages/thirdweb/src/gas/fee-data.ts
```diff
@@ -40,6 +43,7 @@ const FORCE_GAS_PRICE_CHAIN_IDS = [
   1942999413, // Humanity Testnet
   1952959480, // Lumia Testnet
   994873017, // Lumia Mainnet
+  1942999413, // Humanity Mainnet
```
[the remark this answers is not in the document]
joaquim-verges: looks like the bot is right @d4mr

joaquim-verges (approved):

packages/thirdweb/src/gas/fee-data.ts
```diff
@@ -115,20 +121,27 @@ export async function getGasOverridesForTransaction(
...
+    };
+  }
+  const feeData = await getDynamicFeeData(client, chain);
+  if (feeData.maxFeePerGas !== null && feeData.maxPriorityFeePerGas !== null) {
+    return {
+      maxFeePerGas: feeData.maxFeePerGas,
+      maxPriorityFeePerGas: feeData.maxPriorityFeePerGas,
+    };
   }
+  // TODO: resolvedFeeType here could be "EIP1559", but we could not get EIP1559 fee data. should we throw?
```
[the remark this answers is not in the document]
joaquim-verges: lets not throw, i think there's some cases where both type=1559 + gasprice are valid. polygon being one example IIRC

joaquim-verges (approved):

PR #6409, opened 2025-03-04
[service-utils] remove unneeded cf worker imports

arcoraven:
## PR-Codex overview
This PR focuses on removing unneeded Cloudflare worker types from the `service-utils` package and updating the way headers are handled in the `sendUsageV2Events` function.

### Detailed summary
- Updated the `@thirdweb-dev/service-utils` package version.
- Removed unnecessary imports of Cloudflare worker types in `packages/service-utils/src/cf-worker/index.ts`.
- Changed header handling in `sendUsageV2Events` function to use `Object.fromEntries` for better compatibility.


packages/service-utils/src/cf-worker/usageV2.ts
```diff
@@ -1,5 +1,5 @@
-import { Headers, type Request, fetch } from "@cloudflare/workers-types";
```
arcoraven, line 1: This `fetch` import is causing issues when imported from cf worker projects.

PR #6464, opened 2025-03-12
[SDK] Feature: Adds Universal Bridge

gregfromstl:
---

## PR-Codex overview
This PR introduces a new `Bridge` module to the `thirdweb` SDK, enabling users to perform buy and sell operations across chains, retrieve bridge routes, and check transaction statuses.

### Detailed summary
- Added `Bridge` module for Universal Bridge access.
- Implemented `Bridge.Buy` and `Bridge.Sell` for token transactions.
- Created functions for quoting and preparing transactions.
- Introduced `routes` function to retrieve available bridge routes.
- Added `status` function to check transaction statuses.
- Enhanced error handling with descriptive messages.

### DIFF
--- a/.changeset/clear-olives-know.md
+++ b/.changeset/clear-olives-know.md
@@ -1,0 +1,177 @@
+---
+"thirdweb": minor
+---
+
+Adds a new `Bridge` module to the thirdweb SDK to access the Universal Bridge.
+
+## Features
+
+### Buy & Sell Operations
+
+The Bridge module makes it easy to buy and sell tokens across chains:
+
+- `Bridge.Buy` - For specifying the destination amount you want to receive
+- `Bridge.Sell` - For specifying the origin amount you want to send
+
+Each operation provides two functions:
+1. `quote` - Get an estimate without connecting a wallet
+2. `prepare` - Get a finalized quote with transaction data
+
+#### Buy Example
+
+```typescript
+import { Bridge, toWei, NATIVE_TOKEN_ADDRESS } from "thirdweb";
+
+// First, get a quote to see approximately how much you'll pay
+const buyQuote = await Bridge.Buy.quote({
+  originChainId: 1, // Ethereum
+  originTokenAddress: NATIVE_TOKEN_ADDRESS,
+  destinationChainId: 10, // Optimism
+  destinationTokenAddress: NATIVE_TOKEN_ADDRESS,
+  buyAmountWei: toWei("0.01"), // I want to receive 0.01 ETH on Optimism
+  client: thirdwebClient,
+});
+
+console.log(`To get ${buyQuote.destinationAmount} wei on destination chain, you need to pay ${buyQuote.originAmount} wei`);
+
+// When ready to execute, prepare the transaction
+const preparedBuy = await Bridge.Buy.prepare({
+  originChainId: 1,
+  originTokenAddress: NATIVE_TOKEN_ADDRESS,
+  destinationChainId: 10,
+  destinationTokenAddress: NATIVE_TOKEN_ADDRESS,
+  buyAmountWei: toWei("0.01"),
+  sender: "0x...", // Your wallet address
+  receiver: "0x...", // Recipient address (can be the same as sender)
+  client: thirdwebClient,
+});
+
+// The prepared quote contains the transactions you need to execute
+console.log(`Transactions to execute: ${preparedBuy.transactions.length}`);
+```
+
+#### Sell Example
+
+```typescript
+import { Bridge, toWei } from "thirdweb";
+
+// First, get a quote to see approximately how much you'll receive
+const sellQuote = await Bridge.Sell.quote({
+  originChainId: 1, // Ethereum
+  originTokenAddress: NATIVE_TOKEN_ADDRESS,
+  destinationChainId: 10, // Optimism
+  destinationTokenAddress: NATIVE_TOKEN_ADDRESS,
+  sellAmountWei: toWei("0.01"), // I want to sell 0.01 ETH from Ethereum
+  client: thirdwebClient,
+});
+
+console.log(`If you send ${sellQuote.originAmount} wei, you'll receive approximately ${sellQuote.destinationAmount} wei`);
+
+// When ready to execute, prepare the transaction
+const preparedSell = await Bridge.Sell.prepare({
+  originChainId: 1,
+  originTokenAddress: NATIVE_TOKEN_ADDRESS,
+  destinationChainId: 10,
+  destinationTokenAddress: NATIVE_TOKEN_ADDRESS,
+  sellAmountWei: toWei("0.01"),
+  sender: "0x...", // Your wallet address
+  receiver: "0x...", // Recipient address (can be the same as sender)
+  client: thirdwebClient,
+});
+
+// Execute the transactions in sequence
+for (const tx of preparedSell.transactions) {
+  // Send the transaction using your wallet
+  // Wait for it to be mined
+}
+```
+
+### Bridge Routes
+
+You can discover available bridge routes using the `routes` function:
+
+```typescript
+import { Bridge, NATIVE_TOKEN_ADDRESS } from "thirdweb";
+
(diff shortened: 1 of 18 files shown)


packages/thirdweb/src/universal/Buy.ts
```diff
@@ -0,0 +1,263 @@
...
+import type { PreparedQuote, Quote } from "./types/Quote.js";
+
+/**
+ * Retrieves a Universal Bridge quote for the provided buy intent. The quote will specify the necessary `originAmount` to receive the desired `destinationAmount`, which is specified with the `buyAmountWei` option.
+ *
+ * @example
+ * ```typescript
+ * import { Universal } from "thirdweb";
+ *
+ * const quote = await Universal.Buy.quote({
```
joaquim-verges, line 14: not sure about `Universal` about the entrypoint, but also can't find something i feel much better about 😅 

kinda want to make it more explicit and call the entrypoint the full `UniversalBridge` word

```
import { UniversalBridge } from "thirdweb";

await UniversalBridge.Buy.quote(...)
```
 bit mouthful though.

gregfromstl: Yeah I feel the same, definitely not set on anything yet

gregfromstl: Settled on `Bridge`

joaquim-verges: I like bridge, let's go with that

.changeset/clear-olives-know.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### Buy Example
+
+```typescript
+import { Bridge, toWei } from "thirdweb";
+
+// First, get a quote to see approximately how much you'll pay
+const buyQuote = await Bridge.Buy.quote({
+  originChainId: 1, // Ethereum
+  originTokenAddress: "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE", // ETH
```
joaquim-verges, line 28: Would use the constant here. That's our constant too so would not expect ppl to know/understand it.

Another option would be to make the address optional, and only use tokenAddress for erc20.

Wdyt

.changeset/clear-olives-know.md
```diff
@@ -0,0 +1,177 @@
...
+```typescript
+import { Bridge, toWei } from "thirdweb";
+
+// First, get a quote to see approximately how much you'll receive
+const sellQuote = await Bridge.Sell.quote({
+  originChainId: 1, // Ethereum
+  originTokenAddress: "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE", // ETH
+  destinationChainId: 10, // Optimism
+  destinationTokenAddress: "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE", // ETH
+  sellAmountWei: toWei("0.01"), // I want to sell 0.01 ETH from Ethereum
```
joaquim-verges, line 64: This also makes the dev handle the decimals themselves, can be tricky.

Wonder if we should handle the conversion?

.changeset/clear-olives-know.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Bridge Routes
+
+You can discover available bridge routes using the `routes` function:
+
+```typescript
+import { Bridge } from "thirdweb";
+
+// Get all available routes
+const allRoutes = await Bridge.routes({
```
joaquim-verges, line 97: How slow is this call?

.changeset/clear-olives-know.md
joaquim-verges: maybe we can make a separate conversion helper that handles it, passing amount + contract, but leave this API pure 🤔

.changeset/clear-olives-know.md
gregfromstl: Very fast, just pulls from the db

.changeset/clear-olives-know.md
gregfromstl: Yeah, interstate accepted both which I thought was a little messy. I figured bigint/wei would be best since usually apps are storing onchain values as bigints anyway but we could definitely give a conversion helper. (/routes does return token decimals btw)

.changeset/clear-olives-know.md
gregfromstl: Will swap out for the constant, trying to keep the query params explicit as much as possible to reduce confusion

.changeset/clear-olives-know.md
```diff
@@ -0,0 +1,177 @@
...
+import { Bridge } from "thirdweb";
+
+// Check the status of a bridge transaction
+const bridgeStatus = await Bridge.status({
+  transactionHash: "0xe199ef82a0b6215221536e18ec512813c1aa10b4f5ed0d4dfdfcd703578da56d",
+  chainId: 8453, // The chain ID where the transaction was initiated
+  client: thirdwebClient,
+});
+
+// The status will be one of: "completed", "pending", or "failed"
```
joaquim-verges, line 131: or not_found?

packages/thirdweb/package.json
```diff
@@ -40,6 +40,11 @@
       "import": "./dist/esm/exports/chains.js",
       "default": "./dist/cjs/exports/chains.js"
     },
+    "./universal": {
```
joaquim-verges, line 43: do we need this export since you have it in the root?

also, you renamed the file to bridge.ts - not universal

packages/thirdweb/src/bridge/Sell.ts
```diff
@@ -0,0 +1,263 @@
...
+ * @param options.destinationTokenAddress - The address of the destination token.
+ * @param options.sellAmountWei - The amount of the origin token to sell.
+ * @param options.sender - The address of the sender.
+ * @param options.receiver - The address of the recipient.
+ * @param options.client - Your thirdweb client.
+ *
+ * @returns A promise that resolves to a non-finalized quote for the requested buy.
+ *
+ * @throws Will throw an error if there is an issue fetching the quote.
+ */
```
joaquim-verges, line 189: should we mark these APIs as @beta in case we iterate a bit on the types?

packages/thirdweb/src/bridge/Status.ts
```diff
@@ -0,0 +1,146 @@
...
+ *   ]
+ * }
+ * ```
+ *
+ * This status is for a **single origin transaction only**. If your route involves multiple transactions, you'll need to get the status for each of them individually.
+ *
+ * If sending multiple dependent sequential transactions, wait until `status` returns `completed` before sending the next transaction.
+ *
+ * You can access this function's input and output types with `status.Options` and `status.Result`, respectively.
+ *
```
joaquim-verges, line 82: would either create a new tag for these docs, or use `@buyCrypto` which is the tag we use for the other pay apis.

prob a new tag makes sense? we should also deprecate the old ones btw and point to the new ones

packages/thirdweb/src/bridge/index.ts
```diff
@@ -0,0 +1,8 @@
+export * as Buy from "./Buy.js";
+export * as Sell from "./Sell.js";
```
joaquim-verges, line 2: you might need to add these as entrypoints in scripts/typedoc.mjs so its picked up correctly by the autogen docs

packages/thirdweb/src/bridge/Sell.ts
gregfromstl: Yes for sure

packages/thirdweb/package.json
gregfromstl: Actually going to include it so users can import from root or do `thirdweb/bridge` since that's consistent with the other modules

packages/thirdweb/src/bridge/index.ts
joaquim-verges: reminder to add this

joaquim-verges (approved):

Issue #6810, opened 2025-04-22
Thirdweb changing return type of wagmi useAccount hook


encryptedDegen:
When implementing Thirdweb into my Next.js app, I had noticed unusual type errors. They were about type `string` not being assignable to type `Address` (`0x${string}`). 
I checked the type of the address I was getting from wagmi's useAccount hook and it was returning `string | uindefined` which was not the case before.

After removing the `<ThirdwebProvider>` from my app it was back to returning the correct types.

I have not used thirdweb before, however, I am positive that this is not the intended behaviour, since the address should maintain the type set by wagmi & viem to maintain type safety. Many web3 libraries do support the `Address` type therefore I am unable to implement thirdweb into my application, and converting every instance of `useAccount` instance that returns the `address` top the `Address` type is not desired.

### Package versions
Wagmi version: 2.14.16
viem cersion: 2.27.3
@thirdweb-dev/wagmi-adapter version: 0.2.58
thirdweb version: 5.95.2

jnsdls:
hi @encryptedDegen - thirdweb does use the option to override the internal types of AbiType

```ts
declare module "abitype" {
  export interface Register {
    AddressType: string;
  }
}
```

You can learn about that configuration here: https://abitype.dev/config

In the short term you should be able to add a similar declaration into your own app to do the opposite:

```ts
declare module "abitype" {
  export interface Register {
    AddressType: `0x${string}`;
  }
}
```

In the longer term we may re-visit this, but it would likely have to come with a major version bump to `thirdweb`.

0xDAEF0F:
I'm curious as to why the abitype got overriden to a more lenient one (string)? Trying to tackle this but there are many implications for this change.

0xDAEF0F:
I already have this solved. I touched a lot of files so tomorrow im gonna double check everything is good before I make a PR.

<img width="1158" alt="Image" src="https://github.com/user-attachments/assets/49ab6211-a599-4616-8e64-613659e39912" />

PR #6852, opened 2025-04-25
Fix: `useWalletBalance` and the `useActiveAccount` to work with `0x{string}` addresses instead of strings.

0xDAEF0F:
- Updated type annotations for wallet/account addresses from generic string types to the more specific `0x${string}` type across several files.
- This change addresses the issue where users would see a plain string instead of the expected `0x{string}` type for addresses, improving type safety and developer experience.
- Ensured that this fix applies to both useActiveAccount and useWalletBalance hooks, so address types are now consistent and accurate in these commonly used hooks.
- Only performed a type coercion (using as `0x${string}`) in a single location, specifically in the deprecated fromViemWalletClient function, to maintain backward compatibility without introducing unnecessary changes elsewhere.
- No functional logic was altered; only type definitions were updated for clarity and correctness.

solves issue #6810 

---

## PR-Codex overview
This PR focuses on enhancing type safety by changing the type of various `address` properties across multiple files to a more specific format, ensuring they conform to the Ethereum address pattern.

### Detailed summary
- Changed `address` type from `Address` to `0x${string}` in `wallet.ts`.
- Cast `viemAccount.address` to `0x${string}` in `viem.ts`.
- Updated `address` type to `0x${string} | undefined` in `useWalletBalance.ts`.
- Modified optional `address` type to `0x${string}` in `TokenListScreen.tsx`.
- Changed `walletAddress` type from `string` to `0x${string}` in `index.ts`.

### DIFF
--- a/packages/thirdweb/src/adapters/viem.ts
+++ b/packages/thirdweb/src/adapters/viem.ts
@@ -394,7 +394,7 @@ function fromViemWalletClient(options: {
     );
   }
   return {
-    address: viemAccount.address,
+    address: viemAccount.address as `0x${string}`,
     signMessage: async (msg) => {
       return options.walletClient.signMessage({
         account: viemAccount,
--- a/packages/thirdweb/src/react/core/hooks/others/useWalletBalance.ts
+++ b/packages/thirdweb/src/react/core/hooks/others/useWalletBalance.ts
@@ -13,7 +13,7 @@ import {
 
 type UseWalletBalanceOptions = Prettify<
   Omit<GetWalletBalanceOptions, "address" | "chain"> & {
-    address: string | undefined;
+    address: `0x${string}` | undefined;
     chain: Chain | undefined;
   }
 >;
--- a/packages/thirdweb/src/react/native/ui/connect/TokenListScreen.tsx
+++ b/packages/thirdweb/src/react/native/ui/connect/TokenListScreen.tsx
@@ -64,7 +64,7 @@ export const TokenRow = (props: {
   theme: Theme;
   client: ThirdwebClient;
   chain?: Chain;
-  address?: string;
+  address?: `0x${string}`;
   onTokenSelected?: (token?: TokenInfo) => void;
 }) => {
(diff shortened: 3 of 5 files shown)


jnsdls:
I feel like if we're going to do this we should do it... for all addresses RETURNED by the SDK across the entire codebase

I **think** there is no harm in RETURNING the more strict type everywhere, we obviously can't force existing users to PASS the more strict type everywhere because that would be a breaking change, but returning a stricter type should be non-breaking for everyone.

@joaquim-verges @MananTank @gregfromstl for your thoughts on this

jnsdls:
also thanks @0xDAEF0F for kicking this off, we'll get to a resolution on the overall plan here from our side and we'll get this merged if it fits the direction we want to go down

0xDAEF0F:
> I feel like if we're going to do this we should do it... for all addresses RETURNED by the SDK across the entire codebase
> 
> I **think** there is no harm in RETURNING the more strict type everywhere, we obviously can't force existing users to PASS the more strict type everywhere because that would be a breaking change, but returning a stricter type should be non-breaking for everyone.
> 
> @joaquim-verges @MananTank @gregfromstl for your thoughts on this

It’s technically a breaking change, but only in cases where the user wasn’t passing the address correctly to begin with. In that sense, it’s a good break — it surfaces a misuse that would have failed at runtime anyway.

jnsdls:
hm it's actually the other direction that this becomes doable without breaking every user:

1. we can remove the ABIType override
2. everywhere (across the entire codebase) where an address is an INPUT param to any function we need to make sure it is explicitly typed as `string` and not as `Address` (we do NOT want to break inputs without a major version bump)
3. everywhere where an address is an OUTPUT from any function it should be typed as `Address` (`0x${string}`)

returning a more strict type is completely non breaking because `0x${string}` is a subset of `string` (which is what we are returning today), meaning anything that relies on it being a `string` in a user's codebase will still be happy to receive `0x${string}`.

Happy for you to take a stab at this @0xDAEF0F - however this is definitely a big chunk of the codebase (even though only tiny individual parts) that will need to change.

PR #7108, opened 2025-05-21
[OnRamp] Feature: Allow for supported providers to be passed

RobbyUitbeijerse:
Currently the FiatScreen in PayEmbed relies on a predefined set of providers: Coinbase, Transak and Stripe. All of these providers would work if not for the reality that some tokens are only supported by a subset of these providers.

With this PR I've added an additional prop to the `payOptions.buyWithFiat`, `supportedProviders` . It's an array with FiatProvider values which can be used in tandem with the preferred provider.

If you pass an empty array, or a `payOptions.buyWithFiat.preferedProvider` that doesn't match with the provided values, the property will be ignored and the default set of providers is being used, making it completely optional.

If there's only a single provider passed down, the dropdown is replaced for a <Text> component rather than a <Button> and <Link> combination to handle the provider selection.


packages/thirdweb/src/react/web/ui/ConnectWallet/screens/Buy/fiat/FiatScreenContent.tsx
```diff
@@ -84,6 +87,15 @@ export function FiatScreenContent(props: {
       : undefined,
   );
 
+  const supportedProviders = (() => {
+    if (!buyWithFiatOptions) return [];
+    const options = buyWithFiatOptions?.supportedProviders ?? [];
+    return options.length > 0 &&
+      (!preferredProvider || options.includes(preferredProvider))
```
joaquim-verges, line 94: if i dont pass preferredProvider, supportedProviders is not respected.

i think its because at this level preferredProvider is always defaulted to coinbase ?

would be nice to make preferredProvider default to the first one in my supportedProviders if not set

couple more nits:

- would be nice to not have the mini modal open at all if only 1 provider is supported
- we should probably do the first letter uppercase treatment inside the modal too :p

RobbyUitbeijerse: Yep sorry, did this blind yesterday. Actually ran the Playground this time around. Here's what I did:

* Adjusted the presentation logic to include a non interactive view of the provider name
* Applied the `uppercaseFirstLetter` throughout the flow
* Adjusted prop logic for preferred/supported provider(s)


I covered the following situations while testing just now:

1. `payOptions.buyWithFiat: undefined` - assumes that buy with fiat is allowed with all providers, default list of fiat providers is now being rendered.
2. `payOptions.buyWithFiat.preferredProvider: 'STRIPE' - allows for all providers, auto-selects stripe (or whatever you pass down)
3. In all scenarios I tested where the supportedProviders conflicts with the preferredProvider, the default list of fiat providers is being rendered. (eg: empty supportedProviders array, preferredProvider with different value than passed in supportedProviders)

Should be quite alright now. Let me know if you find something odd. 

PS In case you're wondering what's up with the coderabbitai comments; did a merge rather than a rebase and made coderabbitai throw up. I fixed that but it opened a lot of comments for things commits and changes that were already in main due to my fork it's commit history flowing differently. Resolved that, and closed those comments.

PR #7157, opened 2025-05-26
[SDK] ensure that the default spinner stroke color gets inherited from the parentnode

RobbyUitbeijerse:
In a few cases the <Spinner> got an explicit color passed, eg `accentButtonText`, but this isn't great if the element in which is rendered has a variant or style prop that affects the color property of said element, as the Spinner will simply keep rendering in the color you passed it, unless you also dynamically switch the color property of the Spinner in similar fashion to the parent.

In the case of the FiatScreenContent and SwapScreenContent, the Button component that's rendered switches between the 'accent' and 'outline' variant when the buttons are disabled, which causes the Spinner to render the 'accentButtonText' color even though the 'outline' variant of the Button was active.

I considered two ways of fixing this

1 Removing the variant switching
2 Simply allowing the Spinner to inherit the color of the parent node

The latter seemed like the most flexible solution. The current color property on the Spinner is required so there's no risk in breaking any existing instance of the Spinner component, and it allows for simpler usage without explicit color passing.

Let me know if you prefer option 1 here, or whenever you see a different solution, happy to adjust where needed.

---

## PR-Codex overview
This PR focuses on updating the `Spinner` component by making the `color` prop optional and setting its default stroke color to `currentColor`. This change simplifies the usage of the `Spinner` in different contexts, particularly in the `FiatScreenContent` and `SwapScreenContent` components.

### Detailed summary
- Updated the `Spinner` component to make the `color` prop optional.
- Set the default stroke color of the `Spinner` to `currentColor`.
- Removed the `color` prop from `Spinner` instances in `FiatScreenContent.tsx` and `SwapScreenContent.tsx`.

## Summary by CodeRabbit

- **Style**
  - Updated spinner appearance in wallet screens by removing explicit color styling, enabling it to inherit default or theme-based colors.  
- **Refactor**
  - Enhanced spinner component flexibility by making the color optional and defaulting to the current text color when unspecified.

### DIFF
--- a/.changeset/fast-onions-obey.md
+++ b/.changeset/fast-onions-obey.md
@@ -1,0 +1,5 @@
+---
+"thirdweb": patch
+---
+
+default the spinner stroke color to currentColor and make the color prop optional
--- a/packages/thirdweb/src/react/web/ui/ConnectWallet/screens/Buy/fiat/FiatScreenContent.tsx
+++ b/packages/thirdweb/src/react/web/ui/ConnectWallet/screens/Buy/fiat/FiatScreenContent.tsx
@@ -230,7 +230,7 @@ export function FiatScreenContent(props: {
       </Container>
 
       <Button
-        variant={disableSubmit ? "outline" : "accent"}
+        variant="accent"
         data-disabled={disableSubmit}
         disabled={disableSubmit}
         fullWidth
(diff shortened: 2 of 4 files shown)


joaquim-verges (approved):

joaquim-verges (commented):
can you add a lil changeset?

pnpm changeset from the root of the repo, press enter until you get to enter the msg

PR #7462, opened 2025-06-27
[SDK] Feature: Adds location to platform headers

gregfromstl:

### DIFF
--- a/packages/thirdweb/src/utils/fetch.ts
+++ b/packages/thirdweb/src/utils/fetch.ts
@@ -228,6 +228,10 @@ export function getPlatformHeaders() {
     ...(bundleId ? { "x-bundle-id": bundleId } : {}),
   });
 
+  if (typeof window !== "undefined") {
+    previousPlatform.push(["x-sdk-location", window.location.href]);
+  }
+
   return previousPlatform;
 }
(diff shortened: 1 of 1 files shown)


packages/thirdweb/src/utils/fetch.ts
```diff
@@ -228,6 +228,10 @@ export function getPlatformHeaders() {
     ...(bundleId ? { "x-bundle-id": bundleId } : {}),
   });
 
+  if (typeof window !== "undefined") {
+    previousPlatform.push(["x-sdk-location", window.location.href]);
```
joaquim-verges, line 232: this feels a bit odd no? also make sure to handle location undefined as well for some platforms. RN actually can have a window but no location i think

gregfromstl: How is it odd?

joaquim-verges: i think it fits better as a param to the tracking event rather than a header to all requests

joaquim-verges: can we just add it to the trackPayEvent / trackConnectEvent for now?

PR #7638, opened 2025-07-17
fix gasPrice for legacy chain 88/89

nischitpra:


packages/thirdweb/src/wallets/smart/lib/userop.ts
```diff
@@ -269,6 +269,11 @@ async function getGasFees(args: {
       maxPriorityFeePerGas =
         resolvedMaxPriorityFeePerGas ?? feeData.maxPriorityFeePerGas ?? 0n;
       maxFeePerGas = resolvedMaxFeePerGas ?? feeData.maxFeePerGas ?? 0n;
+
+      if (chain.id === 88 || chain.id === 89) {
```
joaquim-verges: Should this logic live in getDefaultGasOverride instead? That way it'll apply to regular txs too

PR #7641, opened 2025-07-17
[SDK] Feature: Adds gateway usage source

gregfromstl:


packages/service-utils/src/core/services.ts
```diff
@@ -83,6 +83,13 @@ export const SERVICE_DEFINITIONS = {
     name: "storage",
     title: "Storage",
   },
+  gateway: {
```
joaquim-verges, line 86: i think the dashboard reads this btw, would make sure it does what you want (particularly enabled by default) also api server needs to have this if this is gonna be a proper projectService

we can store dev settings in there in the future

PR #7973, opened 2025-09-03
[SDK] Feature: added glyph wallet in extra wallets

0xShoobs:
## PR-Codex overview
This PR focuses on adding support for the `Glyph Wallet` and making various updates to existing wallet configurations, including changes to app details, image references, and URL updates for several wallets.

### Detailed summary
- Added `Glyph Wallet` to supported wallets.
- Updated wallet names and homepages for several wallets.
- Changed image IDs for specific wallets.
- Modified app platform details (native and universal) for multiple wallets.
- Removed deprecated wallet entries and images.
- Added auto-generated image files for new wallets.

> The following files were skipped due to too many changes: `packages/thirdweb/src/wallets/__generated__/wallet/xyz.timelesswallet/image.ts`, `packages/thirdweb/src/wallets/__generated__/wallet/io.useglyph/image.ts`, `packages/thirdweb/src/wallets/__generated__/wallet/com.dcentwallet/image.ts`, `packages/thirdweb/src/wallets/__generated__/wallet-ids.ts`, `packages/thirdweb/src/wallets/__generated__/wallet-infos.ts`, `packages/thirdweb/src/wallets/__generated__/getWalletInfo.ts`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

## Summary by CodeRabbit

* **New Features**
  * Added Glyph Wallet as an available option in the wallet connection flow; users can now select it alongside existing wallets.
  * Glyph appears with its display name and icon for easier discovery and use.
  * This addition does not affect existing wallets, settings, or current wallet selections; it will be available in the upcoming patch release.

PR #8034, opened 2025-09-10
[SDK] Add apechain to supported chains and Add Glyph Wallet Connector

shubham-yuga:
As discussed [here](https://github.com/thirdweb-dev/js/pull/7973#issuecomment-3266290442), this is the second part of adding Glyph support to Thirdweb. 

## How to test?

Go to playground and choose Glyph Wallet from the Extra wallets option. Should be able to login.


packages/thirdweb/src/exports/wallets/glyph.ts
```diff
@@ -0,0 +1,4 @@
+export {
+  glyphWalletTW,
+  glyphWalletWagmiConnector,
+} from "../../wallets/glyph/glyph-wallet.js";
```
joaquim-verges, line 4: hm these should be in their own package, not directly baked into the thirdweb package. Same thing we do with abstract global wallet.

packages/thirdweb/package.json
```diff
@@ -25,6 +26,7 @@
     "@tanstack/react-query": "5.81.5",
     "@thirdweb-dev/engine": "workspace:*",
     "@thirdweb-dev/insight": "workspace:*",
+    "@wagmi/core": "^2.17.2",
```
joaquim-verges, line 29: yeah these dependencies cannot be here. Needs to be its own package and users will install both your package and thirdweb together. Check how we do it for abstract

PR #5354, opened 2024-11-08
[SDK] Feature: Adds EIP-1193 Adapter

joaquim-verges:
CNCT-2339

---

## PR-Codex overview
This PR introduces EIP-1193 adapters to facilitate seamless integration between Thirdweb wallets and EIP-1193 providers, enhancing wallet functionality and allowing for better compatibility with web3 libraries.

### Detailed summary
- Added `EIP1193.fromProvider()` to create Thirdweb wallets from EIP-1193 providers.
- Added `EIP1193.toProvider()` to convert Thirdweb wallets to EIP-1193 providers.
- Updated various wallet configurations and IDs.
- Enhanced wallet info generation for new wallets (`social.gm2`, `greenhood`, `blazpay`, `abstract`).
- Improved type definitions for better TypeScript support.
- Added tests for EIP-1193 functionality.
- Modified wallet connection logic to support EIP-1193.

> The following files were skipped due to too many changes: `packages/thirdweb/src/wallets/injected/index.ts`, `pnpm-lock.yaml`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #8886, opened 2026-08-13
[SDK] Fix: EIP1193.toProvider() removeListener is now functional

blockgroot:
## Notes for the reviewer

`EIP1193.toProvider()` converts a thirdweb `Wallet` into a standard EIP-1193 `provider` for use with any EIP-1193-consuming library (wagmi, viem's `custom()` transport, ethers, etc.).

`removeListener` was a permanent no-op — the code itself left a comment admitting the gap:

```ts
on: wallet.subscribe,
removeListener: () => {
  // should invoke the return fn from subscribe instead
},
```

`wallet.subscribe(event, cb)` returns an unsubscribe function, but assigning `on: wallet.subscribe` directly discards that return value (the `EIP1193Provider.on` type is declared `void`), so nothing ever called it. Any consumer that does `provider.on("accountsChanged", handler)` and later `provider.removeListener("accountsChanged", handler)` — the standard pattern for cleanup on unmount/disconnect — leaked the listener; `handler` kept firing after the caller believed it had unsubscribed.

This isn't hypothetical inside this repo either: `packages/wagmi-adapter/src/connector.ts` calls `EIP1193.toProvider()` directly to implement a wagmi `Connector`'s `getProvider()`, which is exactly the kind of caller that registers/deregisters provider listeners for lifecycle management.

I checked git history and the original PR (#5354) review before filing this — the no-op was introduced in the adapter's founding commit and hasn't been touched across 7 follow-up PRs to this file; no issue or PR has discussed it since. Looks like a straightforward oversight rather than an intentional simplification.

**Fix:** track the unsubscribe function returned by `wallet.subscribe()` per `(event, listener)` pair, and have `removeListener` invoke and clear it. No signature/type changes — `on`/`removeListener` keep the same shape, just implemented correctly.

## How to test

```
cd packages/thirdweb
pnpm test:dev src/adapters/eip1193/to-eip1193.test.ts
```

Added test `removeListener should detach a listener registered via on`:
- registers a listener via `provider.on(...)`, emits the event, confirms it fires
- calls `provider.removeListener(...)`, emits again, confirms it does **not** fire again

Confirmed this test fails on `main` (listener fires twice instead of once) and passes with this fix. The only other failure in this file (`should handle eth_sendTransaction`) is pre-existing on `main` and unrelated — it requires a `TW_SECRET_KEY` for the mainnet fork that isn't set in this environment.

Ran `pnpm lint` from repo root — clean, no new warnings introduced by this change.

---

## PR-Codex overview
This PR enhances the `removeListener` method in `EIP1193.toProvider()` to properly detach event listeners, addressing a previous issue where listeners could not be unsubscribed. 

### Detailed summary
- Updated `removeListener` to invoke the correct unsubscribe function for each `(event, listener)` pair.
- Introduced a `Map` to track unsubscribe functions for registered listeners.
- Added tests to verify that `removeListener` correctly detaches listeners.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed event listener removal for EIP-1193 providers.
  * Listeners registered through `provider.on` are now properly detached when removed, preventing further callbacks.

* **Tests**
  * Added coverage confirming removed listeners no longer receive event notifications.

* **Documentation**
  * Added a patch release note for the listener removal fix.

### DIFF
--- a/.changeset/eip1193-remove-listener-noop.md
+++ b/.changeset/eip1193-remove-listener-noop.md
@@ -1,0 +1,5 @@
+---
+"thirdweb": patch
+---
+
+Fix: `EIP1193.toProvider()`'s `removeListener` is no longer a no-op. Previously, `removeListener` discarded the unsubscribe function returned by `wallet.subscribe()`, so listeners registered via `provider.on(...)` (e.g. `accountsChanged`, `chainChanged`, `disconnect`) could never actually be detached — they kept firing after callers (such as wagmi connectors) believed they had unsubscribed. `removeListener` now tracks and invokes the correct unsubscribe function per `(event, listener)` pair.
--- a/packages/thirdweb/src/adapters/eip1193/to-eip1193.test.ts
+++ b/packages/thirdweb/src/adapters/eip1193/to-eip1193.test.ts
@@ -190,6 +190,25 @@ describe("toProvider", () => {
     ).resolves.toEqual([]);
   });
 
+  test("removeListener should detach a listener registered via on", () => {
+    const provider = toProvider({
+      chain: ANVIL_CHAIN,
+      client: TEST_CLIENT,
+      wallet: mockWallet,
+    });
+
+    const listener = vi.fn();
+    provider.on("accountsChanged", listener);
+
+    emitter.emit("accountsChanged", [mockAccount.address]);
+    expect(listener).toHaveBeenCalledTimes(1);
+
+    provider.removeListener("accountsChanged", listener);
+
+    emitter.emit("accountsChanged", [mockAccount.address]);
+    expect(listener).toHaveBeenCalledTimes(1);
+  });
+
   test("should use custom connect override when provided", async () => {
     const walletWithoutAccount = {
       ...mockWallet,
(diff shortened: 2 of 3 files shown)


blockgroot:
@coderabbitai Checked this against the underlying implementation before making a change.

Every `Wallet` in this repo is built with `createWalletEmitter()` (`packages/thirdweb/src/wallets/wallet-emitter.ts`), which wraps `createEmitter()` from `packages/thirdweb/src/utils/tiny-emitter.ts`. That emitter stores subscribers in a `Set<callback>` per event (`tiny-emitter.ts:31-58`), so calling `wallet.subscribe(event, cb)` twice with the *same* `cb` reference is a no-op on the second add — the `Set` still holds exactly one entry, and either of the two returned unsubscribe closures removes that same single entry.

So overwriting the `Map<listener, unsubscribe>` entry on a duplicate `on()` call doesn't leak anything in practice: the stored closure (whichever one it is) still fully detaches the listener on the next `removeListener` call.

I added a test (`removeListener fully detaches a listener registered twice via on`) that registers the same listener twice, then confirms a single `removeListener` call fully stops it from firing — verified failing before this reasoning was checked, passing now, and it'll catch a regression if the emitter's dedup behavior ever changes.

Didn't change the bookkeeping structure itself since the risk isn't reachable given the current emitter contract, and adding per-registration counting would be complexity without a corresponding bug.

PR #8938, opened 2026-09-07
[SDK] Stop swallowing token lookup errors in useTokenQuery

### LINKED ISSUE #8937: CheckoutWidget shows "Token Not Supported" for any token lookup failure (swallowed error in useTokenQuery)
### Summary

`CheckoutWidget` (and `TransactionWidget`, which shares the same hook) renders the dead-end **"Token Not Supported – This token or chain is not supported by the Bridge"** screen for *any* failure of the destination-token lookup, not only for genuinely unsupported tokens. A dropped request on mobile, a `401` from a host that is not on the client-id allowlist, a `429`, a `5xx`, or a non-JSON error page from the edge all end up on that screen, with no retry, no `onError` callback, and the verdict cached for the session.

### Root cause

`packages/thirdweb/src/react/web/ui/Bridge/common/token-query.ts` ([permalink](https://github.com/thirdweb-dev/js/blob/9b9c072929f0461393cec276ff29b58d89195bf2/packages/thirdweb/src/react/web/ui/Bridge/common/token-query.ts#L25-L37)):

```ts
const token = await getToken(params.client, tokenAddress, params.chainId).catch((err) => {
  err.message.includes("not supported") ? undefined : Promise.reject(err);
});

if (!token) {
  return { type: "unsupported_token" };
}
```

The arrow function has a block body and never `return`s, so the `.catch` handler resolves to `undefined` for every error. `Promise.reject(err)` is created but discarded (it only produces an unhandled-rejection warning). `token` is therefore `undefined` on any failure and the hook reports `unsupported_token`.

Two knock-on effects make it worse in production:

1. Because the queryFn *returns* instead of throwing, react-query treats the failure as a successful result: the default `retry` never runs, and with `refetchOnMount: false` + the provider's `staleTime: 60_000` the cached `unsupported_token` result is reused when the widget is re-mounted (for example when a user closes and reopens a checkout dialog) until `gcTime` expires. Only a full page reload clears it.
2. `UnsupportedTokenScreen` does not call `props.onError`, so integrators cannot observe or log these failures.

Additionally, if the rejection reason is not an `Error` (for example the `"timeout"` string passed to `controller.abort("timeout")` in `getClientFetch`), `err.message.includes` itself throws a `TypeError`.

### Reproduction

1. Render `<CheckoutWidget client={client} chain={polygon} tokenAddress="0x3c499c542cEF5E3811e1192ce70d8cC03d5c3359" amount="9" seller="0x..." />` with a token that *is* supported by the Bridge (Polygon USDC).
2. In DevTools, block `bridge.thirdweb.com` (Network → Request blocking), or serve the page from a host that is not on the client-id allowlist so `/v1/tokens` answers `401`.
3. The widget shows "Token Not Supported" immediately; `onError` is not invoked; unblocking the host and re-mounting the widget within a few minutes still shows the same screen.

Observed on `thirdweb@5.120.0` and `5.121.1`; the code is unchanged on `main`.

### Suggested fix

```ts
const token = await getToken(params.client, tokenAddress, params.chainId).catch((err) => {
  if (err instanceof Error && err.message.includes("not supported")) {
    return undefined;
  }
  throw err;
});
```

With this change a real "not supported" answer still renders `UnsupportedTokenScreen`, while every other failure propagates to `tokenQuery.error`, react-query retries with its default policy, and the widget falls through to its existing `ErrorBanner` (which already has a Retry button). It would also be worth calling `props.onError` from that branch so integrators can log the failure.

Happy to open a PR with this change if that helps.

devtechedge:
The `.catch` in `useTokenQuery` used a block body and never returned, so `Promise.reject(err)` was discarded. Every destination-token lookup failure — 401, 429, timeout, non-Error abort — became `{ type: "unsupported_token" }`. React Query then cached that as success (`refetchOnMount: false`), and CheckoutWidget / TransactionWidget rendered the dead-end Token Not Supported screen instead of the existing ErrorBanner retry path.

### Changes

- [x] Return `undefined` only for genuine `Error`s whose message includes `"not supported"`; rethrow everything else (including non-Error rejections)
- [x] Add unit tests for success, `"Token not supported"`, 401, and non-Error (`"timeout"`) rejections
- [x] Leave `UnsupportedTokenScreen` / `onError` alone — that is a separate UX change. The widget already renders `ErrorBanner` when the query actually errors

Fixes #8937

### Test plan

- `pnpm test:dev token-query` from `packages/thirdweb` (mocked `getToken`; no network)
- Manual: render `<CheckoutWidget>` with a supported token, block `bridge.thirdweb.com` in DevTools — should show ErrorBanner + Try Again, not Token Not Supported

---

## PR-Codex overview
This PR focuses on improving error handling in the `CheckoutWidget` by refining how token lookup failures are managed, ensuring unsupported tokens are handled gracefully, and enhancing the associated tests for the `useTokenQuery` hook.

### Detailed summary
- Updated error handling in `token-query.ts` to throw an error for unsupported tokens.
- Added tests for `useTokenQuery` in `token-query.test.tsx` to validate:
  - Successful token retrieval.
  - Handling of unsupported tokens.
  - Proper error reporting for 401 errors.
  - Handling non-Error rejections.

## Summary by CodeRabbit

- **Bug Fixes**
  - Checkout token lookups now distinguish unsupported tokens from other failures.
  - Authentication errors and unexpected lookup failures are surfaced correctly instead of being incorrectly reported as unsupported tokens.
  - Supported token results continue to load normally.
- **Tests**
  - Added coverage for successful lookups, unsupported tokens, authentication failures, and unexpected error responses.


devtechedge:
Thanks @0xFirekeeper for reviewing and landing this.

Tracing CheckoutWidget taught me something I will not miss next time: the `.catch` in `useTokenQuery` used a block body, so `Promise.reject(err)` was discarded.

Every lookup failure — 401, 429, timeout, non-Error abort - collapsed into `{ type: "unsupported_token" }`, and React Query then cached that as success (`refetchOnMount: false`).

Distinguishing a genuine `"not supported"` `Error` from the rest, and letting the query actually error so `ErrorBanner` can retry, was a cleaner split than I first expected.

Appreciate the fast turnaround, and the chance to work on this path.

